PR #247, opened 2015-12-17
admin page

paczian:
fixed affiliation bug
cleaned up js, html and css
added admin page
added impersonation


meren:
Yay! This looks awesome, Tobi. I can't wait to play with it.

Issue #283, opened 2016-01-18
Django template issue


meren:
The way we initialize Django templates will soon be obsolete, and it needs a fix.

To reproduce the problem run the mini test once:

``` bash
cd tests/
./run_mini_test.sh
```

You will get this warning at some point in the process:

```
RemovedInDjango110Warning: You haven't defined a TEMPLATES setting. You must do so before
upgrading to Django 1.10. Otherwise Django will be unable to load templates.
     "unable to load templates.", RemovedInDjango110Warning)
```

To get it over and over again, you can simply run this from within the `tests/` directory:

``` bash
anvi-summarize -p sandbox/test-output/204-MERGED/PROFILE.db -c sandbox/test-output/CONTIGS.db -o sandbox/test-output/204-MERGED-SUMMARY -C 'CONCOCT'
```

The relevant code is in the `anvio/summaryhtml.py` file, and it looks like this:

``` python
try:
    from django.conf import settings
    absolute = os.path.join(os.path.dirname(os.path.realpath(__file__)))
    template_dir = os.path.join(absolute, 'data/static/template')
    html_content_dir = os.path.join(absolute, 'data/static/content')
    settings.configure(DEBUG=True, TEMPLATE_DEBUG=True, DEFAULT_CHARSET='utf-8', TEMPLATE_DIRS = (template_dir,))

    try:
        import django
        django.setup()
    except:
        pass

    from django.template.loader import render_to_string
    from django.template.defaultfilters import register
except ImportError:
    raise ConfigError, 'You need to have Django module (http://djangoproject.com) installed on your system to generate HTML output.'
```

Something changes here, and the warning goes away. That is my intuition, however, I didn't have an opportunity to extensively research this.

If @gkmngrgn or @ozcanesen wants to sevabına take a stab at this, I would be forever grateful ;) Otherwise I will get back to this before Django people release v1.10!

gkmngrgn:
@meren you can assign me, I will look at this night.

meren:
Yay :) Feel free to release yourself from this burden if it ends up taking much of your time. I'm afraid the fix is going to be a one-liner, but finding the right one might take some time.

Best,

meren:
You are the best, Gokmen. Thank you *very* much.

PR #284, opened 2016-01-19
refs #283: define TEMPLATES variable for supporting Django 1.10+

gkmngrgn:


gkmngrgn:
Tested in:

* Django 1.9.1
* Django 1.7.11

Issue #274, opened 2016-01-11
Changes in the way we compute variability profile


meren:

meren:
We have a very computationally challenging problem. We want to know whether a nucleotide position in contigs db occurs in an open reading frame, whether in a complete gene call (proper start and stop positions), and which base position in a codon it corresponds to. We need this information while creating variability profile, however, the best place to recover this information is when the contigs database is generated --way before profiling, during which we learn which nucleotide positions we need to focus on due to their variable nature.

The commit 005faa441be5cf080f1cce7622622aa618dd9ee1 lays the groundwork In order to store positional information for each nucleotide position in the most compact way possible. After this we will simply compress the information into a small int, and send it into an HDF auxiliary database.

In this schema every nt position will be represented by four bits, and it will translate like this:

``` bash
            0000
            ||||
            ||| \
            |||   Third codon?
            || \
            ||   Second codon?
            | \
            |   First codon?
             \
               Whether the position in a partial gene call

        8: int('1000', 2); nt position is in a partial gene call
        4: int('0100', 2); nt position is in a complete gene call, and is at the 1st position in the codon
        2: int('0010', 2); nt position is in a complete gene call, and is at the 2nd position in the codon
        1: int('0001', 2); nt position is in a complete gene call, and is at the 3rd position in the codon
```

Following commits will facilitate this.

tdelmont:
I would like to suggest one thing here.
Eventually, we also want to know the amino acid (aa) of the consensus?
Then, knowing the ref aa, and the position in the aa, it would be relatively easy to understand the effect of any SNV position on the aa, no?
Because there are 20 aa, two more numbers would be enough to provide this information:

            000000
            ||||||
            ||||  \
            ||||  Amino acid?
            ||| \
            |||   Third codon?
            || \
            ||   Second codon?
            | \
            |   First codon?
             \
               Whether the position in a partial gene call

Or is this collected somewhere else? Or is this not relevant???

Tom

PR #314, opened 2016-02-02
Fix ColumnProfile C extension as #274

farukuzun:
`run_column_profilers_test.py` outputs are identical


farukuzun:
Something is wrong, please do not merge it yet, I'm rebasing it.

Issue #440, opened 2017-01-06
cogs module looking for blastdb in wrong place


jmeppley:
I'm running the anvi-run-ncbi-cogs module in the default setup and getting this error:

```
Config Error: Anvi'o understands that you want to use 'blastp' to search for COGs, however,
              there is no database formatted under the COGs data directory for that program :/
              You may need to re-run the COGs setup, UNLESS, you set up your COG data
              directory somewhere else than what anvi'o attempts to use at the moment
              ('/home/jmeppley/.conda/envs/anvio/lib/python2.7/site-
              packages/anvio-2.1.1.dev0-py2.7-linux-x86_64.egg/anvio/data/misc/COG'). If that
              is the case, this may be the best time to point the right directory using the
              --cog-data-dir parameter.
```

I found this in the anvio/cogs.py which looks like a bug (it sets blast_db_path, but check diamond_db_path):
```
        blast_db_path = J(self.COG_data_dir, 'DB_BLAST')
        if os.path.exists(diamond_db_path):
            formatted_db_paths['blastp'] = J(blast_db_path, 'COG')
```

I am working on a patch and will submit a pull request when I have it.

meren:
Oh, yes, this is absolutely a bug :) I will wait for your PR. Thank you very much!

PR #441, opened 2017-01-06
Issue #440, fixing db check to be blast not diamond

jmeppley:
Sorry for the delay, I accidentally deleted my COG files when setting up a devel environment.


meren:
Thank you very much! :)

PR #443, opened 2017-01-12
removing useless code that is causing collisions in multi-user enviro…

jmeppley:
…nments


meren:
thanks, John! :) clearly @tdelmont should be more careful with his changes in the repository. we don't have time with such irresponsible coding.

tdelmont:
Yes, I apology. I am a bit lost with all the contributions of made in this so called "code",
so thanks for cleaning after me :)

meren:
@tdelmont, well, I think you owe John a beer.

tdelmont:
perfect, you get me the flight ticket, I will pay both of you a beer!

PR #457, opened 2017-02-15
Fix anvi-gen-variability-matrix due to missing min-ratio-of-competings-nts

blankenberg:


meren:
Hey Daniel,

Thank you very much! I have no idea who took it out, because I am sure it was in there at some point :)

In fact this fix reminded me of something else. The output of this program is a subset of the information that is generated by `anvi-gen-variability-profile` (i.e., this one focuses only on the competing nts information). I feel like maybe we should remove or merge this one into `anvi-gen-variability-profile` to reduce confusion. But that is a decision that can be done later :)

Thank you for the fix!

PR #534, opened 2017-06-05
Split coverages update

meren:
Hi @ShaiberAlon,

The performance is about the same, but I improved the design a little bit. Now when you do this, it should give you access to a dictionary that keeps the coverage of each nt position across all samples for every split:

``` python
p = dbops.ProfileSuperclass(args)
p.init_split_coverage_values_per_nt_dict()

# do something with `p.split_coverage_values_per_nt_dict`
```

But only calling that function will not initialize gene coverages. The following are operationally the same, but they will also give access to `p.gene_detection_dict` and `p.gene_coverages_dict`, along with `p.split_coverage_values_per_nt_dict`:

``` python
p = dbops.ProfileSuperclass(args)
p.init_split_coverage_values_per_nt_dict()
p.init_gene_coverages_and_detection_dicts()
```
or

``` python
p = dbops.ProfileSuperclass(args)
p.init_gene_coverages_and_detection_dicts()
```

Use either, and enjoy the nucleotide-level coverage values :)

Please review the code, and merge it to the master if you are fine with these changes.

Best,

### DIFF
--- a/anvio/auxiliarydataops.py
+++ b/anvio/auxiliarydataops.py
@@ -29,7 +29,10 @@ pp = terminal.pretty_print
 
 
 class HDF5_IO(object):
-    def __init__(self, file_path, unique_hash, create_new=False, open_in_append_mode=False, ignore_hash=False):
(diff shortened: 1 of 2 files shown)


ShaiberAlon:
Looks good to me! :-)

PR #535, opened 2017-06-05
Update anvi-export-splits-and-coverages

AstrobioMike:
Hey there, I noticed the splits coverage table from this script was exporting with all splits having the same coverage, i.e. the parent contig coverage. Things were correct in the sqlite splits table so i 'think' this is the only change needed

### DIFF
--- a/bin/anvi-export-splits-and-coverages
+++ b/bin/anvi-export-splits-and-coverages
@@ -48,7 +48,7 @@ def main(args):
     if not args.output_file_prefix:
         args.output_file_prefix = merged_profile_db.meta['sample_id']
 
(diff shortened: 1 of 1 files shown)


meren:
Hey Mike,

The purpose of this script is to provide data for external binning algorithms, and since we don't want them to break linkage between splits and contigs, we lean the coverage data from the parent (this way two splits coming from the same contig never ends up in a different cluster unless their TNF is exceptionally different, and the soft linkage is protected).

So instead of this, you *could* add a flag, like `--use-actual-split-coverages`, and explain in the help menu why by default parent coverages are exported, and what would happen if this flag is used. Then you can conditionally read the data from `mean_coverage_splits` or `mean_coverage_contigs` table.

Best,

PR #558, opened 2017-07-13
Update anvi-export-splits-and-coverages

AstrobioMike:
adding optional flag to export 'split' coverages


meren:
I see! 

I think it would have been much better if you added a community parameter, i.e., `--splits-mode` in `anvio/__init__.py` file, and add it in the program the others are added. Would you like me to do that or would you like to do it and update your PR? :)

AstrobioMike:
Ha, i tried that at first but had a hard time finding the right __init__.py
file on my install and testing things, but I wanna give it another shot
now, I'll try first

meren:
It is the one underneath the anvi'o main module directory. You will see
crap loads of parameters listed there :)

PR #559, opened 2017-07-13
adding --splits-mode as an optional argument to anvi-export-splits-and-coverages and as a community parameter

AstrobioMike:


meren:
Yay :)

PR #590, opened 2017-09-16
HMMops filter for missing genes

meren:
For instance you have 100 genomes, and you want to focus on 5 HMMs to do a phylogenomic analyses. But some of these genomes have only 1 of those 5, some others have none. How to subset a list of genomes that would be sufficiently rich with respect to how many of those 5 they have so the phylogenomic analysis is not utterly noisy?

One way to do it is to eliminate genomes based on their level of completion. But that doesn't really work this information does not guarantee that the genes you are interested will be present in genomes you pass. 

This new parameter, `--max-num-genes-missing-from-bin` solves that issue, by letting the user be able to say "Report only from genomes that contains at least X of my Y genes".

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -432,6 +432,15 @@ D = {
                       the gene names that appear multiple times, and remove all but the one with the lowest e-value. Good\
                       for whenever you really need to get only a single copy of single-copy core genes from a genome bin."}
                 ),
+    'max-num-genes-missing-from-bin': (
+            ['--max-num-genes-missing-from-bin'],
+            {'default': None,
+             'help': "If you have a list of gene names, you can use this parameter to omit any bin (or external genome) that\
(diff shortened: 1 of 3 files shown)


rbartelme:
I think this is a good start for the flag(s), but the newly added flag, --max-number-of-genes-missing, requires some a priori knowledge about which genes from the HMM are present in bins/genomes.

I think this flag will be useful for others, but the following could be useful for preliminarily analyzing HMM-based phylogenomics:

If g = number of genomes, and n = number of times a gene from the HMM appears across all bins/genomes. Export & report all genes for which g=n. 

So if you have 30 genomes, a gene from the hmm must appear 30 times to be included in the alignment. Does that make sense? It's more of a coarse analysis than the HMMops filter. Unless I'm misinterpreting the usage of #590, which is entirely possible.

PR #608, opened 2017-10-13
attempting to edit anvi-search-functions-in-splits

AstrobioMike:
I have helped a couple people who wanted to search for functions by accession ID (like PFXXXXX/COGXXXX). I usually hack around by uploading another functions table where the accession ID is in the 'function' column also. I tried to propose a change here that would search the 'accession' column would be searched too, but i'm not sure of 2 things: 1) if you didn't do it this way for a reason; and B) if this is the correct or only thing that should be changed to actually add what i'm talking about

hopefully that makes sense!
-mike

### DIFF
--- a/anvio/dbops.py
+++ b/anvio/dbops.py
@@ -454,7 +454,7 @@ class ContigsSuperclass(object):
         for search_term in search_terms:
             self.progress.new('Search function')
             self.progress.update('Searching for term "%s"' % search_term)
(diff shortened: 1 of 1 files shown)


meren:
Sorry for the late response, Mike. I think it would be very useful to allow search terms to be searched also in the accession column, but I don't see how this change allows that :)

The actual part that returns matches still look for the search term among functions:

> function LIKE "%%''' + search_term + '''%%"

Your current change only adds the accession id to the search hits.

Does it make sense?

PR #624, opened 2017-10-31
Gene coverage values per nt

ShaiberAlon:
@meren, I need to get the coverage information for genes in samples for the mcg-classifier. I created this function in dbops so it would be available for general use. If you don't think it is necessary or if the design is not good, I could make this an helper function within the mcg-classifier code.

### DIFF
--- a/anvio/dbops.py
+++ b/anvio/dbops.py
@@ -1495,8 +1495,15 @@ class ProfileSuperclass(object):
         self.progress.end()
 
 
-    def get_gene_coverage_values_per_nt_in_samples(self, gene_callers_id, samples_of_interest=None):
-        ''' returns a dictionary with samples as keys and the nucleotide level coverage values for a gene_callers_id as a numpy vector'''
+    def get_gene_coverage_values_per_nt_in_samples(self, genes_of_interest=None, samples_of_interest=None):
+        ''' returns a dictionary with the nucleotide level coverage values
+            
+            The output format is a dictionary with the samples of interest as keys.
+            If no samples were specified, then all samples are included.
+            Each dictionary entry is another dictionary with the genes_of_intereset as keys,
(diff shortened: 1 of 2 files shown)


meren:
I think this is good, Alon. There is some redundancy with `init_gene_level_coverage_stats_dicts`, such as whether genes are called, or splits considered gene calls. Whenever a non-atomic, logical piece of code occurs in two places redundantly, it is almost always a great indication of the need for improvement :) Maybe I can later take a look at this.

ShaiberAlon:
Hi Meren,

Thanks for looking anf merging.
I realized the redundancy, and moreover, I believe there is some further
redundancy between `init_gene_level_coverage_stats_dict` and the
mcg-classifier, and I would like to discuss it in our next meeting.

Thank you,

Alon

ShaiberAlon:
Hi Meren, 

What do you think about the following solution:
instead of having a separated function to get the nt level gene coverage, maybe we can add a flag `populate_nt_level_coverage`, with default `False` to `init_gene_level_coverage_stats_dicts`. Then, when given a `True`, it will add a key `gene_coverage_per_position` (and maybe also `non_outlier_positions`) to the dictionary `gene_level_coverage_stats_dict`?

Both `gene_coverage_per_position`, and `non_outlier_positions` are already available as numpy arrays when running `init_gene_level_coverage_stats_dicts`, and I need it for the mcg-classifier. Let me know what you think. It would be an easy change, and would remove all the redundancy.

While you are at it, here is a separated, but related issue: I create a `sample_level_coverage_stats_dict`, with all the same fields as in `gene_level_coverage_stats_dict`, but for sample level. I don't have any suggestion yet, but I thought it would be something for you to keep in mind when considering the design of this function.

Thank you!

PR #633, opened 2017-11-06
add gene_coverage_per_position to gene_level_coverage_stats_dict

ShaiberAlon:
And also add non_outlier_positions.
@meren, if you accept this change then we can delete
`get_gene_coverage_values_per_nt_in_samples`, and thus address your
comment in #624.


meren:
I am fine with this solution :) thank you very much, alon!

ShaiberAlon:
Great! Thank you very much.

PR #660, opened 2017-12-06
Pandas variabilityops

ekiefl:
In comparison to master, here are the differences I have documented:

1. Entries are ordered by unique_pos_identifier. For a given unique_pos_identifier, entries are ordered alphabetically by sample_id.

2. Ties from consensus values are now resolved alphabetically.

3. Competing item logic is now logical. There are three rules that are always upheld.
     - In the case of ties for the most common, or second most common item, they are always resolved alphabetically. E.g. a tie between Ala and Trp for the second most common item goes to Ala. 
     - The competing items are always arranged alphabetically, regardless of which is most common. In the previous example Ala was second most common, and if Ser was the most common, competing_aas will still be AlaSer
     - If the coverage of the most common item is equal to the total coverage, the item is paired with itself, e.g CysCys.

4. To get a first glance at variation, we display competing_nts during inspect mode in the interactive interface, which means they are stored. I don't know how the competing_nts are computed during this process, but after looking at 3 datasets (Infant Gut, E. Faecalis; TARA Oceans, HIMB083; and Mushroom Spring, Synechococcus), there were no "N" counts (--engine NT), which makes me think that consensus and competing_nts are calculated without consideration of "N" values. In contrast, the variability table is not prejudiced to N, and treats it like any other item in self.items. This means it can be the consensus value, or be one of the items in competing_nts, etc.

5. `STP` from --engine AA and `N` from `--engine NT` can be competing items or consensus values. This may or may not be different behavior.

6. For kullback_leibler_divergence_normalized, it used to be the case that if one or more of the samples for a given unique_pos_identifier had 0 coverage, the value was infinite. I now do something that is mathematically equivalent to simply discarding that sample for that unique_pos_identifier, i.e. I give it zero weight and calculate what the entropy is without that sample.

7. In my last update weighted SSM scores were not identical between master and pandas_variability. This is just due to small numerical error, when rounding the results to within 3 decimal places everything is identical. So this is equally as correct as master.

8. Any `--quince-mode` and `--min-coverage-in-each-sample 0` can produce entries with 0 coverage. In these cases `n2n1ratio` and `departure_from_consensus` used to be `-1.0`. Now, they resolve to `0.0`.

9. All `np.nan`s `-np.inf`s and `np.inf`s are set to the empty string `""` when exporting to tab-delimited file. This choice was to increase compatibility between how other programs handle things like "NaN", "inf", "-inf", etc.

10. `recover_base_frequencies_for_all_samples` may be partially vectorizable, but for now is definitely the slowest part of the code.

11. `compute_comprehensive_variability_scores` is now split into metrics that require `--quince-mode`, and those that don't. Before, some metrics didn't actually require `--quince-mode`, but were wrapped in  `compute_comprehensive_variability_scores` which did require `--quince-mode`.

12. It was discovered in the master version that `filter_based_on_scattering_factor` didn't have its intended behaviour. It still doesn't, but now a `ConfigError` is invoked if the function is called. I will fix this.

Aside from the above differences, the identicalness of the table between pandas_variability and master has been verified on the INFANT GUT for the following tables:

```
anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine AA -o VARIABILITY_TESTS/PANDAS_AA --quince
anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine AA -o VARIABILITY_TESTS/PANDAS_AA_DFR --quince -r 0.1 -z 0.9
anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine AA -o VARIABILITY_TESTS/PANDAS_AA_DFC --quince -j 0.1 -a 0.9
anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine AA -o VARIABILITY_TESTS/PANDAS_AA_MINCOV --quince --min-coverage-in-each-sample 20
anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine AA -o VARIABILITY_TESTS/PANDAS_AA_X --quince -x 2
anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine AA -o VARIABILITY_TESTS/PANDAS_AA_HEAVYPARAMS --quince --min-coverage-in-each-sample 1 -r 0.05 -z 0.95 -j 0.05 -a 0.95 -x 2

anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine NT -o VARIABILITY_TESTS/PANDAS_NT --quince
anvi-gen-variability-profile -p DATA/PROFILE21.db  -c DATA/CONTIGS.db -C merens -b E_facealis --engine NT -o VARIABILITY_TESTS/PANDAS_NT_HEAVYPARAMS --quince --min-coverage-in-each-sample 1 -r 0.05 -z 0.95 -j 0.05 -a 0.95 -x 2
```

### DIFF
--- a/anvio/db.py
+++ b/anvio/db.py
@@ -6,6 +6,7 @@
 
 import os
 import sqlite3
+import pandas as pd
 
 import anvio
 import anvio.filesnpaths as filesnpaths
@@ -255,6 +256,50 @@ class DB:
                 results_dict[row[0]] = entry
 
         return results_dict
+
+
+    def get_table_as_dataframe(self, table, table_structure=None, columns_of_interest=None, keys_of_interest=None, omit_parent_column=False, error_if_no_data=True):
+        """
+        get_table_as_dict() uses the first column as the key in the resulting
+        dictionary. For pandas DataFrames there are two reasonable design
+        approaches. The first mimics this approach and uses the first column as
+        the index of the DataFrame. The approach I take instead is to keep the
+        first column as a column in the DataFrame (it is afterall, a column)
+        and use numerical indices for the DataFrame.
+        """
+        if not table_structure:
+            table_structure = self.get_table_structure(table)
+
+        columns_to_return = table_structure
+
+        if omit_parent_column:
+            if '__parent__' in table_structure:
+                columns_to_return.remove('__parent__')
+                table_structure.remove('__parent__')
+
+        if columns_of_interest:
+            for col in table_structure[1:]:
+                if col not in columns_of_interest:
+                    columns_to_return.remove(col)
+
+        if len(columns_to_return) == 1:
+            if error_if_no_data:
+                raise ConfigError("get_table_as_dataframe :: after removing an column that was not mentioned in the columns\
+                                    of interest by the client, nothing was left to return...")
+            else:
+                return {}
+
+        if keys_of_interest:
+            keys_of_interest = set(keys_of_interest)
+
+        rows = self.get_all_rows_from_table(table)
+        results_df = pd.DataFrame(rows, columns=table_structure)
+
+        if keys_of_interest:
+            results_df = results_df.loc[results_df.index.isin(keys_of_interest)]
+        results_df = results_df.loc[:, columns_to_return]
+
+        return results_df
 
 
     def get_some_rows_from_table_as_dict(self, table, where_clause, error_if_no_data=True, string_the_key=False):
--- a/anvio/utils.py
+++ b/anvio/utils.py
@@ -17,6 +17,9 @@ import configparser
 import multiprocessing
 import urllib.request, urllib.error, urllib.parse
 
+import numpy as np
+import pandas as pd
+
 from email.mime.text import MIMEText
 from collections import Counter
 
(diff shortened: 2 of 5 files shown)


meren:
Thank you very much for all the work and the detailed report, Evan.

ekiefl:
I remember within the first couple of weeks of my rotation I asked why the variability table was not a pandas DataFrame, and joked that I could change it but that it would take me 6 months. Actually it just took 10 days.... (+8 months of training) :)

ekiefl:
Note to self:

If we come back to this, the working directory of my aforementioned tests is at `/Users/evan/Academics/Research/Meren/INFANT_GUT/`

PR #661, opened 2017-12-07
Blast aa sequences against gene calls

ekiefl:
I wrote this awhile ago because I wanted to test whether Sean's photoproteins were found in Synechococcus. Julie also wanted to know whether some specific genes were found in her phage. That's already two people. If this is anything like the Tiger Woods sex scandal, more and more people will start coming forward, so I think it's time to add this function.

There is a lot to do on this front, for example visualizing blast results in the interactive interface. But for now, this will be sufficient for people asking the question: "Are homologs of proteins X, Y or Z found in my contigs?"


meren:
This is a good direction, but I think we could start with a much cleaner design. let's talk about this PR tomorrow and make a plan.

Issue #707, opened 2018-01-22
error in display-pan after using anvi-import-misc-data


ShaiberAlon:
I imported the following table:
https://github.com/merenlab/Oral-MGx/blob/master/scripts/metapangenomics/samples-information.txt

to a pangenome, and then running `anvi-display-pan`, I got the following error:
```
Traceback (most recent call last):
  File "/Users/alonshaiber/github/anvio/bin/anvi-display-pan", line 74, in <module>
    d = interactive.Interactive(args)
  File "/Users/alonshaiber/github/anvio/anvio/interactive.py", line 106, in __init__
    self.layers_order_data_dict = TableForLayerOrders(self.args).get(self.layers_additional_data_keys, self.layers_additional_data_dict) if a_db_is_found else {}
  File "/Users/alonshaiber/github/anvio/anvio/tables/miscdata.py", line 256, in get
    return self.update_orders_dict_using_additional_data_dict(d, additional_data_keys, additional_data_dict)
  File "/Users/alonshaiber/github/anvio/anvio/tables/miscdata.py", line 276, in update_orders_dict_using_additional_data_dict
    order_data_dict['>> ' + data_key] = {'newick': None, 'basic': ','.join([t[1] for t in sorted(layer_name_layer_data_tuples)])}
TypeError: '<' not supported between instances of 'NoneType' and 'str
```

ShaiberAlon:
This reproducible using this:
https://github.com/merenlab/Oral-MGx/blob/master/temp.tar.gz

meren:
Nice catch, Alon. I am looking into this :)

ShaiberAlon:
I see what the problem is. My additional layers information contains columns that are empty for some of the samples. I guess sorted doesn't like `None`, and would have preferred that these would have been read as either an empty string or as a string `"None"` or `"NONE"`.

meren:
Can you pull and test this on `master`, alon? if it looks OK please close the issue :)

ShaiberAlon:
Thank you!

PR #708, opened 2018-01-22
deal with empty data

ShaiberAlon:
in order to create a default ordering with additional information we
need to deal with missing data. this solves #707


meren:
this solution will cause an error if the data type is not str. this needs to be aware of the data type of incoming data. I will propose something soon.

PR #742, opened 2018-02-14
warning the user

ShaiberAlon:
@meren, please look at this and let me know if you agree with my
diagnosis.


meren:
The following change doesn't make much sense:

``` diff
-        if args.profile_db and not args.collection_name:
+        if args.profile_db != args.collection_name:
```

Why? :)

ShaiberAlon:
@meren, it is not fair to slander me twice for the same error (https://merenlab.slack.com/archives/C08H4G30V/p1518650736000375?thread_ts=1518648824.000531&cid=C08H4G30V)

On the other hand it is me that didn't fix it back when you commented....

ShaiberAlon:
Ok, fixed it, I'll leave it to you to merge, just in case I'm being foolish again...

PR #745, opened 2018-02-16
Bug fix for color coding conservation of genomes with multiple genes

mahmoudyousef98:
Fixed a bug where only the last gene in a genome with multiple genes will be color coded


meren:
Thanks, @mahmoudyousef98 👍

PR #788, opened 2018-03-19
Workflows

ShaiberAlon:
I think it is stable enough to be merged with master


ShaiberAlon:
@meren, I tested it on barhal to see that the `--additional-params` works as expected and allows to use (for example): `--additional-params --cluster 'clusterize -log {log} -n {threads}' --resources nodes=40 --jobs 10`, and it worked smoothly. So this is ready to use.

meren:
This is very happy news, Alon! Thank you very much.

ShaiberAlon:
:-) Thanks!

Issue #775, opened 2018-03-07
Memory requirements of anvi-export-gene-coverage-and-detection when working with big projects


meren:
An investigation of memory needs for gene coverage recovery tasks is necessary.

Here is a report from Les:

https://groups.google.com/forum/#!msg/anvio/vARl21bw6o8/GRd9Jvo9CAAJ

ozcan:
I have tested this on both Infant Gut and one of the big FMT files, results are exciting.

FMT (black old, blue new)
![screen shot 2018-04-19 at 10 18 28 am](https://user-images.githubusercontent.com/940884/39001634-5a9c3a30-43bc-11e8-9cf0-345cf4f6ce92.png)

INFANT GUT (black old, blue new)
![screen shot 2018-04-19 at 10 31 24 am](https://user-images.githubusercontent.com/940884/39001888-ed9aed0e-43bc-11e8-9a26-978b4f843ecd.png)

File contents are the same, except the new one does not order by gene_callers_id, it follows to split order.

I will open a PR once I check other places may be affected by these changes.

meren:
These look extremely encouraging, @ozcan!

PR #810, opened 2018-04-19
Reduce memory usage of anvi-export-gene-coverage-and-detection, Related #775

ozcan:


meren:
Yay :) Thank you very much!

PR #836, opened 2018-05-28
anvi-get-short-reads-from-bam gets beefed up

ekiefl:
Tested on unpaired and paired bam files, w/ all combinations of user input I could think of. 

Multiple files separated by `,` for `-o` is a little ugly, but documentation is good with an example. Also, code catches funny instances like `file1,file2,`, `,file1,file2,file3` and `file2,,file2,file2`

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -745,15 +745,12 @@ D = {
                       your short reads, as well as the length of the gene you are targeting.\
                       The default is %(default)d nts."}
                 ),
-    'split-by-forward-and-reverse': (
-            ['-Q', '--split-by-forward-and-reverse'],
+    'split-R1-and-R2': (
+            ['-Q', '--split-R1-and-R2'],
             {'default': False,
              'action': 'store_true',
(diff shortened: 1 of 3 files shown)


meren:
Hey Evan. Great job. Please check my changes. 

In mini test this guy reports no R1s an R2s. If it was working for the BAM files you were playing with, can you please test it and make sure I didn't break something?


Thanks!

ekiefl:
> Please check my changes

Nice changes it is much cleaner. `gzip` was removed during your commits, and I assume that was a mistake so I re-added it. I also added progress during storing reads because it looks very un-anvio-like to not have the cyan status bar during a step that takes more than a few seconds.

> In mini test this guy reports no R1s an R2s. If it was working for the BAM files you were playing with, can you please test it and make sure I didn't break something?

I experienced that before your commits and assumed that those reads were not paired. If they are, then we may have some consistency issues with pysam? It works on TARA:

image

If you're happy let's merge this.

meren:
Thanks! Looks great!

PR #929, opened 2018-08-06
Add virsorter rules to Snakemake MG workflow

jarrodscott:
A first attempt at VirSorter rules in Snakemake MG workflow

### DIFF
--- a/anvio/workflows/contigs/Snakefile
+++ b/anvio/workflows/contigs/Snakefile
@@ -63,7 +63,6 @@ if run_taxonomy_with_centrifuge:
     if not M.get_param_value_from_config(["centrifuge", "db"]):
         raise ConfigError("If you plan to run centrifuge, then you must "\
                           "provide a path for the centrifuge db in the "\
-
                           "config file. See documentation for more details.")
 
 # sanity check for virsorter db
@@ -382,7 +381,7 @@ rule anvi_script_run_eggnog_mapper:
     shell: "anvi-script-run-eggnog-mapper -c {input.contigs} --annotation {input.eggnog_output} {params.use_version}  >> {log} 2>&1"
 
 if run_classify_with_virsorter:
-    rule run_virsorter: 
+    rule run_virsorter:
         '''
             Search for viral sequences using VIRSorter on your contigs file.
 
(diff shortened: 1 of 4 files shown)


ShaiberAlon:
@jarrodscott,

I suggest that the outputs of the misc data import will be done to individual profile databases. See the way the "percent reads mapped" is imported::
This is the rule where it's imported to single profiles:
https://github.com/merenlab/anvio/blob/6a02671f24f0c1c268626d56e2b9b736f6167f3a/anvio/workflows/metagenomics/Snakefile#L980

And this is what makes snakemake understand that these need to be imported:
https://github.com/merenlab/anvio/blob/6a02671f24f0c1c268626d56e2b9b736f6167f3a/anvio/workflows/metagenomics/Snakefile#L827

Importing the collection needs to be done after merging, so it needs to be something separate. You will just add a rule for importing a collection, and then you will need some output for this rule (probably a flag file: http://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#flag-files), and then just add this output file as an optional input to the "target" rule, `metagenomics_workflow_target_rule` (so that snakemake would know to run it). For an example of an optional input to the target rule, look at `qc_report`(https://github.com/merenlab/anvio/blob/6a02671f24f0c1c268626d56e2b9b736f6167f3a/anvio/workflows/metagenomics/Snakefile#L218).

The functional annotation should be configured similarly to the COGS annotation in the contigs workflow.

I hope this is helpful. Feel free to contact me whenever.

ShaiberAlon:
Hi @jarrodscott, I forgot to mention that this:
https://github.com/jarrodscott/anvio/blob/526a995b66dc5afe6179318495a87ad27f07d6f2/anvio/workflows/contigs/Snakefile#L422
Will not work when `references_mode` is used, since then the user could skip reformat fasta. The correct way, would be to get the fasta file the same way that anvi_gen_contigs_database rule gets it:
https://github.com/jarrodscott/anvio/blob/526a995b66dc5afe6179318495a87ad27f07d6f2/anvio/workflows/contigs/Snakefile#L232
i.e. using the `get_fasta` function as an input, you can learn more from the snakemake doc [here](http://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#functions-as-input-files).

meren:
This branch now has too many conflicts. I hate the idea of closing it, but I am not sure what is the best course of action here either :(

What do you think, @jarrodscott?

jarrodscott:
Hi @meren 

Indeed, this is a little long in the tooth :) and since there is a [new version of Virsorter](https://github.com/jiarong/VirSorter2) that has undergone major changes, I think we should close this issue. That said, the output files from Virsorter2 seem easier to parse and the pipeline is pretty user friendly. If Virsorter2 continues to be supported, it may make sense to incorporate it into anvi'o somewhere, if not in the snakemake workflow, then perhaps as a standalone function.  If this is something you are interested in, I am certainly willing to work on it :)

meren:
That sounds like a good plan, @jarrodscott, I will then close this PR as a part of my effort to tie all loose ends as we approach to `v7`.

I would be very interested to see a workflow for Virsorter2. Perhaps we can think about this together after January? :)

PR #932, opened 2018-08-11
Align date better with academic time scales.

ctb:
cc @AstrobioMike and https://twitter.com/AstrobioMike/status/1027724624052674560 specifically.


meren:
The community is pushing for more realistic and modest promises? We shall allow it.

PR #935, opened 2018-08-14
change scikit-learn requirment

ShaiberAlon:
With the new brew and new python versions pip install
scikit-learn==0.19.1 fails, but `pip install scikit-learn` works (which
is version `0.19.2`)

Homebrew 1.7.1
Python 3.7.0

@ozcan, @meren, do you have an objection to this change?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy
 cython
 bottle==0.12.13
-pysam==0.14.1
+pysam>=0.14.1
 ete3==3.1.1
 scipy>=0.13.3
 scikit-learn>=0.19.1
@@ -16,4 +16,4 @@ pandas>=0.20.1
 matplotlib>=2.2.0
 pyani==0.2.7
 statsmodels==0.9.0
-snakemake==4.7.0
+snakemake>=4.7.0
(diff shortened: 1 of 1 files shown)


meren:
Hey Alon,

Sorry for the late response. I am fine with these version changes, but it is best if we stick with explicit version numbers rather than using `>=` for whatever is available. It creates really ugly problems in the long run and we try to avoid those as much as possible :/

ShaiberAlon:
I will not merge it yet, since currently, `anvi-interactive` doesn't work properly on my computer.... I don't why that is, and I have ignored it for now, since I haven't needed the interface on my local.
Once I figure this out, and verify that it has nothing to do with any of the updated versions, then what I suggest is updating the requirements for these libraries (i.e. pysam, snakemake, pandas, and scikit-learn) to their latest version with `==`.

To my understanding, the only problem with currently updating the requirement to be `==` is that it would require people to upgrade to python 3.7 (am I right?), and currently that comes with a lot of problems (e.g. the problem with `datrie` that I specify above).

meren:
Let's talk about this later :)

meren:
(putting a note here so we don't forget)

PR #944, opened 2018-08-24
Explicitly define default for --hmm-source.

blankenberg:
Should have no functional impact. 

However, it blew up my Galaxy Tool generator when building help text. 

I can apply an easy one-line fix to my tool generator to default to `None` for any non-defineds, but this is the only instance of a value being referenced in the help but not explicitly declared.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -487,6 +487,7 @@ D = {
     'hmm-source': (
             ['--hmm-source'],
             {'metavar': 'SOURCE NAME',
+             'default': None,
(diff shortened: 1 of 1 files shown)


meren:
Apologies for the oversight, and thanks for your fix!

blankenberg:
No worries, thanks for the very quick merge!

Change in my code was:
```help = help % self.info_dict```
to
```help = help.format(**self.info_dict)```
So really a super-easy fix on my end, but then I started an internal debate as to whether I should want to catch these sort of things...

meren:
I think it is always better to find the lowest levels of organization to enforce explicit descriptions. So your change is much appreciated :)

PR #956, opened 2018-09-03
Amy Willis' functional enrichment score

### LINKED ISSUE #904: Implement an appropriate statistical test for functional enrichment
See thread here: http://merenlab.org/2016/11/08/pangenomics-v2/#comment-3963129765.

ShaiberAlon:
closes #904.

Here we implemented the enrichment score that was suggested by @adw96. More details are here (including a short explanation from Amy):
http://merenlab.org/2016/11/08/pangenomics-v2/#making-some-sense-of-functions-in-your-pangenome

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -396,14 +396,28 @@ D = {
              'help': "Takes a value between 0 and 1, where 1 means that only functions that occur in all members of\
                       one of the compared groups will be included in the output. Default is %(default).1f."}
                 ),
+    'false-detection-rate': (
+            ['--false-detection-rate', '--FDR'],
+            {'metavar': 'PORTION',
+             'default': 0.05,
+             'type': float,
+             'help': "Takes a value between 0 and 1, to determine the false detection rate that will be used \
+                      for the Benjamini–Hochberg procedure. Default is %(default).1f."}
+                ),
+    'core-threshold': (
+            ['--core-threshold'],
+            {'metavar': 'PORTION',
+             'default': 1,
+             'type': float,
+             'help': "Takes a value between 0 and 1, where 1 means that only functions occuring in all genomes \
+                     of a group would be considered as core functions of that group. Default is %(default).1f."}
+                ),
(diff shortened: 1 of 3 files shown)


anvio/summarizer.py
```diff
@@ -1902,3 +1891,39 @@ def store_data_in_file(self, output_file_name_posfix, content):
         output_file_obj.write('%s\n' % content)
         output_file_obj.close()
 
+
+def benjamini_hochberg_procdure(p_values, fdr=0.05):
```
meren: so this doesn't do anything? ;)

By the way, this function (and `get_z_test_statistic`) would be much better if they were . in utils, in my opinion.

anvio/__init__.py
```diff
@@ -396,14 +396,28 @@ def get_args(parser):
...
+            ['--false-detection-rate', '--FDR'],
+            {'metavar': 'PORTION',
+             'default': 0.05,
+             'type': float,
+             'help': "Takes a value between 0 and 1, to determine the false detection rate that will be used \
+                      for the Benjamini–Hochberg procedure. Default is %(default).1f."}
+                ),
+    'core-threshold': (
+            ['--core-threshold'],
+            {'metavar': 'PORTION',
```
meren: metavar is much more useful when it describes the data type. integer, float, name, file path, etc :)

anvio/summarizer.py
ShaiberAlon: Done :-)
https://github.com/merenlab/anvio/commit/c182b3ee6f4b5f6ba4ada33dfa2f76d6aee27cf7
https://github.com/merenlab/anvio/commit/a704917f417b104c5de9666c260c4b065cc1bddd

Thank you!

anvio/__init__.py
ShaiberAlon: Also done :-)
https://github.com/merenlab/anvio/commit/9e2c33c7c1e8dc99a938b19711a1e686127d10ff

PR #960, opened 2018-09-04
Fix import percent reads mapped

### LINKED ISSUE #959: import_percent_reads_mapped fails when SNVs are skipped
In the metagenomics workflow, when SNVs are not profiled, then the step to import the percent reads mapped fails.

ShaiberAlon:
Closes #959.
Thank you @FlorianTrigodet for bringing this problem to my notice. Now it should be fixed.

### DIFF
--- a/anvio/workflows/metagenomics/Snakefile
+++ b/anvio/workflows/metagenomics/Snakefile
@@ -1027,13 +1027,13 @@ rule import_percent_of_reads_mapped:
                     anvi-import-misc-data {output.total_num_reads_txt} \
                                           -p {input.profile} \
                                           -t layers \
-                                          --just-do-it
+                                          --just-do-it >> {log} 2>&1
 
                     # get the layers additional data table, which has two columns
                     # now.
                     anvi-export-misc-data -p {input.profile} \
                                           -t layers \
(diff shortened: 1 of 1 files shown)


anvio/workflows/metagenomics/Snakefile
```diff
@@ -1029,27 +1029,35 @@ rule import_percent_of_reads_mapped:
...
-                             {{print $0 "\tpercent_mapped"}} \
-                          else \
-                              {{print $0 "\t" ($2 * 100 / $4)}}}}' \
-                        {output.layers_additional_data} > {output.layers_additional_data_updated}
+                                          -o {output.layers_additional_data} >> {log} 2>&1
+                  """)
+
+            additional_data = pd.read_csv(output.layers_additional_data, sep='\t', index_col=None)
+            if 'total_reads_mapped' not in additional_data.columns:
+                raise ConfigError('The layers additional data for %s doesn\'t contain a column
```
meren: ignorant question, is this really going to work without the new-line escapes (`/`)?

ShaiberAlon: I don't think it would :-) luckily if someone got there, they deserve an error (just not the error that they will get)

PR #965, opened 2018-09-07
modifying anvi-script-genbank-to-external-gene-calls

AstrobioMike:
not writing out genes to functions table (for importing functions) if no annotation for the gene call, rather than including them with just "hypothetical protein" listed


meren:
Thank you mister!

PR #977, opened 2018-09-14
Homogeneity Index

mahmoudyousef98:
- Implemented an algorithm for calculating gene cluster homogeneity index, used by default in pangenomics (anvi-pan-genome)
- Created a script anvi-compute-gene-cluster-homogeneity that will compute and return the homogeneity indices of one or more gene clusters of a pangenome
- Compatible with multiprocessing/threading


meren:
Hey @mahmoudyousef98, thank you very much for this PR. I made a bunch of changes, and it would be great if you can take a look at the commits log and changes to see what I've done. Using 'tig' from your command line could help you go through these changes rather quickly by giving you an interface like this:

image

I will open an issue to suggest two potential next steps :)

mahmoudyousef98:
All looks good! Especially liked what you did with rewriting to existing databases. Thank you!

Issue #979, opened 2018-09-15
Better utilizing and disseminating gene cluster homogeneity indices for pangenomics


meren:
Thanks to @mahmoudyousef98's excellent addition to the codebase (see  #977), the anvi'o pangenomic workflow has two new indices to investigate within gene cluster homogeneity. 

There are at least two necessary steps going forward to make this powerful addition available for filtering gene clusters, and including some information about it in our documentation.

## Documentation

We need to update our tutorial here:

http://merenlab.org/2016/11/08/pangenomics-v2/

Perhaps we can add a new section dedicated to homogeneity indices right after this one:

http://merenlab.org/2016/11/08/pangenomics-v2/#inspecting-gene-clusters

It will help others to better understand what those new layers are.

## Implementing new filters

We currently allow users to filter their gene clusters based on various metrics. This is done through a function that is used both by the program `anvi-get-sequences-for-gene-clusters` (see the "ADVANCED FILTERS" section in the help menu), and through the user interface:

![image](https://user-images.githubusercontent.com/197307/45590969-d2aa7c80-b90a-11e8-98f9-681a685d4051.png)

If a pangenome includes functional or geometric homogeneity indices in additional misc data tables, then we could also make available these filters both through the interface and the command line program. This way a query like this could be possible: "_Give me all gene clusters that occurs in all genomes with a geometric homogeneity >= 1.0, and functional homogeneity of <1.0_". This way a user can get all non-identical core gene clusters at once.

What do you think @mahmoudyousef98?

mahmoudyousef98:
Sounds good with me! I'll start working on it

meren:
Self note: pangenomes without homogeneity indices break `anvi-display-pan` and we need to address this before the minor release.

meren:
The documentation is updated, and the reminder above is fixed with #991 and following commits.

PR #981, opened 2018-09-17
Filtering Gene Clusters by Homogeneity

mahmoudyousef98:
Per issue #979:

- Added ability to filter by minimum and maximum functional and geometric homogeneity indices
- Update pangenome interface to allow for filtering by homogeneity
- Added argument parameters to _anvi-ger-sequences-for-gene-clusters_ which have also been added to the list of global parameters

I had to change the type of the filters parameters input from the interactive end to type _float_ because the homogeneity indices are between 0 and 1 rather than 0 and 100. I retained the type checker in the dbops function _filter_gene_clusters_from_gene_clusters_dict_ so that the previously defined filters are not affected in any way


meren:
Thanks!

meren:
(although we need to find a way to make sure pangenomes without homogeneity indices do not break).

Issue #945, opened 2018-08-26
anvi-import-collection error (entries are not unique in some tables)


nick-youngblut:
I'm running anvio version:

```
Anvi'o version ...............................: margaret (vunknown)
Profile DB version ...........................: 29
Contigs DB version ...........................: 12
Pan DB version ...............................: 12
Genome data storage version ..................: 6
Auxiliary data storage version ...............: 2
Structure DB version .........................: 1
(/ebio/abt3_projects/software/dev/llmga/.snakemake/conda/3b8a2f3c)
```

...and I'm getting the following error when running `anvi-import-collection` on a set of bins generated with `maxbin2`:

```
Contig/split name conversion .................: 56166 contig names converted into 56166 split names.
Item names in input ..........................: 56,166
Num bins in input ............................: 190
Items in profile database ....................: 57,501
Item names shared between input and db .......: 56,166

WARNING
===============================================
1335 item(s) that were in the database, but were not in the input file, will not
be described by any bin in the collection MaxBin2_low_prob. That is totally
fine, but anvi'o hopes that you are aware of that. This means you have more
things in your database than the number of things your input file describes.
Here is an example of something that is in your database but not in any bin in
your input file: coassemble_26409_split_00001.



Config Error: This is one of the core functions of anvi'o you never want to hear from, but
              there seems to be something wrong with the table 'collections_bins_info' that
              you are trying to read from. While there are 516 items in this table, there are
              only 326 unique keys, which means some of them are going to be overwritten when
              this function creates a final dictionary of data to return. This may be a
              programmer error when the data was being inserted into the database, but needs
              fixin' before we can continue. If you are a user, please get in touch with
              anvi'o developers about this error. If you are a programmer, you probably did
              something wrong :(
```

For generating the contig db, I used the `--skip-mindful-splitting` option, since I had already split my contigs prior to the import. The `maxbin2` binning was done on the split contigs, so all of the contig naming should be the same. As far as I can tell, none of my contigs are named with "*_split_00001" in my contigs database or in my maxbin2 bins. 

For the `anvi-import-collection` command, I'm using `--contigs-mode`, which I thought prevented renaming of the contigs. However, the log for the import states: 

```
Contig/split name conversion .................: 56166 contig names converted into 56166 split names.
```

[...]

PR #989, opened 2018-09-19
Check sqlite journal file, Related #945

ozcan:
This pull request makes DB class to make sure there is no other instance of same database open for writing. `.journal` file is only created by sqlite if there is something waiting for commit.

run_all_tests and other tests seemed fine.


meren:
this looks simple and elegant to me. thank you very much, ozcan!

PR #554, opened 2017-07-10
Issue 551: allow to export fast contig in GFF3 format

gkmngrgn:
Hello reviewers,

I added a line to our test file, but I'm not sure that is the expected output. Could you review and give me feedback to continue fix and close this ticket?

Thank you very much.

Issue #551, opened 2017-06-30
Export of gff and/or gbk files from Anvio 


rbeinart:
Hello,

It would be great to have the option to export the Prodigal annotations done through Anvio as gff and/or gbk files. Even better if the annotations that are done in Anvio or imported into Anvio could be exported as part of this file.

Thank you!
-Roxanne-

meren:
`anvio.dbops.ContigsSuperclass` has a member function `gen_FASTA_file_of_sequences_for_gene_caller_ids` that is primarily used by the program `anvi-get-dna-sequences-for-gene-calls` to export gene calls in FASTA format.

I guess a reasonable strategy would be to add another function in `anvio.dbops.ContigsSuperclass` like `gen_GFF3_file_of_sequences_for_gene_caller_ids`, and add a flag to the `anvi-get-dna-sequences-for-gene-calls` (something like `--export-gff3`) to take care of this.

We may also need to consider coming up with a better design. Because the way `anvi-get-dna-sequences-for-gene-calls` works is much different than the way `anvi-get-aa-sequences-for-gene-calls` works. But this could be a discussion for later :)


2 cents.

gkmngrgn:
@meren Aah, sorry for removing whitespaces. It's my editor's bullshit. Can you do a review for my method? I didn't test this code yet, but I think to progress like this. Thanks.

meren:
Hi @gkmngrgn I just realized something :/ Now I can do this, and get a proper GFF file:

    anvi-get-dna-sequences-for-gene-calls -c CONTIGS.db -o sequences.gff --export-gff

While looking at the output and reading more about GFF, I am realizing that the right place to add this flag is `anvi-export-gene-calls` :( It is OK for `anvi-get-dna-sequences-for-gene-calls` to report only FASTA files for gene sequences (as the name suggests). But currently `anvi-export-gene-calls` only exports a TAB-delimited file only anvi'o recognizes for gene calls, and adding the `--export-gff` flag for that program makes much more sense.

This one,

```
anvi-export-gene-calls -c CONTIGS.db -o gene_calls.txt
```

Produces this one:

```
head gene_calls.txt
gene_callers_id	contig	start	stop	direction	partial	source	version
0	204_10M_MERGED.PERFECT.gz.keep_contig_878	0	933	f	1	prodigal	v2.6.2
1	204_10M_MERGED.PERFECT.gz.keep_contig_878	1113	1677	f	0	prodigal	v2.6.2
[...]

PR #997, opened 2018-09-27
Exported gff3 files are now formatted correctly

brymerr921:
The files exported by `anvi-get-sequences-for-gene-calls -c CONTIGS.db --export-gff3` were incorrectly formatted.
This change addreses issues #551 and #554. See gff3 format at: https://uswest.ensembl.org/info/website/upload/gff3.html
Within the lines modified by this commit, ideally the second column ("source") would be populated by the gene caller (e.g. "prodigal") but I don't know how to make that change.


brymerr921:
Sounds great. I agree -- fine for now. Thank you!!!

PR #1022, opened 2018-10-25
Genes DB

meren:
I think this is ready for testing. 

We need to test this both with MCGC and `anvi-interactive` in `--gene-mode`.

## Summary

Genes DB is a new anvi'o database to store gene-level coverage stats for a given collection and bin. It tracks splits in a collection, so if the user changes the contents of a bin (i.e., through `anvi-refine` or `anvi-import-collection`), the available gene db will not be used to recover data (instead of an error, a blank one will be generated, and the old one will sit there like an asshole). It also tracks the critical parameters to recover gene-level coverages and determination of outlier nucleotide positions per gene. If any of those changes, anvi'o will complain, and tell the user the stored parameters and new parameters do not match, and they need to remove the current genes database and let anvi'o recompute it.

After a blank database is generated, anvi'o will fill in everything, and will re-use it as long as it is possible. 

Genes DB can also store collections and states, however this PR does not include an `anvi-summarize`, or `anvi-export-collection` hooks. Perhaps when we need them we will look into implementing them.


ShaiberAlon:
This is great! but how do I use this?

meren:
```
anvi-interactive -p PROFILE.db -c CONTIGS.db -C COLLECTIN -b BIN --gene-mode
```

:)

For MCGC you don't need to do anything.

ShaiberAlon:
Oh got it!

But why the weird name: `CONCOCT-Bin_1-b6ae98be.db`? Why do we need to hash at the end of the name?

meren:
No more hash at the end of genes db file names. I think we are kinda ready for `master`.

PR #1025, opened 2018-11-01
Ref based short read removal

### LINKED ISSUE #1011: Add reference-based short read removal to the anvi'o snakemake metagenomics workflow
I chose the following approach:
**after quality filtering**, each fastq file would be:
1. mapped using bowtie2 to the references in `references-for-removal.txt`
2. output sam file would be converted to a bam file, and will be either deleted at the end of the process or saved (depending on what the user specified in their config file).
3. get a list of reads using `samtools view BAM_FILE | cut -f 1` and remove the reads from the fastq file (maybe using `iu-remove-ids-from-fastq` (if this is addressed: https://github.com/merenlab/illumina-utils/issues/19)).

## Some notes on configurations and other internal debates between Alon and himself
By default the contamination will be removed using the human genome, and there will be a command to setup the human genome in a known (i.e. known to anvi'o) location (maybe `anvio/data/misc/human_genome/`?). Otherwise, the user could provide a `references-for-removal.txt` with one line and two columns: reference name, path to reference fasta file. That way if you use mouse metagenomes for example you could provide a mouse reference genome for contamination removal. Multiple contamination removal references would be allowed.
We will add an example/examples to our tutorial of how to use this feature with the human reference genome GRCh38 `ftp://ftp.ncbi.nlm.nih.gov/refseq/H_sapiens/annotation/GRCh38_latest/refseq_identifiers/GRCh38_latest_genomic.fna.gz`.

The results of the contamination removal (if it was done) would be added to the qc-report.txt. Currently, the final number of reads is in the column `total pairs passed`/`total pairs passed (percent of all pairs)`. So now `total pairs passed` will still show the final number of reads (so after QC AND after contamination removal), and `total pairs passed QC` will show the number of reads that passed QC. In addition, there would be two more columns: `number of pairs removed due to reference contamination removal`, and `number of pairs removed due to reference contamination removal (percent of all pairs)`

The `no-remove` flag, if set to `true` in the config file, will make anvio count the number of reads that mapped to the references-for-removal, but it will not remove the short reads from the fastq file. The motivation for using this flag is if you want to use this information for a better relative abundance estimation, but you don't want to risk damaging your assembly, AND you don't mind paying the price of having to assemble reads from contamination.

ShaiberAlon:
closes #1011

For details about this please refer tothe new section of the tutorial: http://merenlab.org/2018/07/09/anvio-snakemake-workflows/#reference-based-short-read-removal


meren:
Do you know who is the best Alon?

Yes. You. You are the best.

ShaiberAlon:
Hehe, thank you @meren.

Issue #1033, opened 2018-11-19
Additional data layers in gene mode


xvazquezc:
Hi there,

I'm trying to incorporate metaproteomics data into anvi'o but I'm having problems displaying it. So, any suggestion is welcomed.

---
Using `-A ` with `anvi-interactive` in `--gene-mode`, anvi'o loads the data but it shows the following message and the layers for the metaprot can only be displayed as text or colour (only two options in the interactive interface), which doesn't make a lot of sense for a float type

```
$ anvi-interactive --browser-path /usr/bin/chromium-browser -p PROFILE.db -c contigs.db --gene-mode -C default -b Bin_1 -A ../../../../metaprot/metaprot2.tsv

[........]

WARNING
===============================================
OK. While working on the layer 'Iso_4', which actually looked like a numerical
layer, anvi'o realized that 1890 of your items (for instance '5933' was one of
them) did not have a value for this layer. To make sure things will continue
working in the interface, anvi'o took the liberty of adding zeros as values for
these items. Which is not the smartest thing to do, but we unfortunately do not
support empty values for numerical layers yet. In MetalBeard's voice: things
shall continue to work, but ye here be warned. Back to anvi'o regular voice:
Please keep this in mind while you are studying the interactive interface be
extra careful how to interpret your analysis when you see zero values in the
layer 'Iso_4'.
```

Having so many items with no values is normal given the sparsity of any proteomic dataset.

This is a sample of the additional layer file sample lines, where `item_name` is the gene/protein id from the anvi'o gene calling:
```
item_name       H2_HCO3_1       H2_HCO3_2       H2_HCO3_3       H2_HCO3_4       Iso_1   Iso_2   Iso_3   Iso_4
2       1.0721  0       0.9196  0.96816 0       0       0       0
23      0       0       0       0       0       0       0.95913 0
24      0       0       1.8392  0       1.4378  0.93081 1.9183  0
32      1.0721  0.97968 0.9196  0.96816 0       0       0       0
46      0       0       0.9196  2.9045  1.4378  0.93081 0.95913 0.89786
47      0       0       0.9196  0       0       0       0       0.89786
48      1.0721  0.97968 0       0       0       0       0       0
49      1.0721  0       0       0.96816 0       0       0.95913 0.89786
53      1.0721  0       2.7588  2.9045  0       0.93081 1.9183  2.6936
54      2.1443  3.9187  1.8392  1.9363  0       0.93081 0.95913 1.7957
[...]

PR #1039, opened 2018-11-21
Genes database + misc data tables

meren:
Since Xabi is trying to incorporate metaproteomics results into `--gene-mode`, I thought the best way to do it would be through misc data tables infrastructure. This PR adds gene database the misc data capabilities among other critical improvements.


meren:
Related to #1033.

Issue #1021, opened 2018-10-25
Making anvi-gen-variability-profile to optionally report VCF files


meren:
What `anvi-gen-variability-profile` reports is much more comprehensive than the VCF file format that is documented here:

http://samtools.github.io/hts-specs/VCFv4.3.pdf

However, this shall stay here as a small project for someone who has extra time and interest in contributing to anvi'o.

evensannesriiser:
Hi @meren!

A good colleague of mine created a Python script for converting the output of `anvi-gen-variability-profile` to a .vcf file which may be used for different variant calling tools. By using the .vcf files as input in smartpca in the EIGENSOFT package, we were able to formally test the patterns seen in the heatmap created using your description "[Visualizing SNV profiles using R](http://merenlab.org/tutorials/infant-gut/#chapter-iv-microbial-population-genetics)", see attached heatmap. See also the attached PCA, where each dot represents a sample (column in heatmap). In this case, no differences in SNV distribution are observed between the two groups, which is also expected from eyeballing the heatmap.

Anvi'o users are free to use the script, please credit Srinidhi Varadharajan, Center for Ecological and Evolutionary Synthesis (CEES), University of Oslo, Norway.

Best,

Even Sannes Riiser,
PhD candidate,
University of Oslo, Norway

![snvs](https://user-images.githubusercontent.com/11213057/48490830-8a360f00-e826-11e8-859f-0d7199145856.png)

![smartpca](https://user-images.githubusercontent.com/11213057/48490483-d7fe4780-e825-11e8-9872-240402dc3c04.png)

meren:
Hi @evensannesriiser,

Thank you very much for the heads-up and for your time. This is definitely excellent news!

We can add something to the variability tutorial and offer a link to Srinidhi's repository, so the users who are interested in generating VCFs could benefit from it. That is the simples and quickest solution. 

But I think a more appropriate way to include this addition to the standard workflow is to add this function into the variability module of anvi'o and associate it with a flag like `--report-vcf-output`. This way anvi'o can call it with its internal data structures, and it would reach to a larger number of users. I'm sure we can find multiple ways to credit Srinidhi.

Do you think Srinidhi would be interested in such a solution?

Best wishes,

meren:
Ping!

evensannesriiser:
Haha, I passed your request on to Srinidhi, and figured she had answered. I'll remind her! ;)

E

srinidhi202:
Hi @meren,
[...]

PR #1046, opened 2018-11-23
Anvi'o variability output to VCF format conversion

srinidhi202:
Added the script to convert  anvi-gen-variability-profile  output to  VCF


meren:
related to #1021.

PR #1053, opened 2018-12-02
Update __init__.py

AstrobioMike:
When running `anvi-gen-contigs-db` and setting `-L` to "0", would fail with this message 

"""
Config Error: Creating a new contigs database requires split length information to be
              provided. But the ContigsDatabase class was called to create one without this
              bit of information. Not cool.
"""

So removed suggesting "0" from the program help


meren:
Hi Mike,

I solved it with b570ae616f26f887d60ea59ff16395bf0eadacbc, as I felt that this is a better approach to this problem :) Thank you for the PR.

PR #1066, opened 2018-12-09
should count number of genomes and not number of profile databases

ShaiberAlon:
@meren, this is no big deal, but I noticed this so why not fix.

right now this is counting the number of unique profile databases within the internal genomes file. But the message (three lines of code below) is talking about how many genomes are coming fro internal genomes. right?


meren:
sure ;)

meren:
thank you!

Issue #926, opened 2018-08-01
in metagenomics workflow - minimum length of contigs in the contigs database and in profiling should be identical by default


ShaiberAlon:
See discussion on slack: https://anvio.slack.com/archives/C8SFMGYF3/p1533142464000471?thread_ts=1531174260.000042&cid=C8SFMGYF3

Both would be set to 2,500.

Since there is no `--min-contig-length` flag to `anvi-gen-contigs-database`, in order to do this we must change the `--min-len` for `anvi-script-reformat-fasta` to 2,500. 

So far we have been deleting the raw output of the assembly and only keeping the reformatted format, but not I see that it's a mistake, since it doesn't allow the user to go back and look at smaller contigs if they decide to. So from now on I suggest to do the following thing:
1. first run `anvi-script-reformat-fasta` with `--simplify-names`, but with `--min-len 0` (and `--prefix`) - this copy will include all the contigs from the raw assembly, but it will have simplified names that would be consistent with all the following analysis.
2. then run `anvi-script-reformat-fasta` **without** `--simplify-names`, but with `--min-len 2,500` (or whatever was set in the config). This file will be deleted at the end of the workflow so as to not keep two copies of the assembly. We can add a flag to the config file so that the user can decide if they want to keep this output too. But also, it is pretty easy to create if needed.

ShaiberAlon:
@meren, let me know what you think about this please ^^

meren:
I suggest we set the default to 1,000 for all steps, and put a warning statement for both contigs and the profile step. We can change the default in `anvi-profile`.

I like the two-step approach for `anvi-script-reformat-fasta` :)

ShaiberAlon:
Opening this issue because we need to set the output of the second reformatted fasta as temporary (as described in the original issue comment above).

ShaiberAlon:
@meren, in your comment above, you suggested to change the default `--min-contig-length` of anvi-profile to 1000. Do you still want to do that?

meren:
Yes :)

ShaiberAlon:
So me, or you? :-)

ShaiberAlon:
Ok PR awaits for you @meren :-)

PR #1095, opened 2019-02-04
changed min contig from 2500 to 1000

ShaiberAlon:
This is re: https://github.com/merenlab/anvio/issues/926
And closes https://github.com/merenlab/anvio/issues/926.


meren:
ONE OF THE BIGGEST COMMITS IN THE HISTROY OF AN'VIAR

jarrodscott:
I am curious. Is min-length of 1000 what you are all using for your metagenomics workflows now?

meren:
It changes from study to study, but we thought keeping 1,000 as default is better as a general default suggestion.

PR #1107, opened 2019-02-18
kraken_hll -> krakenuniq

ShaiberAlon:
@meren, I have made all the changes and tested with the mini test.


meren:
Seems great :) thank you!

PR #1114, opened 2019-02-27
Add if __name__ == '__main__': for consistency across scripts

blankenberg:
These 4 files were the odd ones out.


meren:
Thank you very much! :) We like consistency but lack it sometimes :p This is great.

PR #1115, opened 2019-03-01
Allow a user to shutdown anvi-interactive server via the web interface if it was run with --user-server-shutdown

blankenberg:


meren:
Looks great! Thank you very much :)

PR #1117, opened 2019-03-07
user-server-shutdown by user in anvi-display-pan

xvazquezc:
This adds the `--user-server-shutdown` option to `anvi-display-pan` like #1115 does for `anvi-interactive`


meren:
Thanks! :)

PR #1123, opened 2019-03-14
Fix for LinkMers.list_contigs().

blankenberg:
Since we require all BAMs to have the same contigs, we'll just use the values from the first BAM file.


meren:
Great catch, Daniel. Thank you very much!

PR #1129, opened 2019-03-28
Add parameter --max-coverage-depth to anvi-profile, with default 60.000

ozcan:


meren:
This is a pretty significant addition, @ozcan. Great!

PR #1131, opened 2019-04-02
Add max coverage flag to snakemake metagenomics workflow

efogarty11:
@ShaiberAlon can you check to make sure that this makes sense and merge? It worked when I did a test run for snakemake with one metagenome. Thank you!


ShaiberAlon:
Looks great! Thank you @efogarty11 !

PR #1136, opened 2019-04-12
Combined Homogeneity (Pangenomics)

mahmoudyousef98:
- **Fixed a bug** in the calculation of Functional Homogeneity
- **Implemented Combined Homogeneity** using the inverse reciprocal mean of the functional and geometric homogeneity indices
- Implemented gene cluster **filtering by combined homogeneity** and adjusted it so that pangenomes without combined homogeneity information will not crash


meren:
this looks very good @mahmoudyousef98!

did you have a chance to test it with Prochlorococcus and see whether things made sense?

mahmoudyousef98:
Thank you! I did, I found it to be a relatively decent estimate for the gene clusters I looked at!

Issue #1125, opened 2019-03-18
Output issues with anvi-export-table on a  GENOMES.db 


jarrodscott:
Running `anvi-export-table` on a `GENOMES.db` with the `--table gene_function_calls` flag produces a file where each row is a `genome_name`, followed by `entry_id` ,   `gene_callers_id`, `source`, `accession`, `function`, `e_value`. Most of these do not make sense for a list of genomes. 

Running `anvi-export-table` on a `GENOMES.db` with the `--table gene_function_calls` flag produces this error:

```
Config Error: This is one of the core functions of anvi'o you never want to hear from, but   
there seems to be something wrong with the table 'gene_info' that you are trying to read 
from. While there are 113205 items in this table, there are only 47 unique keys, which 
means some of them are going to be overwritten when this     function creates a final 
dictionary of data to return. This may be a programmer error when the data was being 
inserted into the database, but needs fixin'  before we can continue. If you are a user, 
please get in touch with anvi'o  developers about this error. If you are a programmer, you 
probably did something wrong :( 
```

Running with either the  `self` option or `genome_info` option everything seems fine.

```
Anvi'o version ...............................: margaret (v5.4-master)
Profile DB version ...........................: 31
Contigs DB version ...........................: 12
Pan DB version ...............................: 13
Genome data storage version ..................: 6
Auxiliary data storage version ...............: 2
Structure DB version .........................: 1
```

ozcan:
Thank you for reporting @jarrodscott, We now fixed this on master version.

PR #1139, opened 2019-04-17
Fix anvi-export-table problems, Related #1125

ozcan:
Using pandas in `anvi-export-tables` for reading and exporting tables solves exporting problems, However, the real solution is adding `entry_id` to genome storage database, which I also implemented but decided to skip as that will cause so many untested side effects. @meren Please let me know if you want me to keep the `entry_id` changes and add an upgrade script for genome storage?


meren:
I agree with your resolution here, @ozcan. I think you implemented a the much cleaner solution for this one.

PR #1150, opened 2019-05-01
The great SCGs update

meren:
This PR implements an entirely new completion/redundancy estimation framework for anvi’o with new HMMs and a random forest classifier to predict which domain specific HMMs to use to assess C/R. The changes for the end user should be minimal.

Everything started with the Random Forest work. Using the random forest algorithm, we solved the mess of which SCG collection should we use to estimate C/R for a given genome. Then multiple reasons such as the fact that Rose Kantor and others had previously pointed out to us that some of the Campbell et al HMMs were redundant (i.e., they have different names but yield the same genes), and our own observations of minor issues with Rinke et al and Campbell et al collections made it look like this would be a good opportunity to revamp the entire C/R approach along with HMMs we use.

In this PR we have new SCG HMMs for Bacteria and Archaea. For both I used Mike Lee's recent HMM collections as a starting point, but then I ended up changing the collection for Bacteria quite a bit as I realized some inconsistencies with it. Archaea is as is, but I have to admit I did not test that collection as throughly as I did Bacteria. Tom Delmont's HMMs for Eukarya didn’t change.

I've done many benchmarks that varied with their details. But here is a good one that uses Tom Delmont's unpublished ocean MAGs, and compares old completion estimates to the new one:

image

Long story short, things are pretty comparable, but we see a decrease in redundancy estimates as the new Bacteria_71 collection removes many genes that were not necessarily single-copy in a serious fraction of bacterial genomes.


luuuuuuuke:
Hey there! Is there any way to access the old Campbell/Rinke SCG profiles? I am revising a manuscript in which I originally used anvio v5.3 and we did a particularly detailed analysis of the Campbell SCGs in our bins. I need to do those same analyses with newly generated bins but now I'm working with anvio v6 (ie, Bacteria_71) and when I try to run anvio v5.3 I get a traceback error. Not sure if there is an easy way to tell anvio v6 to use the old Campbell hmm but that would solve my problem. Any help here would be much appreciated :)

meren:
You can download the directories of HMM data from here:

https://github.com/merenlab/anvio/tree/v5.5/anvio/data/hmm

and run them on any contigs db with `anvi-run-hmms -H Campbell_et_al`, and so on.

Let me know if something doesn't work.

luuuuuuuke:
That worked thank you!

PR #1168, opened 2019-06-06
Timer

ekiefl:
Introduction of a Timer class in terminal.py

Integrates into Progress to provide ETA and has potential to be used generically in other classes as well.

The premise of the class is to build an ordered dictionary, where each key and value combination is a checkpoint name, and timestamp. If a required completion scores provided, the class can estimate how much time is left.

### DIFF
--- a/anvio/terminal.py
+++ b/anvio/terminal.py
@@ -376,6 +376,9 @@ class Timer:
         self.completion_score = 0
         self.complete = False
 
+        self.last_eta = None
+        self.last_eta_timestamp = self.timer_start
+
 
     def timestamp(self):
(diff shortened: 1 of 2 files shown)


meren:
This looks pretty awesome to me :)

My only concern is the `format_time` function. In some cases it will be called thousands of times throughout some progress loops, and the overhead of that function will add up. A very appropriate addition would have been to make sure it is not called and the previous formatted text is returned if it is being called X times within the same second.

ekiefl:
Added and tested described feature @meren . I will let you pull the trigger In case you want to change anything

meren:
Let's go and see it in action :) This is good timing because it will get tested for the next couple of weeks before the release.

PR #1173, opened 2019-06-14
fix strip acting on tabs (our separator string)

ekiefl:
Sometimes valid tab delimited files begin their first non-commented line with the tab character. `strip()` removes this tab character and messes up the number of columns. This was brought to my attention when getting different results from `anvi-matrix-to-newick` for the following data files:

image

image

anvi-matrix-to-newick produces the wrong tree file for the first case, and this pull request fixes that issue.


meren:
Excellent catch, @ekiefl! Thank you.

Issue #1175, opened 2019-06-18
anvi-script-visualize-split-coverages cannot handle files with multiple split_names/contigs


mooreryan:
## Version

anvio git repo version/commit: bfbcbb3592dfd18bd7ffc9880c5dac1e88bb37fb

## Problem

I was doing more work on the `anvi-script-visualize-split-coverages` script when I realized that multiple contigs/split_names per split coverages file will break the program.

I made a toy dataset to recreate the issue.

### Test data

Command run: 

```
./anvi-script-visualize-split-coverages -i multi_contig_multi_sample.split_cov.txt -o multi_contig_multi_sample.split_cov.pdf
```

Input text file: 
[multi_contig_multi_sample.split_cov.txt](https://github.com/merenlab/anvio/files/3302920/multi_contig_multi_sample.split_cov.txt)

Output PDF: 
[multi_contig_multi_sample.split_cov.pdf](https://github.com/merenlab/anvio/files/3302921/multi_contig_multi_sample.split_cov.pdf)

## Solutions?

Not sure the best way to fix this, but here are a couple possible fixes....

* Output a new PDF for each contig
* Add panels for each contig-sample pair, all in the same PDF
  * Grouped by sample: [multi_contig_multi_sample.split_cov.sample_name_by_split_name.pdf](https://github.com/merenlab/anvio/files/3302928/multi_contig_multi_sample.split_cov.sample_name_by_split_name.pdf)
  * Grouped by contig/split_name: 
 [multi_contig_multi_sample.split_cov.split_name_by_sample_name.pdf](https://github.com/merenlab/anvio/files/3302931/multi_contig_multi_sample.split_cov.split_name_by_sample_name.pdf)

meren:
Hey @mooreryan,

Thank you for catching this! Indeed, `anvi-get-split-coverages` can report all splits in a bin/collection .... I totally forgot about this :/

My 2 cents suggestion is to have an additional flag (i.e., '`--report-each-split-separately`') --when declared, the program would report an independent PDF file for each split. Otherwise it would merge coverage values for ALL splits in alphabetical order, and display merged coverages all at once (so the number of nucleotides displayed in the X-axis would be equivalent to the total number of nucleotides in all splits per sample). If doing this is too much of a pain, reporting multiple PDF files in the presence of multiple splits is perfectly acceptable, in my opinion :)

Thank you very much regardless.

Best wishes,

meren:
[...]

PR #1177, opened 2019-06-20
Handle multiple splits (#1175)

mooreryan:
Addresses https://github.com/merenlab/anvio/issues/1175.

Turns out, it is easier to merge all the splits like you mentioned rather than displaying them all separately, so I did that instead.

### DIFF
--- a/sandbox/anvi-script-visualize-split-coverages
+++ b/sandbox/anvi-script-visualize-split-coverages
@@ -87,17 +87,17 @@ smooth_coverage <- function(split_coverages, window_size) {
     )
 }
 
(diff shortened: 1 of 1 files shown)


mooreryan:
Second commit (cbd46e9530004132f124db36edb529bc2bd9f574) addresses an issue with the chart sizing.  If the user put only a subset of samples in the sample data file, the chart sizes were scaling to all the samples in the split_coverage file rather than only the samples that were drawn on the chart.

meren:
Great changes, @mooreryan! I will merge first and then test :)


Take care,

PR #1178, opened 2019-06-20
Unit testing visualize-split-coverages script functions

mooreryan:
Added some unit tests for some of the non-trivial functions in the `visualize-split-coverages` R script.

It uses the `testthat` package.

To run tests, first ensure `testthat` is installed.  Then (from the `anvio` source directory) run 

```
Rscript anvio/tests/sandbox/test_visualize_split_coverages/run_tests.R
```

Not 100% sure if `anvio/tests/sandbox/test_visualize_split_coverages` is the best directory for unit tests for the script however....

### DIFF
--- a/anvio/tests/sandbox/test_visualize_split_coverages/run_tests.R
+++ b/anvio/tests/sandbox/test_visualize_split_coverages/run_tests.R
@@ -1,0 +1,29 @@
+library(testthat)
+
+## First get the directory of this test runner file.
+opts <- commandArgs()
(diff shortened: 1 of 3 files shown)


meren:
> Not 100% sure if anvio/tests/sandbox/test_visualize_split_coverages is the best directory for unit tests for the script however....

`anvio/tests/sandbox` is pretty much free for all, so it *is* a good place to put those files :)

Thank you!

Issue #1179, opened 2019-06-20
Adding a --verbose option to anvi-script-visualize-split-coverages 


mooreryan:
This commit 080db464e632eb5f6ef5982b9d58d8eae41e0ea0 updated the look of the logging for the `anvi-script-visualize-split-coverages`. 

The other day I got about 1/2 way through adding a `--verbose` flag and further cleaning up the logging.  

Something like this: https://gist.github.com/mooreryan/4e67b1ecbe097b43481c433908e81634

Log messages would look some like this then 

```
E, [2019-06-20 18:21:28.150640 #18632] ERROR -- Hi error
W, [2019-06-20 18:21:28.150780 #18632] WARN -- Hi warn
I, [2019-06-20 18:21:28.150978 #18632] INFO -- Hi info
```

The `#18632` is the process ID.

This issue is more of a placeholder in case @meren or anyone else was doing more changes to the logging.

mooreryan:
This commit ac9ebb956c89817f0c90c3c6c2dabbfeeabb1b6e fixes this issue.

PR #1180, opened 2019-06-20
Add --verbose flag and update logging

mooreryan:
References issue #1179.

* Added a `--verbose` option.  It takes 3 levels (1 for just errors, 2 for in the middle, 3 for everything)
* Added more general logging function


meren:
Great :) Thank you, Ryan!

PR #1185, opened 2019-07-04
New views for INSEQ/Tn-Seq data

eburgoswisc:
Code for calculating stats for INSEQ/Tn-Seq data when running gene mode in anvio.


meren:
Looks good to me. There are some potential improvements, but they can be done over time.

@ozcan, please take a look and merge it if/when you feel it is ready :)

meren:
Great job with this, @eburgoswisc!

meren:
The only key component that is missing is some files for testing, @eburgoswisc. similar to mini test, if there were some very tiny BAM files for three samples, an some additions in `run_all_tests.sh` to cover them it would be certain that anvi'o releases continue to maintain the code for potential bugs caused by other changes.

eburgoswisc:
I have some example files I can add.

meren:
Sounds great. You can add some files under the directory `anvio/tests/sandbox/example_files_for_inseq_tnseq/` and create a new batch file (i.e., `run_pangenome_tests.sh`) under the directory `anvio/tests/`. You can copy the logic in `anvio/tests/run_mini_test.sh`.

The key is to make sure those BAM files you add under the `example_files_for_inseq_tnseq` directory (i.e., `INSEQ-0{1,2,3}-RAW.bam` and the corresponding `contigs_inseq.fa`) are small in size (since they are going to be shipped with every anvi'o installation and we would like to keep the package as small as we can), but still meaningfully complex so they are certainly covering multiple scenarios :)

anvio/dbops.py
```diff
@@ -131,7 +133,7 @@ def __init__(self, args, r=run, p=progress):
 
         if not self.contigs_db_path:
             raise ConfigError("Someone (hopefully, you) is trying to initialize the Contigs Super Class without a contigs database path.\
-                               There are many ways this can happen, but .. do you think you were trying to run anvi-interactive in\
```
ozcan: Little typo here

anvio/tables/__init__.py
```diff
@@ -116,6 +116,10 @@
 gene_level_coverage_stats_table_structure = ['entry_id', 'gene_callers_id', 'sample_name', 'mean_coverage', 'detection', 'non_outlier_mean_coverage', 'non_outlier_coverage_std', 'gene_coverage_values_per_nt', 'non_outlier_positions']
 gene_level_coverage_stats_table_types     = [ 'numeric',     'numeric'    ,     'text'   ,    'numeric'   ,  'numeric' ,         'numeric'          ,          'numeric'        ,             'blob'           ,          'blob'        ]
 
+gene_level_inseq_stats_table_name      = 'gene_level_coverage_stats'
+gene_level_inseq_stats_table_structure = ['entry_id', 'gene_callers_id', 'sample_name', 'mean_coverage', 'insertions', 'insertions_count_normalized', 'detection', 'mean_disruption', 'below_disruption_threshold']
+gene_level_inseq_stats_table_types     = [ 'numeric',     'numeric'    ,     'text'   ,    'numeric'   ,  'numeric' ,         'numeric'          ,          'numeric'        ,             'numeric'           ,          'numeric'        ]
+
```
ozcan: We need to bump gene db version and write an upgrade script

anvio/tables/genelevelcoverages.py
```diff
@@ -29,11 +29,14 @@
 
 
 class TableForGeneLevelCoverages(Table):
-    def __init__(self, db_path, parameters, split_names=None, ignore_splits_name_check=False, run=run, progress=progress):
+    def __init__(self, db_path, parameters, split_names=None, ignore_splits_name_check=False, table_name=None, table_structure=None, mode=None, run=run, progress=progress):
```
ozcan: I really liked that you parameterized table info and reused this class

bin/anvi-gen-gene-level-stats-databases
```diff
@@ -128,8 +133,13 @@ if __name__ == '__main__':
     groupD = parser.add_argument_group('PARAMETERS OF CONVENIENCE', "They say they save lives.")
     groupD.add_argument(*anvio.A('just-do-it'), **anvio.K('just-do-it'))
 
+    groupE = parser.add_argument_group('INSEQ DATA', "When analyzing INSeq/Tn-Seq data")
+    groupE.add_argument("--inseq-stats", action="store_true")
```
ozcan: These parameters need to be moved to `anvio/__init__.py`, so you can use the same help_text in multiple programs as you do in here and `anvi-interactive`.

anvio/tables/genelevelcoverages.py
```diff
@@ -174,24 +175,24 @@ def read(self):
...
                         d.append(utils.convert_numpy_array_to_binary_blob(np.array(entry[h]), 'uint16'))
                     else:
                         d.append(entry[h])
 
-                db_entries.append(tuple([self.next_id(t.gene_level_coverage_stats_table_name)] + d), )
+                db_entries.append(tuple([self.next_id(self.table_name)] + d), )
 
         database = db.DB(self.db_path, utils.get_required_version_for_db(self.db_path))
-        database._exec_many('''INSERT INTO %s VALUES (?,?,?,?,?,?,?,?,?)''' % t.gene_level_coverage_stats_table_name, db_entries)
+        database._exec_many('''INSERT INTO %s VALUES (?,?,?,?,?,?,?,?,?)''' % self.table_structure, db_entries)
```
ozcan: There is an error here, it should be table_name instead of structure

anvio/tables/__init__.py
meren: Ah, this is a very important point!

meren:
Hey Ema, thanks for these changes. Looking at the tests, I'm realizing that it is best if we have a contigs database rather than computing it every time we test it (the success of FASTA to CONTIGS.db conversion is being tested in other tests anyway). Can you please remove `example_files_for_inseq_tnseq/contigs_inseq.fa` and instead add a `example_files_for_inseq_tnseq/CONTIGS.db`?

This way we can keep upgrading the contigs database and be able to get the contig sequences if it becomes necessary.

eburgoswisc:
Woops! My dropbox synced it back on my iMac... Now its good

meren:
Excellent :) thank you very much.

Issue #1189, opened 2019-07-17
TypeError for tnseq/inseq tests


meren:
This is in `master`.

Running `./run_gen_gene_level_stats_tests.sh` after the last commit (1045128c801f1ce25637ca5226442f81bd9913bb) results in the following error:

```
Traceback (most recent call last):
  File "/Users/meren/github/anvio/bin/anvi-interactive", line 108, in <module>
    d = interactive.Interactive(args)
  File "/Users/meren/github/anvio/anvio/interactive.py", line 201, in __init__
    self.load_gene_mode()
  File "/Users/meren/github/anvio/anvio/interactive.py", line 1047, in load_gene_mode
    self.init_gene_level_coverage_stats_dicts()
  File "/Users/meren/github/anvio/anvio/dbops.py", line 2464, in init_gene_level_coverage_stats_dicts
    self.init_gene_level_coverage_stats_from_genes_db(parameters)
  File "/Users/meren/github/anvio/anvio/dbops.py", line 2402, in init_gene_level_coverage_stats_from_genes_db
    self.gene_level_coverage_stats_dict = table_for_gene_level_coverages.read()
  File "/Users/meren/github/anvio/anvio/tables/genelevelcoverages.py", line 159, in read
    data[gene_callers_id][sample_name]['gene_coverage_values_per_nt'] = utils.convert_binary_blob_to_numpy_array(g, 'uint16')
  File "/Users/meren/github/anvio/anvio/utils.py", line 531, in convert_binary_blob_to_numpy_array
    return np.frombuffer(gzip.decompress(blob), dtype=dtype)
  File "/Users/meren/miniconda3/lib/python3.6/gzip.py", line 531, in decompress
    with GzipFile(fileobj=io.BytesIO(data)) as f:
TypeError: a bytes-like object is required, not 'int'
```

Probably an easy fix for @eburgoswisc :)

eburgoswisc:
I can fix it! Also, im getting this error when running gene mode in anvi-interactive. I started working on a fix if thats all right.

```
Traceback (most recent call last):
File "/Users/emanuelburgos/venvs/anvio-master/bin/anvi-interactive", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/emanuelburgos/bioinformatic-modules/anvio/bin/anvi-interactive", line 108, in <module>
    d = interactive.Interactive(args)
[...]

PR #1192, opened 2019-07-20
Addressing TypeError issue in inseq/tnseq tests

eburgoswisc:
PR for fixing issue #1189. Had to add gene inseq stats as a table to profile database.


anvio/dbops.py
```diff
@@ -2573,7 +2574,7 @@ def get_gene_level_coverage_stats_entry_for_inseq(self, gene_callers_id, split_c
 
         # INSEQ/Tn-SEQ views
         mean_coverage = 0
-        detection = numpy.count_nonzero(gene_coverage_values_per_nt) / gene_length
+        #detection = numpy.count_nonzero(gene_coverage_values_per_nt) / gene_length
```
meren: so if I am not wrong, this is the part where you are calculating the detection value of a gene in INSEQ/Tn-SEQ results, right?

In metagenomic read recruitment studies it makes a lot of sense to talk about 'detection' as a function of the number of positions in the gene that is covered by at least 1X (as described [here](http://merenlab.org/2017/05/08/anvio-views/#detection)). But in the case of INSEQ/Tn-SEQ, I think detection statistic should behave differently so the user can immediately see genes that recruit no reads at all. For this, one could imagine setting the detection value to 1.0 even if 5% of the positions in the gene are covered. I'm not sure if I'm making any sense, but I wanted to add this as a comment.

PR #1194, opened 2019-07-22
anvi-export-locus with flanking genes

mschecht:
This PR will add additional functionality to `anvi-export-locus`.  With the `--flanking-genes` the function should now be able to cut out a locus based on 2 flanking genes one provides.

### DIFF
--- a/anvio/splitter.py
+++ b/anvio/splitter.py
@@ -736,8 +736,8 @@ class LocusSplitter:
         if len(self.num_genes_list) == 1:
             self.num_genes_list = [0, self.num_genes_list[0]]
 
-            # # add sanity check for flanking genes 
-            # if self.flanking_genes 
+            # # add sanity check for flanking genes
+            # if self.flanking_genes
 
 
         self.run.warning(None, header="Input / Output", lc="cyan")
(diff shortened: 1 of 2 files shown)


anvio/splitter.py
```diff
@@ -822,17 +827,12 @@ def process(self, skip_init=False):
 
     def export_locus(self, gene_callers_id, output_path_prefix):
```
meren: This is not working because you are adding hardcoded instructions into a function that is designed to work with single gene caller ids (see the function header `export_locus`) to make it work with multiple of them.

The critical variable here is `self.gene_caller_ids_of_interest` in the class. The contents of which is set in the `init` function, right after this line:

```
self.run.warning(None, header="Initialization bleep bloops", lc="cyan")
```

You basically need to add one more `elif` there to make sure if the user has flanking genes, the `self.gene_caller_ids_of_interest` is set according to that (in which case it can contain tuples for start and end gene caller ids).

THEN, you can turn `export_locus` function into multiple atomic functions: you can do that by renaming the current function `export_locus` to `export_locus_using_single_gene_calls` and then implement a new one `export_locus_using_flanking_gene_calls` (and add docstring to both to make sure it is clear how and why do they differ), a function that would be very similar to `export_locus_using_single_gene_calls`, but would take two gene callers id. You can do it by changing the for loop that starts like this:

```
for gene_callers_id in self.gene_caller_ids_of_interest:
    (...)
```

into something that looks like this instead:

```
for item in self.gene_caller_ids_of_interest:
    if self.flanking_genes:
        first_gene_callers_id, second_gene_callers_id = item.sorted()
        self.export_locus_using_flanking_gene_calls(first_gene_callers_id, second_gene_callers_id, output_path_prefix)
    else:
        gene_callers_id = item
        self.export_locus_using_single_gene_calls(gene_callers_id, output_path_prefix)
```

Then you would have to make sure there is no redundancy in either of these two functions by creating additional helper functions.

It is essential to make 100% sure that the previous code works as intended while you are making these changes. Which means, you must first get `anvi-export-locus` to work with an example, and then keep that in your test file so you always check whether it is still working.

This is just an initial suggestion.

meren:
@mschecht, is this PR still alive?

mschecht:
No this was my first attempt at anvi-export-locus, am I able to kill the PR on my end?

meren:
Sure. We can close it here :)

(I am doing it as a part of this comment)

PR #1210, opened 2019-08-18
Quality of life updates - anvi-dereplicate-genomes

mahmoudyousef98:


anvio/genomedistance.py
```diff
@@ -65,8 +64,11 @@ def are_redundant(self, genome1, genome2):
         return False
 
     def dereplicate(self):
-        for genome1 in self.genome_names:
-            for genome2 in self.genome_names:
+        names = list(self.genome_names)
+        for i in range(len(names))
```
meren: by looking at this very quickly, these for statements seem to be missing `:` characters at the end of them ;)

plus, the next line also has an additional syntax error I think (`i+`).

meren: more generally, I think there must be a much more pythonic way to do this part. Whenever one sees variable names `i` and `j` they should know that they are doing something not so optimal :) For instance, something like this would have given all permutations of genome names in the `self.genome_names` list:

``` python
(...)
from itertools import permutations 

(...)

genome_pairs = permutations(self.genome_names, 2) 
```

and would remove the need for those nested `for` loops.

mahmoudyousef98: Thank you! I'll take care of this

bin/anvi-dereplicate-genomes
```diff
@@ -48,56 +49,64 @@ def main(args):
...
-	if args.keep_ani_results and args.ani_output_dir:
+	if args.ani_output_dir:
 		filesnpaths.check_output_directory(args.ani_output_dir, ok_if_exists=False)
 		os.mkdir(args.ani_output_dir)
 
 	program = genomedistance.ANI(args)
 
 	temp_dir = program.get_fasta_sequences_dir()
 	results = program.process(temp_dir)
+	try:
```
meren: hey @mahmoudyousef98, it seems the indentation starts using 8 whitespaces rather than 4 in some parts of this file. I am wondering if you have changed your editor or something, but I think it would be much better to standardize this.

mahmoudyousef98: Hey @meren, I'm not seeing the indentation inconsistency on my end, either on github or my remote codebase. Do you mind pointing where you see them?

PR #1224, opened 2019-09-01
Redundancy

mahmoudyousef98:
-`anvi-compute-mash-distance`: use sourmash to compute mash distance matrix for a list of internal and external genomes. Similar to `anvi-compute-ani`, results can be imported into a pan database
-`anvi-dereplicate-genomes`: use either sourmash or pyANI (or the results directory provided by a previous run of either of these programs) to determine redundancy across a list of internal genomes and external genomes as well as fasta files and remove those redundant genomes. Upon user request, a fasta text file can also be produced with the dereplicated genomes.
-Various other bug fixes and typos corrected


ekiefl:
A lot has happened. See the commits for complete changes. Here are the high level changes of this branch compared to master:

- `anvi-compute-ani` is now `anvi-compute-genome-similarity` and can be computed with `--program sourmash` or `--program pyANI`.
- `anvi-compute-genome-similarity` accepts any mix of `fasta.txt`, `internal_genomes.txt`, `external_genomes.txt` will even attempt if prompted to attempt importing results into a pan genome. Appropriate checks of name overlap and compatibility with pan genome are performed. 
- We now have a sourmash driver
- `anvi-dereplicate-genomes` is a new program that computes or imports results of genome similarity, and uses similarity matrices to cluster genomes based on user thresholds. It generates a folder output with lots of goodies. If a sequence source is provided, FASTA directory is reported with representative genomes. Clustering is currently what I consider 'simple_greedy', meaning if the threshold is 90% similarity, and A is 95% with B, which is 95% with C, which is 95% with D, then A,B,C,D will all form one cluster even if D is only 80% similar to A. Representatives are picked based on `centrality`, `Qscore`, or `length`.
- There is a new `genomesimilarity.py` file. This is the backend for the programs `anvi-dereplicate-genomes` and `anvi-compute-genome-similarity` and it is modular and extensible for future development.

PR #1232, opened 2019-09-13
add parsing for emapper v2 to emapper.py

dutter:
I probably added way too many fields, but these are the ones I use myself. Maybe drop some?


meren:
Thank you very much for this, Dan :) It is a way to drop functional annotations, so users can always drop annotations they don't want. So I say we go as is.

Best wishes,

PR #1233, opened 2019-09-17
Undo & Redo implementation for the interactive interface

ozcan:
`Cmd/Ctrl + Z` for undo.
`Cmd/Ctrl + Shift + Z` for redo.

Currently keeps track of:
 - selections
 - bin add/delete
 - bin color changes

### DIFF
--- a/anvio/data/interactive/index.html
+++ b/anvio/data/interactive/index.html
@@ -1150,6 +1150,11 @@
         if (ev.keyCode === 84) { // T = 84
             $('#title-panel').toggle();
         }
+
+        if (ev.keyCode === 90 && (ev.metaKey || ev.ctrlKey)) {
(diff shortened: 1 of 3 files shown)


meren:
This is amazing :) I am testing it now, an the first thing I realized is that removal of bins (as a whole or individually) by clicking on the trash bin icon no longer works:

image

:)

meren:
Another thing I realized:

Make a selection:

image

Make bigger selection:

image

CMD+Z:

image

It is all gone, instead of going to the smaller selection.

ozcan:
Good catch @meren, I am looking into this.

ozcan:
Both fixed now @meren, thank you for testing :)

meren:
It works like magic now 😍

PR #1234, opened 2019-09-18
Parallelize hmmsearch runs

ozcan:
I will post benchmark tables here when I am done with that, but I think this branch stable enough for master.

Do you see any problems or suspicious parts @meren?


meren:
going through it by eye, everything looks great, @ozcan. thank you very much, ellerine saglik :)

PR #1238, opened 2019-09-27
Export locus update

mschecht:


anvio/__init__.py
```diff
@@ -844,6 +844,31 @@ def get_args(parser):
                       filtering criteria. In other programs, you may get everything, nothing, or an error. It really depends\
                       on the situation. Fortunately, mistakes are cheap, so it's worth a try."}
                 ),
+    'flank-mode': (
+            ['-f','--flank-mode'],
```
meren: it is a good idea to make sure the abbreviated forms of these flags (i.e., `-f`) do not appear in this file. 

because when it is the case, mixing them will cause errors, and there will be lots of headache. if you can't find an abbreaviated form that makes sense, you can elect to not use any and keep it as `['--flank-mode']`.

anvio/__init__.py
```diff
@@ -844,6 +844,31 @@ def get_args(parser):
...
+              --search-term, --gene-caller-ids, or --hmm-sources. The --flank-mode option is  \
+              appropriate for extracting loci of variable gene number lengths, but are consistantly \
+              located between the same flanking genes in the genome(s) of interest."}
+              ),
+    'num-genes': (
+            ['-n','--num-genes'],
+            {'metavar': 'NUM_GENES',
+             'type': str,
+             'help':
+             "Required for DEFAULT mode. For each match (to the function, or HMM that was searched) a sequence which includes \
```
meren: please consider not adding a new line and keep value for dicts start with the same line as their corresponding keys :) you can take a look at other instances to see how the rest of the file is organized.

anvio/splitter.py
```diff
@@ -803,64 +842,133 @@ def process(self, skip_init=False):
...
         self.contigs_db.init_functions()
-        counter = 1
-        for gene_callers_id in self.gene_caller_ids_of_interest:
+
+        """
+        Here we will differentiate between being in default-mode OR flank-mode. If in default-mode, we will iterate through
+        the gene-caller-ids-of-interest and cut out the locus X amount of genes above and below the gene-caller-id
+        based on the --num-genes given by the user. If in flank-mode, we will only export 1 locus based on the flanking
+        gene-caller-ids provided.
+        """
```
meren: it is a great idea to add comments to code and thanks for doing it, but for those using comment characters is much better. because technically, when the Python interpreter sees this, it interprets it as text. which is not that critical, but it is just not the best practice :) This is ok under function headers where they serve and interpreted as help for the function.

anvio/splitter_matt.py
```diff
@@ -0,0 +1,1378 @@
+# -*- coding: utf-8
```
meren: I feel like this file shouldn't be here. I thought @ekiefl had removed it. is this keep coming back?

anvio/__init__.py
mschecht: removed abbreviated flag

anvio/__init__.py
mschecht: Thanks Meren, I went back and fixed this style inconsistency

anvio/splitter_matt.py
mschecht: Removed the file, I must of not used `git add` when I originally tried to remove the file

anvio/splitter.py
mschecht: Fixed

mschecht:
@meren may I merge with master? @ekiefl gives his stamp of approval.

meren:
Go for it! :)

PR #1239, opened 2019-09-30
Add contigs counts above 5, 10, 20, 50 and 100Kb

FlorianTrigodet:


meren:
Looks good, Florian! Thank you very much :)

Issue #1248, opened 2019-10-04
Error when running functional enrichment


ShaiberAlon:
When I run `anvi-get-enriched-functions-per-pan-group`, I get the following error:

```
Config Error: It looks like something went wrong during the functional enrichment analysis. We
              don't know what happened, but this log file could contain some clues:
              /tmp/tmpf88fzvye
```

And the aforementioned log file includes this error:

```
Error: Column `function_accession` can't be modified because it's a grouping variable
Execution halted
```

## Potential solution - R version
I am using:

```
R version 3.5.1 (2018-07-02) -- "Feather Spray"
```

So this could be an issue with my local version, but if so, then maybe we should add something to check R version, or at least include a message that we require a certain minimal version of R.

I will try installing the latest `R` version and test this again.

## Reproducing this

To reproduce this, you can download the following data package:
https://drive.google.com/file/d/1crwvvDpK_AqC2ngivcZfETOyj7brDhPL/view?usp=sharing

Uncompress the data folder and `cd` into it:

```bash
tar -xzvf TEST-PACKAGE.tar.gz
cd TEST-PACKAGE
```

And then run the enrichment test:

```bash
anvi-get-enriched-functions-per-pan-group -p PAN.db \
                                          -g GENOMES.db \
                                          -o Functional_enrichment_2_groups.txt \
                                          --category-variable light \
                                          --annotation-source COG_FUNCTION
```

ShaiberAlon:
I ran this again with the latest `R` version:
```
R version 3.6.1 (2019-07-05) -- "Action of the Toes"
```

And, sadly, I still get the same error.

@adw96 , do you have an idea regarding where this error is coming from? Maybe some `R` expert in your lab could help me get to the bottom of this?

meren:
@ShaiberAlon, when I run the enrichment test using your `TEST-PACKAGE.tar.gz` this is how things went:

First try:

```
anvi-get-enriched-functions-per-pan-group -p PAN.db \
>                                           -g GENOMES.db \
>                                           -o Functional_enrichment_2_groups.txt \
>                                           --category-variable light \
>                                           --annotation-source COG_FUNCTION
Genomes storage .............................................: Initialized (storage hash: hash0cde9439)
Num genomes in storage ......................................: 31
Num genomes will be used ....................................: 31
Pan DB ......................................................: Initialized: PAN.db (v. 13)
[...]

PR #1249, opened 2019-10-05
Use nest_legacy if tidyr ver >= 1, nest otherwise

mooreryan:
`tidyr` uses new syntax for `nest` as of version `1.0.0` (see [here](https://tidyr.tidyverse.org/reference/nest.html#new-syntax)).  

This bug fix will check to see which version of `tidyr` that the user has installed/loaded.  If it is `>= 1`, then use the `nest_legacy` function, otherwise, just continue on with the `nest` function as written.

Rather than rewrite any of the code that is there, this should work for users of both the old and the new libraries.

(This should address https://github.com/merenlab/anvio/issues/1248)

### DIFF
--- a/bin/anvi-run-enrichment-analysis.R
+++ b/bin/anvi-run-enrichment-analysis.R
@@ -21,6 +21,39 @@ suppressPackageStartupMessages(library(optparse))
 suppressPackageStartupMessages(library(tidyverse))
 suppressPackageStartupMessages(library(magrittr))
 
+## If we got here, that means tidyverse was loaded, which means it
+## must be installed.  This means tidyr must also be installed.  This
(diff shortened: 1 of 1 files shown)


mooreryan:
I tested that this works using the input file from this comment: https://github.com/merenlab/anvio/issues/1248#issuecomment-538597210, with both `tidyr` versions `0.8.3` and `1.0.0`.

mooreryan:
I requested a review from @meren as this approach is a bit "hackish"....

meren:
I think this is a great solution, @mooreryan :) thank you very much again.

ShaiberAlon:
Thank you very much for getting to the bottom of this and solving it @mooreryan!

PR #1260, opened 2019-10-13
Visualize split coverages bug

mooreryan:
* Fix a bug where the program tried to check samples in files that weren't actually provided
* Fix a bug that happened when `sample_group` wasn't provided in the sample data file
* Strip the extra `.pdf` extension when the user includes it in the basename
* Add a lot more tests
* Add to the help message


meren:
😍 

Thank you, @mooreryan!

PR #1262, opened 2019-10-14
fixed help message for fasta-text-file header

dutter:
While using anvi-compute-genome-similarity I noticed the help message says the FASTA_TEXT_FILE second column name should be "fasta_path" when program expects just "path". So I just changed the help message to prompt "path" for future reference. 

Unless anvi-compute-genome-similarity is the odd one out and it should be accepting "fasta_path" in which case I did it backwards?


meren:
Ah, thanks for this, Dan :)

PR #1264, opened 2019-10-15
Add unit tests for homogeneity index module

mooreryan:
Added unit tests for the `HomogeneityCalculator` class, specifically for the `compute_functional_index` function.

(Tests can be run from the project root directory with `python -m unittest` or `python -m unittest anvio.tests.unit.test_homogeneityindex`.)

I wanted to make sure that I knew how the `compute_functional_index` function was working so I figured I would go ahead and write unit tests for it just to be sure (and help y'all out!).

Let me know if this isn't to your liking!

### DIFF
--- a/anvio/tests/unit/test_homogeneityindex.py
+++ b/anvio/tests/unit/test_homogeneityindex.py
@@ -1,0 +1,131 @@
+# -*- coding: utf-8
+# pylint: disable=line-too-long
+"""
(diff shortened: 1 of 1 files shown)


meren:
Oh, this is absolutely awesome, @mooreryan. We really like component testing, but unit testing is essential for certain parts of the code :)

PR #1269, opened 2019-10-15
Add unit test for ambiguous residue comparison order

mooreryan:
I think that I found another bug in the functional homogeneity index.  I added a unit test with the expected behavior, which fails.

I'm thinking that the order in which ambiguous residues are compared should not affect the functional index, but as currently constructed it does.


meren:
Hmm. This is important. @mahmoudyousef98 should take a look at this at some point and let us know what he thinks :)

mahmoudyousef98:
Thanks for bringing this to my attention. I'll have this fixed ASAP

PR #1275, opened 2019-10-17
Add more unit tests for HomogeneityCalculator

mooreryan:
* Added unit tests for `convert_sequences_to_binary_array`
* Added unit tests for `compute_geometric_index`, both the quick and normal versions

The unit tests for `compute_geometrix_index` are pretty ugly, but I thought that ensuring that the implementation matched doing the algorithm by hand would be useful, especially if you want to refactor the original code at some point.  Hence the big block of hand calculations and comments.


meren:
these will be very helpful for @mahmoudyousef98 to double check his changes :)

PR #1276, opened 2019-10-17
Fix geometric index unit tests

mooreryan:
I realized that the way I set up the test sequences made things confusing:

* The alignment had the same number of sequences and residues, which led to a confusing line in the unit test (even though it worked fine since `num_sequences == num_residues` in the original).
* The geometric score was 0.5, which is not good for checking that you're calculating similarity or differences (for example)

### DIFF
--- a/anvio/tests/unit/test_homogeneityindex.py
+++ b/anvio/tests/unit/test_homogeneityindex.py
@@ -161,10 +161,10 @@ class ConvertSequencesToBinaryArrayTestCase(unittest.TestCase):
 
 class ComputeGeometrixIndexTestCase(unittest.TestCase):
     def setUp(self):
-        s1 = 'A-N-'  # 0101
-        s2 = 'AR-D'  # 0010
-        s3 = 'AR-D'  # 0010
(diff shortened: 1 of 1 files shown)


anvio/tests/unit/test_homogeneityindex.py
```diff
@@ -214,13 +220,14 @@ def setUp(self):
         self.quick_geometric_similarity = sum([aln_col1_similarity_score,
                                                aln_col2_similarity_score,
                                                aln_col3_similarity_score,
-                                               aln_col4_similarity_score]) / num_genes
```
mooreryan: dividing by `num_genes` should really be dividing by the number of residues (aka alignment columns).  The commit fixes this.

mooreryan: (The way the test was set up originally, number of genes was equal to number of residues, so it didn't actually cause a problem, it was just confusing.

meren: always something tricky hides the most important bugs :)

PR #1289, opened 2019-10-31
Adaptive kmer

ekiefl:
This used to be a branch for adaptive kmer picking using sourmash. Now that ive realized sourmash is _bad_ for genome comparisons, I cleaned that mess up as best I could and integrated fastANI  into anvio. Many of the python object handling would be insufficient for people wanting to compute 90,000 pairwise comparisons, but this will work will for standard applications. I am happy that genomesimilarity.py looks almost the same as it did before :)


anvio/drivers/sourmash.py
```diff
@@ -54,46 +62,272 @@ def check_program(self):
...
         self.progress.end()
 
         # restore old working directory
         os.chdir(old_wd)
 
-        return matrix
+        return self.results
+
+
+class IterateKmerSourmash(Sourmash):
```
meren: should we keep this class regardless?

meren:
this is great, @ekiefl. and it looks pretty good to me. next week I will work on the tutorial @mahmoudyousef98 wrote and update it with the new stuff.

anvio/drivers/sourmash.py
ekiefl: No lol

ekiefl: I'm going to remove it tomorrow and then merge. Does that sound good?

ekiefl:
Thanks @meren. I think it's exactly the solution we were looking for. Good for @cjain7 for making such a useful algorithm

anvio/drivers/sourmash.py
meren: It does :)

meren:
Yes! Thank you very much for your contribution, @cjain7!

PR #1304, opened 2019-11-25
Improvements to anvi-script-process-genbank

AstrobioMike:
- was failing if entry did not have a "product" key in gene.qualifiers (which apparently some do), so now there's an if statement for that
- i've come across more ways proteins are listed as hypotheticals in genbank files than just "hypothetical protein", so added those others. Now writes out fewer proteins that have no annotation to the functions table
- does not add "gene name" to hypothetical protein annotations, so can capture them more easily to prevent writing out to functions table
- last change was little typo in "functins"

### DIFF
--- a/anvio/contigops.py
+++ b/anvio/contigops.py
@@ -454,13 +454,20 @@ class GenbankToAnvio:
                 else:
                     accession = "None"
 
-                # storing gene product annotation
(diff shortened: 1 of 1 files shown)


meren:
Yay :) Thank you very much, @AstrobioMike!

PR #1305, opened 2019-11-26
add run-scg-taxonomy to contigs workflow

meren:


meren:
Hey @ShaiberAlon, for some reason sometimes `anvi_run_scg_taxonomy` rule runs before `anvi_run_hmms` rule :/ it is not always the case, but only sometimes (this is what I see when I run `run_contigs_workflow_tests.sh`).

Sorry I couldn't find anything relevant in the code to set a specific order. Apologies if I'm missing something obvious. 

Any help is most welcome :)

ShaiberAlon:
Hi Meren,

If you add the output of anvi_run_hmms rule (it’s a flag file) as an additional input to the anvi_run_scgs_taxonomy, then it will fix this.

On another topic, I saw your comment inside anvi_get_sequences_for_gene_calls. That addition seems harmless to me, but I’ll take a look to see why that rule looks that way, and maybe I’ll clean it up once I get a chance.

meren:
Thank you very much, Alon. efd029d8769e561adcd64eb02f4427c184287577 adds the flag file for the rule HMMs as an input for the rule SCG taxonomy, which changes the plan from this:

image

to this:

image

meren:
This now looks OK to me.

ShaiberAlon:
>  This now looks OK to me.
To me as well :-)

Especially this:
image

I admire honesty.

meren:
> I admire honesty.

I admire you.

PR #1317, opened 2019-12-14
Workflow config v1

### LINKED ISSUE #1313: Change the way we deal with deleted parameters in config files
Discussion here: https://github.com/merenlab/anvio/commit/b6dff45e66f3d8e3da3e4a72344a1332f87cef26#r36167134

## Details of the change:
1. We will make it so that deleted parameters are set to `None`.
2. To track the version of config files we will add a version parameter.
3. From here on, a config file with no `version` parameter will not be accepted (an error will be issued, saying either to manually re-enter the version parameter to the config (in case it was deleted by the user), or to use the migration script to migrate the config file.
3. Add a migration script for config file.

## Migration of config file
If no `version` parameter is found in the config file then it will be considered version zero.

The migration from `v0` to `v1` is easy, we will simply take the config file from the user, and will use the `dict` built-in function `update` to "update" the default config file, and then return that result as the new config file.

Here is a sketch of it in code:

```python
import argparse
import anvio.workflows as w
import anvio.filesnpaths as filesnpaths

args = argparse.Namespace(workflow=workflow_name)

workflow_module_dict = w.get_workflow_module_dict()
workflow_object = workflow_module_dict[workflow_name](args)

user_config_v1 = workflow_object.default_config
user_config_v1.update(user_config_v0)

filesnpaths.is_output_file_writable(file_path)
open(file_path, 'w').write(json.dumps(user_config_v1, indent=4))
```

### How to migrate?
I think that the migration should be done using `anvi-run-workflow` with a new `--migrate-config`, but we could also do it through `anvi-migrate-db` (and then we should maybe rename it to `anvi-migrate`).

### How would we know which workflow?
Notice that migration requires us to know the name of the workflow. There are two solutions for this:
1. Add another new parameter to the config file to specify the name of the workflow.
2. Have the user specify the name of the workflow when they provide a config file.

Adding a workflow_name parameter to config would allow any config with version 1 and above to be upgraded in mass (think `anvi-migrate *json`). But we wouldn't be able to do this for configs of `v0`, which means that we can't avoid having an option for the user to specify workflow name.

If we have the user specify the workflow name then they can still do `anvi-run-workflow -w contigs --migrate-config *json`, but they would need to make sure all those config files are truly for contigs workflows.

The need to specify the workflow name, makes me think the migration should be done using `anvi-run-workflow` and not using `anvi-migrate-db`. I think it would be more natural.

### Bottom line - what I plan for now

I will implement a script for migration in `sandbox` and later we can decide what we do:
`anvi-script-migrate-config-v0-v1`

ShaiberAlon:
This closes #1313.

## The unintuitive and inconsistent treatment of parameters that were removed from a config file

This pull request modifies the way we deal with things that have been removed from the config file. Until now, most, but not all parameters that were removed from the config file would automatically get the default value from the default config file. This is a very unintuitive behavior. And to make things worse, there were a few parameters for which this didn't apply (i.e. when they were removed from the config they would not be assigned to their default value), which is down-right inconsistent.

## How we treat removed parameters as of this pull request

This pull request changes things so that when a parameter is removed from the config file, then it assumes the value of an empty string: `""`.

## Migration of old config files

Since this means that old config files would no longer give the same results, we have added a way to migrate config files. We use `anvi-migrate-db` for this (since @ozcan implement it so beautifully, it was really easy to extend). But since now `anvi-migrate-db` is upgrading more than databases, we renamed it to `anvi-migrate` (@meren, please take a look, and let me know if you are Ok with this change).

## Two mandatory parameters added to config files

In order to facilitate config file migration we added two obligatory parameters to every config file:
`workflow_name`
`config_version`

We need the `config_version` in order to track the version so we know it is compatible with the version of anvi'o that is being used.

We need to `workflow_name` because when using `anvi-migrate` we need to know which workflow generated this config file, since the upgrading process depends on that and for example, it would be different for a `contigs` workflow config file vs. a `metagenomics` workflow config file.

## How to migrate your config from version 0 to version 1 with `anvi-migrate`

In order to use `anvi-migrate` a user must manually add the `workflow_name` to their config file. For example:

```
    "workflow_name": "contigs"
```

In order to use `anvi-migrate` you don't have to add the `config_version` since `anvi-migrate` would assume that a config with no `config_version` value is of version 0.

If you have multiple config files in one directory (and you already added the `workflow_name` for all of them), you can migrate all of them like this:

```
anvi-migrate *json
```

## Can I use an old config with a newer version of anvi'o?

No. Just like with databases, anvi'o will check the version of the config and if it is not the same then the following error will be prompt letting you know that you should use `anvi-migrate`:
image

## What will my config file look like after migrating?

The migration step will add any default value to your config file if it was previously removed. This could include things that in practice will not affect your workflow, but to play it safe, they will be added since that would recapitulate the exact behavior that was happening behind the scenes before this pull request. You can then go ahead and re-edit you upgraded config to remove things that you don't want in your config.

@meren, please let me know if all of this sounds good to you.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2070,7 +2070,7 @@ D = {
     'config-file': (
             ['-c', '--config-file'],
             {'required': False,
-             'help': "TBD"}
+             'help': "A JSON-formatted configuration file."}
                 ),
     'additional-params': (
             ['-A', '--additional-params'],
--- a/anvio/migrations/config/v0_to_v1.py
+++ b/anvio/migrations/config/v0_to_v1.py
@@ -42,6 +42,16 @@ def migrate(config_path):
     new_config.update(workflow_object.config)
     new_config['config_version'] = '1'
 
+    ## Deal with special cases
+    special_params = ['fasta_txt', 'references_for_removal', 'references_mode']
+    for param in special_params:
+        if not workflow_object.get_param_value_from_config(special_params):
+            # this are parameters for which the default values
+            # used to be repressed using "repress_default"
+            # if the user deleted these from their config
+            # then we must delete them here as well.
+            del new_config[param]
+
     filesnpaths.is_output_file_writable(config_path)
     open(config_path, 'w').write(json.dumps(new_config, indent=4))
 
--- a/anvio/migrations/config/v0_to_v1.py
+++ b/anvio/migrations/config/v0_to_v1.py
@@ -40,7 +40,7 @@ def migrate(config_path):
     config = workflow_object.config
 
     default_config = workflow_object.default_config
-    new_config = config
+    new_config = config.copy()
     new_config['config_version'] = '1'
 
     ## Deal with special cases
@@ -51,13 +51,9 @@ def migrate(config_path):
         # if the param belongs to special params then we skip it
             continue
         elif type(default_config[param]) == dict:
-            if 'run' in default_config[param].keys() and not w.A(['param', 'run'], config):
-            # if this is a rule that has a 'run' parameter then
-            # if run is not set to true in the config then skip this rule
-                continue
-            else:
             # otherwise update config rule parameters
-                new_config[param] = default_config[param].update(config.get(param,''))
+                new_config[param] = default_config[param]
+                new_config[param].update(config.get(param,''))
         else:
             # if it's not a dict then it's a general parameter
             # update the general parameter
--- a/anvio/migrations/config/v0_to_v1.py
+++ b/anvio/migrations/config/v0_to_v1.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+
+import sys
+import json
+import argparse
+
+import anvio.workflows as w
+import anvio.terminal as terminal
+import anvio.filesnpaths as filesnpaths
+
+from anvio.errors import ConfigError
+
+current_version, next_version = [x[1:] for x in __name__.split('_to_')]
+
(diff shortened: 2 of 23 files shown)


meren:
All these changes look great to me, alon :) Thank you for adding the versioning to these config files.

ShaiberAlon:
Great. Thank you for taking a look @meren. Should I warn people about this change, or let them just find out on their own next time they use `anvi-run-workflow`?

I figured I could write the following message on the anvi'o Slack:

>Dear `@everyone`,
>
>If you are using `anvi-run-workflow` in the `master` version of anvi'o, then you will experience a certain change as of today: if you try to use a config file that you previously created an error will be raised instructing you to use `anvi-migrate` to migrate your config file to a new version. To learn more about this change you can see the details in this pull request: https://github.com/merenlab/anvio/pull/1317#issue-353141058
>
>If you are not using `anvi-run-workflow`, then you can either ignore this message, or better yet, start using it! Here is where you can get started: http://merenlab.org/2018/07/09/anvio-snakemake-workflows/ (notice that the change mentioned here is not reflected in the tutorial yet, since we like to keep the tutorial consistent with the latest stable version of anvi'o.

meren:
I think it is a good idea to send something to Slack :) Perhaps Jarrod or others can help us test it, too.

ShaiberAlon:
@meren, it appears I can't tag `@everyone` in the anvi'o Slack channel. I will send my message without the tag, and you can decide if you want to tag everyone.

PR #1332, opened 2020-01-18
if --large-index, bowtie2 build extensions -> bt2l

ekiefl:
One day (like today) people will start using `--large-index` on bowtie2, which changes bowtie build extensions from bt2 to bt2l. This handles that.


ShaiberAlon:
This looks good to me @ekiefl. If you want, I can run the metagenomics workflow tests before merging.

ekiefl:
That's okay I ran them without error. It's possible `--large-index` has downstream effects, if so I will find out, but since you think it looks good I'll merge now.

ShaiberAlon:
Great. Thanks EvAn!

PR #1339, opened 2020-01-23
Profile improvements

ekiefl:
Alright so this branch was initially created to try and find a way to speed up anvi-profile by only accessing every nth nucleotide position pileup object and then interpolating the coverage values. The idea being that if you only sampled every hundredth nucleotide position, you could interpolate with at least somewhat decent accuracy faster than it would take to loop through each pileup object.

This approach eventually failed because there is no way to create an iterator object with pysam that samples only every nth object. 

To create an analogy, we wanted this:

```python
for pileup in range(0,1000,10):
    something()
```

but all we could get was this:

```python
for pileup in range(0,1000):
    if pileup % 10 != 0:
        continue
    something()
```

By this point, the coverage is already loaded into memory, so we may as well store it. To extend the analogy further, I also tried this approach, but its extremely slow.

```python
counter = 0
for i in range(0,100,10):
    for pileup in range(i, i+1):
        something()
```

----------------------------

So what's in this branch? 4 things

1. Profile is now a little bit faster after switching to numpy arrays. Primarily speedups resulted from numpy's ability to preallocate memory via arrays and then populate them with splicing operations.

2. Coverage class is now generalized and can be run for anvio.contigops.Split, anvio.contigops.Contig, and just plain old strings of contig names.

3. There's a new program called `anvi-script-get-coverage-from-bam` that uses the new and improved Coverage class to get per-nucleotide, per-contig, or per-bin coverage values. It currently takes 3 inputs: either a contig name, a list of contig names, or a collection txt file. Obviously one can imagine incorporating contigs db as an input and do things like gene coverages, and when that day comes maybe this program will graduate from sandbox to bin

4. The results were yielding different results from anvi-summarize. Investigating a little further, I noticed that bin coverage values were different between anvi-summarize and anvi-script-get-coverage-from-bam. The reason is because anvi-summarize calculates bin averages by taking the mean split averages, but it should really take weighted average of its split coverages, where the weight is the length of the split. After considering for a while, I am convinced that all tables for a collection should be weighted in this manner, not just coverage values. If 60% of a bin's length is contributed by just one split, then all of these summary statistics of this split should be upweighted when defining the bin-wide average.

### DIFF
--- a/anvio/dbops.py
+++ b/anvio/dbops.py
@@ -2950,16 +2950,28 @@ class ProfileSuperclass(object):
             for bin_id in collection:
                 # populate averages per bin
                 averages = copy.deepcopy(samples_template)
+
+                # These weights are used to properly account for differences in split lengths.
+                # Consider table_name == 'mean_coverage', for a bin with 2 splits. Without
+                # weighting, if one split is length 100 with coverage 100 and the other is length
+                # 900 wth coverage 500, the mean_coverage for this bin is (100 + 500)/2 = 300. But
+                # more accurately, mean_coverage of this bin is 100*[100/1000] + 500*[900/1000] =
+                # 460
+                weights = []
+
                 for split_name in collection[bin_id]:
                     if split_name not in table_data:
                         continue
+
+                    weights.append(self.splits_basic_info[split_name]['length'])
 
                     for sample_name in samples_template:
                         averages[sample_name].append(table_data[split_name][sample_name])
 
                 # finalize averages per bin:
                 for sample_name in samples_template:
-                    averages[sample_name] = numpy.mean([a or 0 for a in averages[sample_name]])
+                    # weights is automatically normalized in numpy.average such that sum(weights) == 1
+                    averages[sample_name] = numpy.average([a or 0 for a in averages[sample_name]], weights=weights)
 
                 self.collection_profile[bin_id][table_name] = averages
 
--- a/anvio/sequence.py
+++ b/anvio/sequence.py
@@ -11,6 +11,8 @@ from itertools import permutations
 import anvio
 import anvio.constants as constants
 
+from anvio.errors import ConfigError
+
 __author__ = "Developers of anvi'o (see AUTHORS.txt)"
 __copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
 __credits__ = []
@@ -150,18 +152,61 @@ class Coverage:
         self.mean_Q2Q3 = 0.0
 
 
-    def run(self, bam, split, ignore_orphans=False, max_coverage_depth=constants.max_depth_for_coverage):
-        # a coverage array the size of the split is allocated in memory
-        self.c = numpy.zeros(split.end - split.start).astype(int)
+    def run(self, bam, contig_or_split, start=None, end=None, ignore_orphans=False, max_coverage_depth=constants.max_depth_for_coverage):
+        """Loop through the bam pileup and calculate coverage over a defined region of a contig or split
 
-        for pileupcolumn in bam.pileup(split.parent, split.start, split.end, ignore_orphans=ignore_orphans, max_depth=max_coverage_depth):
-            if pileupcolumn.pos < split.start or pileupcolumn.pos >= split.end:
+        Parameters
+        ==========
+        bam : pysam.Samfile
+
+        contig_or_split : anvio.contigops.Split or anvio.contigops.Contig or str
+            If Split object is passed, and `start` or `end` are None, they are automatically set to
+            contig_or_split.start and contig_or_split.end. If str object is passed, it is assumed to
+            be a contig name
+
+        start : int
+            The index start of where coverage is calculated. Relative to the contig, even when
+            `contig_or_split` is a Split object. 
+
+        end : int
+            The index end of where coverage is calculated. Relative to the contig, even when
+            `contig_or_split` is a Split object.
+        """
+
+        if isinstance(contig_or_split, anvio.contigops.Split):
+            contig_name = contig_or_split.parent
+            if not start: start = contig_or_split.start
+            if not end: end = contig_or_split.end
+
+        elif isinstance(contig_or_split, anvio.contigops.Contig):
+            contig_name = contig_or_split.name
+            if not start: start = 0
+            if not end: end = contig_or_split.length
+
+        elif isinstance(contig_or_split, str):
+            contig_name = contig_or_split
+            if contig_name not in bam.references:
+                raise ConfigError('Coverage.run :: Your contig %s was not found in the bam file' % contig_name)
+            if not start: start = 0
+            if not end: end = dict(zip(bam.references, bam.lengths))[contig_name]
+
+        else:
+            raise ConfigError("Coverage.run :: You can't pass an object of type %s as contig_or_split" % type(contig_or_split))
+
+        # a coverage array the size of the defined range is allocated in memory
+        self.c = numpy.zeros(end - start).astype(int)
+
+        for pileupcolumn in bam.pileup(contig_name, start, end, ignore_orphans=ignore_orphans, max_depth=max_coverage_depth):
+            if pileupcolumn.pos < start or pileupcolumn.pos >= end:
                 continue
 
-            self.c[pileupcolumn.pos - split.start] = pileupcolumn.n
+            self.c[pileupcolumn.pos - start] = pileupcolumn.n
 
         if len(self.c):
-            split.explicit_length = len(self.c)
+            try:
+                contig_or_split.explicit_length = len(self.c)
+            except AttributeError:
+                pass
             self.process_c(self.c)
 
 
--- a/anvio/sequence.py
+++ b/anvio/sequence.py
@@ -152,7 +152,7 @@ class Coverage:
         self.mean_Q2Q3 = 0.0
 
 
-    def run(self, bam, contig_or_split, start=None, end=None, method='approximate', **kwargs):
+    def run(self, bam, contig_or_split, start=None, end=None, method='accurate', **kwargs):
         """Loop through the bam pileup and calculate coverage over a defined region of a contig or split
 
         Parameters
@@ -171,6 +171,10 @@ class Coverage:
         end : int
             The index end of where coverage is calculated. Relative to the contig, even when
             `contig_or_split` is a Split object.
+
+        method : string
+            How do you want to calculate? Options: ('accurate', 'approximate'). 'accurate' accounts
+            for gaps in the alignment, 'approximate' does not.
         """
 
         if isinstance(contig_or_split, anvio.contigops.Split):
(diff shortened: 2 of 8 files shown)


ekiefl:
So I ran `anvi-profile` on a large bam file. It's ETA was 1d2h on 1 thread, memory footprint 54GB

In contrast, `anvi-script-get-coverage-from-bam` on the same bam file over all contigs had an ETA of 2d3h, memory footprint 500MB.

I simply can't reason why it's so much slower, but oh well. Maybe you have an idea @meren. I set out to make something faster but I guess what I made instead is a very reasonable way to keep your memory footprint to almost nothing, and also to query coverage levels at various hierarchies directly from bam files.

anvio/dbops.py
```diff
@@ -2950,16 +2950,28 @@ def init_collection_profile(self, collection_name):
...
                 # populate averages per bin
                 averages = copy.deepcopy(samples_template)
+
+                # These weights are used to properly account for differences in split lengths.
+                # Consider table_name == 'mean_coverage', for a bin with 2 splits. Without
+                # weighting, if one split is length 100 with coverage 100 and the other is length
+                # 900 wth coverage 500, the mean_coverage for this bin is (100 + 500)/2 = 300. But
+                # more accurately, mean_coverage of this bin is 100*[100/1000] + 500*[900/1000] =
+                # 460
+                weights = []
```
meren: I like this!

anvio/sequence.py
```diff
@@ -150,28 +152,68 @@ def __init__(self):
...
+            `contig_or_split` is a Split object. 
+
+        end : int
+            The index end of where coverage is calculated. Relative to the contig, even when
+            `contig_or_split` is a Split object.
+        """
+
+        if isinstance(contig_or_split, anvio.contigops.Split):
+            contig_name = contig_or_split.parent
+            if not start: start = contig_or_split.start
```
meren: for some reason I find this more readable in these cases:

``` python
start = contig_or_split.start if not start else start
```

meren:
I haven't seen anything that would explain the memory/speed thingy. the memory needs should change as a function of the number of threads you use with `anvi-profile`, and 54Gb memory for a 15Gb BAM file is truly very depressing :(

I'm still trying to get over the fact that there is no efficient way to go through a range of pileup positions. When someone wants to use a subset of positions to learn about coverage in a rapid fashion, we would set the `--skip-SNV-profiling` `True` since you can't have _some_ SNVs characterized.. But then the `pileup` function is not really very useful and perhaps we could use another function, i.e., `count_coverage`, which makes me wonder if it would be a significantly faster option?

ekiefl:
I have some opinions about this.

1. there's this problem: https://github.com/pysam-developers/pysam/issues/744
2. Second, IIRC I think it was actually slower. I implemented it and it gave different results and I am pretty sure it was slower.

Lastly, I don't know where the magic happens when `--skip-SNV-profiling` is `False`, so does that take place in another pileup loop?

anvio/sequence.py
ekiefl: Ok I can change that

meren:
> Second, IIRC I think it was actually slower. I implemented it and it gave different results and I am pretty sure it was slower.

So it gives you less information that is not correct and it is actually slower. Truly great trade offs. Very good function.

ekiefl:
Here's the results from commit c20d4a5 -- its much slower

image

ekiefl:
How ridiculous would it be to write a primitive bam reader that literally ONLY did coverage and allowed for better iterator objects? We use it to analyze coverage and otherwise we use pysam. Or can you look at the pysam code and see if we can patch it or best case a proper PR?

meren:
I have the feeling that people must have tried to do that many times already and must have come back empty handed. That said, this looks like a likely promising option:

https://github.com/alimanfoo/pysamstats

Here is one of their examples:

``` python
import pysam
import pysamstats

mybam = pysam.AlignmentFile('/path/to/your/bamfile.bam')
a = pysamstats.load_coverage(mybam, chrom='Pf3D7_01_v3', start=10000, end=20000)
```

Easy to install too:

```
pip install pysamstats
```

Although it seems to be using pysam, people claim it is faster. Since nothing makes sense, I feel like I'm OK to take their word for it :/

ekiefl:
Ok I tried pysamtools. It's really really slow. On a 9 second job was 3 seconds slower. It also has capability of counting base pairs, so I'll check that out to see if that implmenetation is faster for cases when we don't --skip-SNV-profiling

image

meren:
:(

Thank you for your thoroughness with this. 

I can see at some point we will have to consider implementing something that is particularly effective when one wants to only learn about coverages from BAM files. Since all we need to know is read start/stop positions in the context of contigs to do this (if you don't give a crap about CIGAR strings (I don't think one should care about it for the application I'm thinking about)), recovering approximate coverages could be done in blazing speeds through a single pass of and would only be limited by I/O and within a very small memory fingerprint (if I had to guess I'd say it would be ~10 seconds per 20,000 nts contig covered around 100X even if we implement it in Python). And if someone turns off the SNV profiling, then profiler would use that procedure to calculate coverages.

ekiefl:
@meren, `analyze_coverage` is called _regardless_ of `--skip-SNV-profling`s value. Therefore, the tool you speak of should be implemented in ALL cases. There is actually a second pileup loop responsible for handling the `--skip-SNV-profling = True` case. I'm looking at that code now and seeing if I can vectorize anything to speed up that case.

meren:
That sounds like some shit design there. Thanks, Meren.

ekiefl:
Ok so when the coverage values are the speed increase is now huge. We're talking 3 day ETA for current master, versus 4 hours in this branch--both on a single thread. Profiling with 10 cores this would be like 20 minutes. 9f3d1c0 was ran for 1000 contigs and produced identical results, which is nice, but its possible the results aren't completely identical.

I want to add a `--fast` flag to `anvi-profile` that skips everything possible (for example, finding outlier coverage, if its possible), and then I think this is ready to merge.

meren:
> That's a 1700% speed increase.

image

Multiple orders of magnitude speed increase _without_ having to go beyond the standard api :) How about the memory fingerprint? Any dramatic change?

> I want to add a --fast flag to anvi-profile that skips everything possible

I think adding a `--profile-only-coverage` flag that simply sets all SNV profiling related flags to skip = True would be the best thing to do. I don't think we will gain much more speed by skipping those steps, but skipping those steps in profiling will make profile dbs less useful for other purposes. If someone is not interested in SNVs, with the `--profile-only-coverage` flag they can enjoy shit ton of speed gain without having to sacrifice anything else :)

meren:
Here is more on this (as an update to what I said above).

Using the start/stop positions of reads via `fetch` without the parsing of CIGAR strings will now show accurate coverage information for positions where there are deletions in short reads compared to the reference (rather than drop in coverages in certain positions as we see when we analyze proper pileups). I think changing the default behavior like this will require much more testing. So I think it is better for now to keep the default behavior as is, and use the new speedy alternative only when the user includes the `--approximate-coverage` flag. With this flag it would use the speedy version of coverage profiling *and* do SNVs. In contrast, the use of `--profile-only-coverage` flag would set both the `--approximate-coverage` and the `--skip-SNV-profiling` flags true, and life would be complete.

If you do this let me know before merging because I will have to write a migration script as we will have to store the information in the self table regarding whether approximate coverage was used (so we can put appropriate warnings for sensitive steps and see when people report issues if it was a speedy profiling to find problems more quickly).

ekiefl:
Memory utilization is hard to pin down because the more memory profile has access to the more it uses. The 54gb from before was on a 200gb core. But on my laptop it was only 4gb? And that went down to 3gb with the new method.

The flag you're describing is `--skip-SNV-profiling`. I want to clarify that this new and improved method is called regardless of whether people skip SNVs, so all users enjoy this speed increase. If they profile SNVs, obviously the percentage speed increase won't be that huge since most time is spent profiling SNVs/SCVs.

So actually I think this is ready to merge...

But as a last step I could look at vectorizing snv profiling since that is another huge bottleneck.

ekiefl:
My response was before reading your latest comment. I'll do that tomorrow

ekiefl:
0b3f835 demonstrates something I'm really surprised by. The pileup function that we have been using produces the same coverages as the new algorithm. That means that the pileup was including reads that are gaps at that position

meren:
I am very surprised to hear this. If the pileup doesn't account for gaps in alignment what is it good for? :(

> It also shows that the proper coverage can be found with little time loss with the function _accurate_routine

From this I realize you have a realistic test data that reveals that. Do you have coverage plots from the inspect page just to compare them qualitatively as well?

ekiefl:
Ok I just made some. There is no difference in the visualization:

Pileup:
image

Read looping with gaps accounted for:
image

But there are 73 positions with mismatched coverages, one of them is position 985. Zooming in, we see the difference:

Pileup:
image

Read looping with gaps accounted for:
image

The code I used to get these numbers is commit

meren:
Very helpful. Thank you, @ekiefl. So revisiting our previous conversation in the light of this new evidence this is what I see:

> The pileup function that we have been using produces the same coverages as the new algorithm.

This means we _could_ switch to the faster algorithm with `fetch` as you had suggested since recovering coverage data this way will not make us lose anything more than what we had been already losing with pileup WRT information content. Am I correct?

You also mentioned that looping through reads with `fetch` and `get_blocks` is not _much more_ slower than `pileup`. I'm curious whether this is also the case for very highly covered contigs. But regardless of the answer of the last question, the `fetch` + `get_blocks` approach could be the alternative way to profile BAM files for those who wish to do a more sensitive analysis of coverage in the expense of additional time that is not far too much compared to previous pileup approach. Right?

... I don't want to open a can of worms now, but I can't help but think about how does the `get_blocks` vs `pileup` regarding their attention to gaps influence our understanding of SNVs for different scenarios (i.e., in/dels in the environment vs in/dels in the reference). We don't even have the 'gap' character in our list of allowed nucleotides. What a mess.

ekiefl:
Ok I think we shouldn't bother giving people the option for `fetch` and should only give `fetch` + `get_blocks`. Here are the times on a highly covered dataset:

`fetch` + `get_blocks`:
image

`get_blocks`:
image

This is in comparison to >2 days for master. If you agree, I will merge this tonight. I just have to proofread

ekiefl:
I agree with your comments about the pileup routine for SNVs, but I may start another branch for that if we go down that route. It is the right time to open up that can of worms I think.

meren:
> Ok I think we shouldn't bother giving people the option for fetch and should only give fetch + get_blocks. Here are the times on a highly covered dataset

It seems they're identical o_O .. probably the reason to that is most reads in this dataset comes as a single block, so fancy indexing is not costing a lot of time. In another dataset most reads may be coming as multiple blocks due to in/dels, in which case a larger performance gap could occur.

But I don't think we should be concerned about it now. I agree with you. Let's don't bother giving distinct options for `fetch` and `fetch` + `get_blocks`.

So no new flags, but just changing the default from `pileup` to `fetch` + `get_blocks`. Right?

ekiefl:
Yeah, the behavior is the same from the user's experience, its just faster now.

Also, I figured out why `anvi-script-get-coverage-from-bam` was slow -- it was due to over-the-top checks. See 477906d. It now does the coverage computed in the anvi-profile job that's 3 hours 50 minutes in 1 hour 30 minutes so its now useful lol. 

I am going to merge now. Very scary. Hopefully no one will notice.

Issue #1345, opened 2020-02-05
anvi-script-process-genbank doesn't process the annotation source and version


LucasPaoli:
Hi,

I ran into the issue with the following version:
```
$anvi-self-test --version
Anvi'o version ...............................: margaret (v5.5)
Profile DB version ...........................: 31
Contigs DB version ...........................: 12
Pan DB version ...............................: 13
Genome data storage version ..................: 6
Auxiliary data storage version ...............: 2
Structure DB version .........................: 1
```

But I think I could track it down and it's probably still an issue.

The `anvi-script-process-genbank` uses the arguments `--annotation-source` and `--annotation-version`

But the GenbankToAnvio function looks for these:
```
line 333        self.source = A('annotations_source') or 'NCBI_PGAP'
line 334        self.version = A('annotations_source') or 'v4.6'
```

I suppose fixing the typo (`A('annotation_source')` and `A('annotation_version')`) should fix it.

Cheers,

Lucas

PS: Also I'm not entirely sure if it's desirable to have default values that are not placeholders altogether

meren:
That's what it looks like, Lucas :)

Would you like to install the `master` version of anvi'o, fix it, and send a PR so your contribution is set on stone? :)

meren:
(you don't even need to install anything, actually. you could even fork anvi'o, clone the master on your computer, commit to your fork after fixing the typo, and then send a PR through GitHub from your fork)

PR #1346, opened 2020-02-05
Fixing typo in anvi-script-process-genbank / GenbankToAnvio args

LucasPaoli:
See https://github.com/merenlab/anvio/issues/1345

This fixes a args issue that prevented anvi-script-process-genbank to write custom annotation sources and versions.


meren:
Thank you very much, @LucasPaoli!

Issue #1349, opened 2020-02-07
One of the commits from 2/5 or 2/6 appears to break anvi-pan-genome


jaybake5:
Hey guys,
I'm new to the community, so if I'm not following protocol with my posts, let me know!  I noticed this issue today so I figured I would report it.

```
anvi-self-test -v
Anvi'o version ...............................: esther (v6.1-master)
Profile DB version ...........................: 31
Contigs DB version ...........................: 14
Pan DB version ...............................: 13
Genome data storage version ..................: 6
Auxiliary data storage version ...............: 2
Structure DB version .........................: 1
```
Installed on Linux server using 'Following the active codebase' (wizard status!)

One of the commits from either 2/5/20 or 2/6/20 appears to make problems for `anvi-pan-genome`.  Either running `anvi-pan-genome` on actual data or `anti-self-test --suite pangenomics` now causes an error each time anvi'o gets to the computing homogeneity step.  The below error is from `anvi-self-test --suite pangenomics`:

```
* Gene clusters are initialized for all 121 gene clusters in the database.

[06 Feb 20 18:17:25 Computing gene cluster homogeneity indices] Initializing 1 threads...                                                                                         ETA: ∞:∞:∞
WARNING
==============================================================
Homogeneity indices computation for gene cluster GC_00000101 failed. This can
happen due to one of three reasons: (1) this gene cluster is named incorrectly,
does not exist in the database, or is formatted into the input dictionary
incorrectly, (2) there is an alignment mistake in the gene cluster, and not all
genes are aligned to be the same lenght; or (3) the homogeneity calculator was
initialized incorrectly. As you can see, this is a rare circumstance, and anvi'o
will set this gene cluster's homogeneity indices to `-1` so things can move on,
but we highly recommend you to take a look at your data to make sure you are
satisfied with your analysis.

Process Process-4:
Traceback (most recent call last):
  File "/usr/local/projdata/0718/projects/jon/conda_envs/github/anvio/anvio/dbops.py", line 1165, in homogeneity_worker
    funct_index, geo_index, combined_index = homogeneity_calculator.get_homogeneity_dicts(gene_cluster)
  File "/usr/local/projdata/0718/projects/jon/conda_envs/github/anvio/anvio/homogeneityindex.py", line 174, in get_homogeneity_dicts
    fun = self.compute_functional_index(cluster_sequences)
[...]

PR #1350, opened 2020-02-07
Amino Acid Conservation Code Clean Up

mahmoudyousef98:
Addresses the bug that caused the fatal crash in #1349. 

I forgot that list.append changes the original list and does not return a new list. All of the ambiguous amino acids (as well as their representative groups) thus had a conservation group of "None" because nothing was returned. @meren 's fix suppressed the error but didn't prevent homogeneity from being miscalculated.  Everything works on my end now.

Lesson learned: rigorous tests on every line of code changed

### DIFF
--- a/anvio/constants.py
+++ b/anvio/constants.py
@@ -239,12 +239,12 @@ conserved_amino_acid_groups = {
     'None': []
 }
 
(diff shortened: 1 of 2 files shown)


meren:
Thank you, @mahmoudyousef98! I assume you run the pangenome tests :)

Issue #1366, opened 2020-03-10
anvi-delete-hmms does not remove genes from genes_in_contigs table, or gene_amino_acid_sequences


ekiefl:

ekiefl:
Solved by #1368

PR #1368, opened 2020-03-10
Delete all remnant info from the dbs

ekiefl:
aims to solve https://github.com/merenlab/anvio/issues/1366

Is this an appropriate change to make? If `anvi-profile` is run, then `anvi-delete-hmms`, there will now be incompatibilities in genes. On the other hand, if you ask to delete an HMM, it seems correct to get rid of the impact it had on the contigs database.

@meren please go ahead and merge if you agree with this change.

### DIFF
--- a/anvio/tables/hmmhits.py
+++ b/anvio/tables/hmmhits.py
@@ -265,7 +265,32 @@ class TablesForHMMHits(Table):
 
 
     def remove_source(self, source):
-        self.delete_entries_for_key('source', source, [t.hmm_hits_info_table_name, t.hmm_hits_table_name, t.hmm_hits_splits_table_name])
+        gene_caller_ids_to_remove = set(key for key, val in self.gene_calls_dict.items() if val['source'] == source)
+
+        tables_with_source = [
+            t.hmm_hits_info_table_name,
+            t.hmm_hits_table_name,
+            t.hmm_hits_splits_table_name,
+            t.genes_in_contigs_table_name,
+            t.gene_function_calls_table_name,
+        ]
+
+        tables_with_gene_callers_id = [
+            t.gene_amino_acid_sequences_table_name,
+            t.genes_taxonomy_table_name,
+        ]
+
+        # delete entries from tables with 'source' column
+        self.delete_entries_for_key('source', source, tables_with_source)
+
+        # delete entries from tables with 'gene_callers_id' column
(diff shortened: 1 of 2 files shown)


meren:
Very nice catch and fix, @ekiefl! Although I'd propose the following change:

``` diff
diff --git a/anvio/tables/hmmhits.py b/anvio/tables/hmmhits.py
index c16aedde..d6ddee77 100644
--- a/anvio/tables/hmmhits.py
+++ b/anvio/tables/hmmhits.py
@@ -265,8 +265,6 @@ class TablesForHMMHits(Table):


     def remove_source(self, source):
-        gene_caller_ids_to_remove = set(key for key, val in self.gene_calls_dict.items() if val['source'] == source)
-
         tables_with_source = [
             t.hmm_hits_info_table_name,
             t.hmm_hits_table_name,
@@ -283,14 +281,19 @@ class TablesForHMMHits(Table):
         # delete entries from tables with 'source' column
         self.delete_entries_for_key('source', source, tables_with_source)

-        # delete entries from tables with 'gene_callers_id' column
-        database = db.DB(self.db_path, utils.get_required_version_for_db(self.db_path))
+        # collect gene caller ids that were added to the db via the HMM source
+        gene_caller_ids_to_remove = set(key for key, val in self.gene_calls_dict.items() if val['source'] == source)

-        CLAUSE = "gene_callers_id in (%s)" % (','.join([str(x) for x in gene_caller_ids_to_remove]))
-        for table in tables_with_gene_callers_id:
-            database.remove_some_rows_from_table(table, CLAUSE)

-        database.disconnect()
+        # if there are any, remove them from tables with 'gene_callers_id' column
+        if len(gene_caller_ids_to_remove):
+            database = db.DB(self.db_path, utils.get_required_version_for_db(self.db_path))
+
+            CLAUSE = "gene_callers_id in (%s)" % (','.join([str(x) for x in gene_caller_ids_to_remove]))
+            for table in tables_with_gene_callers_id:
+                database.remove_some_rows_from_table(table, CLAUSE)
+
+            database.disconnect()


     def append(self, source, reference, kind_of_search, domain, all_genes, search_results_dict):
```

This way we wouldn't be opening and closing the database to run with `gene_callers_id in ()` if the HMM source that is being removed did not add any genes to the genes table.

Apply the patch, take a look, and if it looks OK feel free to merge it :)

ekiefl:
Issue resulting from this merge: https://github.com/merenlab/anvio/issues/1369

Issue #1369, opened 2020-03-11
During creation of HMM entries, HMM entries are deleted


ekiefl:
Here is a minimal reproducible example. Save this fasta as test.fa:

```
>test
TCCCAATAGGATTCAAACCTATAACCTTTTGATCCGTAGTCAAATGCTCTATTCAGTTGAGCTATGGGACCAGTTCAGGA
CTTAGTCCTGTAACTTATTTAGTCTCTGCGTTAGCAGCACCTTCTTCTTTAAGAATCACGAGCTTCTTCTCCTTGATTCT
AGCCTTCTTACCAGAAAGTCCTCTGAGGTAGAAGATTCTTGCTCTACGTACGTGACCAACCTTGTTAACTTCGATAGATT
CGATAGATGGTGACTGGAATGGGAAAATTCTCTCAACACCTACACCGCCTGAAATCTTACGAACTGTGAAAGTCTTCGTG
AATGGTGATGATCCGCTGATCTGGATAACTACACCTCTAAACTTCTGAAGTCTTTCCTTGTCACCCTCTTTAATCTTGAG
GCTTACGGTGATTGTGTCACCTGCCTTGAATTTTGGATAACTTGCAGCTTTCTCGTTTGTAAAAGACTGCTCTACTAATT
TTATCAAATCCATTTCTAAATCACTTTTTAAGCGCAACGTTACTCAAATACTCAACGATTGTAAGAGGTTGCTTTGTTAA
TGGGATTGCAAAGGTACAACCTTTTTTTTAATCTGCAAGCTTTTTTAGAAATTTTCTTTCATTTTTTCGAAAAGATTTCG
ATCCTCGTTGCTGAGCACAGGTGTGAACGTGTCACTTGACCTCATGCCTTCGAGGGCACTCTTCTCGTTGCTCTTCAGCT
TCTTAGGTATCCAGACAAGGTATTTTACATATAGGTCTCCGGTGCCTGAACCATATCCTTTCACTGCAGGAAGTCCACGT
CCTCTCAGACGTACGATTGTTCCGGACTGTGTTCCAGGTTCTACCTTTATCTTATAGTTACCATCTAAGCAAGGTACTGA
AACTTCTCCTCCGAGGATAGCGTCCGTAACAGGTATTATCTTGGTATAGAACAGGTTCTTGCCCTCTCTCTTCAGTGATT
[...]

PR #1371, opened 2020-03-11
Ensure cleanup of contigs databases when new HMM sources are added/updated

meren:
This PR offers a code cleanup that prevents future errors related with HMM addition, removal, and updates to contigs databases independent of target context (whether models work on genes or contigs).

It is compatible with previous contigs databases, so no version update is necessary.

Related to #1369.

### DIFF
--- a/anvio/tables/hmmhits.py
+++ b/anvio/tables/hmmhits.py
@@ -296,6 +296,7 @@ class TablesForHMMHits(Table):
         tables_with_gene_callers_id = [
             t.gene_amino_acid_sequences_table_name,
             t.genes_taxonomy_table_name,
+            t.genes_in_splits_table_name
         ]
 
         # delete entries from tables with 'source' column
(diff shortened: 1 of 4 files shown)


anvio/tables/hmmhits.py
```diff
@@ -276,6 +296,7 @@ def remove_source(self, source):
         tables_with_gene_callers_id = [
             t.gene_amino_acid_sequences_table_name,
             t.genes_taxonomy_table_name,
+            t.genes_in_splits_table_name
```
ekiefl: great catch

ekiefl:
I ran `rm -rf TEST.db; anvi-gen-contigs-database -f test.fa -o TEST.db; anvi-run-hmms -c TEST.db`, verified presence of Transfer_RNAs source in genes_in_contigs, then ran `anvi-delete-hmm --hmm-source Transfer_RNAs -c TEST.db`, verified absence of Transfer_RNAs source in genes_in_contigs.

I'll merge this now. Thank you for fixing this meren. It looks much cleaner now.

PR #1386, opened 2020-03-28
Synteny analysis

mschecht:
Hi Meren, 

I added tests for `anvi-analyze-synteny` to `run_all_tests.sh` and reviewed the code of `synteny.py`, let me know if you have any suggestions.

Cheers,
Matt


anvio/tests/run_all_tests.sh
```diff
@@ -656,6 +655,24 @@ anvi-dereplicate-genomes --ani-dir $output_dir/GENOME_SIMILARITY_OUTPUT \
                          --program pyANI
 SHOW_FILE $output_dir/DEREPLICATION_FROM_PREVIOUS_RESULTS/CLUSTER_REPORT.txt
 
+INFO "Testing anvi-analyze-synteny ignoring genes with no annotation"
+
+# make a external-genomesfile
+echo -e "name\tcontigs_db_path\ng01\t$output_dir/01.db\ng02\t$output_dir/02.db\ng03\t$output_dir/03.db" > $output_dir/external-genomes-file.txt
```
meren: This line is not necessary as there is already an external genomes file at `$output_dir/external-genomes.txt` (see lines above in this file).

anvio/tests/run_all_tests.sh
```diff
@@ -656,6 +655,24 @@ anvi-dereplicate-genomes --ani-dir $output_dir/GENOME_SIMILARITY_OUTPUT \
...
+
+# run anvi-analyze-synteny
+anvi-analyze-synteny -e $output_dir/external-genomes-file.txt \
+          --annotation-source COG_FUNCTION \
+          --window-range 2:3 \
+          -o $output_dir/synteny_output_no_unknowns.tsv
```
meren: This,

```
anvi-analyze-synteny -e $output_dir/external-genomes-file.txt \
          --annotation-source COG_FUNCTION \
          --window-range 2:3 \
          -o $output_dir/synteny_output_no_unknowns.tsv
```

Should be like this for compliance with the rest of the file (again, small things, but they count):

```
anvi-analyze-synteny -e $output_dir/external-genomes-file.txt \
                     --annotation-source COG_FUNCTION \
                     --window-range 2:3 \
                     -o $output_dir/synteny_output_no_unknowns.tsv
```

bin/anvi-analyze-synteny
```diff
@@ -0,0 +1,102 @@
...
+
+__author__ = "Developers of anvi'o (see AUTHORS.txt)"
+__copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
+__credits__ = []
+__license__ = "GPL 3.0"
+__version__ = anvio.__version__
+__maintainer__ = "A. Murat Eren"
+__email__ = "a.murat.eren@gmail.com"
+__requires__ = ['contigs-db']
+__provides__ = ['locus-fasta']
```
meren: does it really provide a `locus-fasta`? I'd say it provides `ngrams`.

bin/anvi-analyze-synteny
```diff
@@ -0,0 +1,102 @@
...
+    groupB = parser.add_argument_group('Additonal INPUT')
+    groupB.add_argument('--analyze-unknown-functions',action="store_true",
+                        help="Provide this flag if you want anvi-analyze-synteny to take into account "
+                              "Ngrams that contain gene calls that have no annotation.")
+
+    args = anvio.get_args(parser)
+
+    try:
+        ngram = NGram(args)
+        ngram.report_ngrams_to_user()
```
meren: Why not just `ngram.report()`?

bin/anvi-analyze-synteny
```diff
@@ -0,0 +1,102 @@
...
+
+
+__author__ = "Developers of anvi'o (see AUTHORS.txt)"
+__copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
+__credits__ = []
+__license__ = "GPL 3.0"
+__version__ = anvio.__version__
+__maintainer__ = "A. Murat Eren"
+__email__ = "a.murat.eren@gmail.com"
+__requires__ = ['contigs-db']
```
meren: I think it requires `external-genomes` and `functions` :)

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+            genes_and_functions_list = self.get_genes_and_functions_from_contigs_db(contigs_db_path)
+
+            # Get unique list of the contigs from this contigsDB (there could be more than one)
+            contigs_set = set(([entry[2] for entry in genes_and_functions_list]))
+
+            for contig_name in contigs_set:
+                contig_function_list = []
+
+                for gci, accession, contigname in genes_and_functions_list:
+                    if contig_name == contigname:
```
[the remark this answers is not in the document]
meren: one more thing: `gene_callers_id` (or `gene_caller_ids` when it is a list) is the common variable name for gene calls throughout the code base. I know `gci` is shorter, but then standardizing core/essential variable name as much as possible helps a lot of things in the long run.

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+        else:
+            self.run_object = terminal.Run(verbose=False)
+
+
+    def sanity_check(self):
+        """Sanity_check will confirm input for NGram class"""
+
+        # checking if the annotation source is common accross all contigs databases
+        g = genomedescriptions.GenomeDescriptions(self.args)
+        g.load_genomes_descriptions(init=False)
```
meren: if you load genomes in `__init__` these two lines will not be necessary when you change the line below to,

```
if self.annotation_source not in self.genomes.function_annotation_sources:
    (...)
```

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+
+        if self.annotation_source not in g.function_annotation_sources:
+            raise ConfigError("The annotation source you requested does not appear to be in the "
+                              "contigs database :/ "
+                              "Please confirm you are only providing one annotation source :)")
```
meren: This error message needs to be updated. `g.function_annotation_sources` shows sources for functions that are common to all contigs databases. if a functional annotation source is not in this list does not mean 'a contigs database' is missing it, it means 'not all contigs databases' has it.

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+
+        if not self.args.output_file:
+            raise ConfigError("You should provide an output file name.")
+
+        # checking window-range input
+        if ":" not in self.window_range:
+            raise ConfigError("anvi'o would love to slice and dice your loci, but... the "
+                              "Format of window_range must be x:y (e.g. Window sizes 2 to 4 would be denoted as: 2:4)")
+
+        self.window_range = [int(n) for n in self.window_range.split(":")]
```
meren: What happens when you send `a:b` as window range? :) Probably `int(n)` will explode, right?

So this line could have been in a try/except block to avoid that, and if there is a `TypeError`, the user could be warned to use only integers.

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+
+        self.args = args
+        self.run = run
+        self.progress = progress
+
+        # Parse arguments
+        A = lambda x: args.__dict__[x] if x in args.__dict__ else None
+        self.external_genomes = A('external_genomes')
+        self.annotation_source = A('annotation_source')
+        self.window_range = A('window_range')
```
meren:
```
self.window_range = A('window_range') or 500
```

May be better. `anvio/__init__.py` has a default for this, but it is not the only way people can use this class. if I get an instance directly from the module (and not through `anvi-analyze-synteny` program) I wouldn't know what is the programmer suggested default for this.

Unfortunately we are not always careful about setting proper defaults for class variables throughout the codebase, but I wanted to point it out here anyway.

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+            if not self.in_in_unknowns_mode and "unknown-function" in ngram: # conditional to record NGrams with unk functions
+                continue
+            else:
+                # if ngram is not in dictionary add it, if it is add + 1
+                if ngram in NGramFreq_dict:
+                    NGramFreq_dict[ngram] +=  1
+                else:
+                    NGramFreq_dict[ngram] = 1
+
+        return NGramFreq_dict
```
meren: very mixed :)

It should either be `NGramFreqDict` (which is a better class name than variable name) or `ngram_frequencies_dict` (which is what you mean here).

anvio/__init__.py
```diff
@@ -2019,6 +2019,15 @@ def get_args(parser):
                      "class based on the number of threads. If you have *any* hesitation about whether you know what you are doing, "
                      "you should not change this value."}
                 ),
+    'window-range': (
+            ['--window-range'],
+            {'default': 500,
+             'metavar': 'WINDOW_RANGE',
```
meren: `metavar` could be `INT`

bin/anvi-analyze-synteny
```diff
@@ -0,0 +1,102 @@
...
+from anvio.errors import ConfigError, FilesNPathsError
+
+
+__author__ = "Developers of anvi'o (see AUTHORS.txt)"
+__copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
+__credits__ = []
+__license__ = "GPL 3.0"
+__version__ = anvio.__version__
+__maintainer__ = "A. Murat Eren"
+__email__ = "a.murat.eren@gmail.com"
```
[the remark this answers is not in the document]
mschecht: @meren thank you so much for your in-depth comments! I'll try to address everything by the end of the day and make a loaf of bread at the same time :)

anvio/__init__.py
```diff
@@ -2019,6 +2019,15 @@ def get_args(parser):
                      "class based on the number of threads. If you have *any* hesitation about whether you know what you are doing, "
                      "you should not change this value."}
                 ),
+    'window-range': (
```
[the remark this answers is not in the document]
mschecht: I agree it's more clear

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+        self.run = run
+        self.progress = progress
+
+        # Parse arguments
+        A = lambda x: args.__dict__[x] if x in args.__dict__ else None
+        self.external_genomes = A('external_genomes')
+        self.annotation_source = A('annotation_source')
+        self.window_range = A('window_range')
+        self.in_in_unknowns_mode = A('analyze_unknown_functions')
+        self.external_genomes = utils.get_TAB_delimited_file_as_dictionary(self.external_genomes)
```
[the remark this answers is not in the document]
mschecht: This is great! just added

anvio/synteny.py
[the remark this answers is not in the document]
mschecht: Wow, my variable naming was a disaster waiting to happen!

This was challenging to name but I hope this is better. I added a more in-depth comment to address confusion

```
# Extract gene_callers_id and gene_function_accession from genes_and_functions_list using contig_name
# genes_and_functions_list has multiple contigs gene info in it, so we filter one contig at a time 
# (contig_ID = name of contig in genes_and_functions_list)
for gene_callers_id, gene_function_accession, contig_ID in genes_and_functions_list:
   	if contig_name == contig_ID:
        contig_function_list.append([gene_callers_id,gene_function_accession])
```

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+
+        return ngram_count_df_final
+
+
+    def report_ngrams_to_user(self):
+        """Counts ngrams per contig and reports as tab-delimited file"""
+
+        self.populate_genes()
+        df = self.convert_to_df()
+        df.to_csv(self.output_file, sep = '\t',index=False)
```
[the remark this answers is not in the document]
mschecht: I'm all about it!

anvio/synteny.py
```diff
@@ -0,0 +1,303 @@
...
+            ngram = tuple(window)
+
+            if not self.in_in_unknowns_mode and "unknown-function" in ngram: # conditional to record NGrams with unk functions
+                continue
+            else:
+                # if ngram is not in dictionary add it, if it is add + 1
+                if ngram in NGramFreq_dict:
+                    NGramFreq_dict[ngram] +=  1
+                else:
+                    NGramFreq_dict[ngram] = 1
```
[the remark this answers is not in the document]
mschecht: Wow that's super clean, good to know!

anvio/synteny.py
mschecht: yeah using `self.genomes` solved all these problems, really glad I know this now

anvio/__init__.py
mschecht: Sorry Meren I tried replacing `str` with `INT` but got an error. The user is supposed to enter in x:y (x = smallest window size, y = largest window size) which gets parsed in `sanity check` as a string. So I think using `INT` won't work, but honestly I don't think I understand `metavar` enough.

meren: metavar is nothing but giving the user an idea about what is expected as input for that variable. You can write anything there and it will show up in the help menu output.

anvio/synteny.py
mschecht: Yup just tested this.

What do you think of this solution:

```
self.window_range = [n for n in self.window_range.split(":")]

if not isinstance(self.window_range[0], int) or not isinstance(self.window_range[1], int):
        raise ConfigError("anvi'o would love to slice and dice your loci, but the "
            "window-ranges need to be integers :)")
```

mschecht: nvm that solution didnt work. I understand what your saying about try/except

How's this:

```
try:
    self.window_range = [int(n) for n in self.window_range.split(":")]
except:
    raise ConfigError("anvi'o would love to slice and dice your loci, but the "
                              "window-ranges need to be integers :)")
```

bin/anvi-analyze-synteny
mschecht: Are you referring to renaming the method `report_ngrams_to_user` to `report`?

anvio/synteny.py
mschecht: Sorry Meren this concept was a little over my head but I think I understand the idea of default values for instances. I tried to fix this with this:

I added this to `synteny.py`:

```
 self.window_range = A('ngram_window_range') or "2:3"
```

and this to `anvio/__init__.py`:

```
'ngram-window-range': (
            ['--ngram-window-range'],
            {'default': "2:3",
             'metavar': "NGRAM_WINDOW_RANGE",
             'type': str,
             'required': False,

```

anvio/tests/run_all_tests.sh
mschecht: got it, just removed it and the test is still working :)

mschecht:
@meren I wasn't quite sure about conventions for printing messages to the user. I thought it would be nice to report that `anvi-analyze-synteny` worked and where to find the output file. What do you think of [this](https://github.com/merenlab/anvio/blob/synteny-analysis/anvio/synteny.py#L216-L217)?

anvio/synteny.py
meren: Great. But the best practice is this: You are trying to catch an error due to conversion of a value by the 'int' function. So this is what I do:

``` python
$ python
Python 2.7.10 (default, Aug 17 2018, 19:45:58)
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> int('a')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: 'a'
>>>
```

So this tells me that I should `expect` a `ValueError` if someone sends a string instead of an int. This way if there is something else is wrong, it would be reported full, rather than saying "your window is not integer". Makes sense?

(so basically I'm saying change `expect:` to `expect ValueError:`).

meren:
I think a single run.info that mentions the output file path would be enough. The line above is redundant with that information :)

bin/anvi-analyze-synteny
```diff
@@ -0,0 +1,102 @@
...
+
+
+__author__ = "Developers of anvi'o (see AUTHORS.txt)"
+__copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
+__credits__ = []
+__license__ = "GPL 3.0"
+__version__ = anvio.__version__
+__maintainer__ = "Matthew S. Schechter"
+__email__ = "mschechter@uchicago.edu"
+__requires__ = ['contigs-db', 'external-genomes', 'annotation-source', 'ngram-window-range']
```
meren: we can't just put any keyword here :) __requires__ and __provides__ must be defined here:

https://github.com/merenlab/anvio/blob/master/anvio/programs.py#L38

mschecht: ahhhh I didn't know that! Ok great Ill fix that

mschecht: fixed!

anvio/synteny.py
mschecht: Gotcha, I had no idea one could be that specific with errors!

meren:
Thanks for all the changes, @mschecht!

These will be in `v6.2`.

PR #1390, opened 2020-03-30
Add trna scan to contig workflow

mschecht:
Hi Meren, I ran the contigs workflow with the newly added `anvi_run_trna_scan` on a staph genome and everything seemed checkout. FYI I left the `anvi_run_trna_scan` to run in the default snakemake config file so it would be easier to test immediately. There is also one FIXME where I wasn't sure how to work with one parameter in `anvi_run_trna_scan`, the `--trna-hits-file` parameter. 

Cheers,
Matt


meren:
This looks good to me. Thank you, @mschecht. We can address the output file thingy if it ever becomes necessary. Bu

PR #1391, opened 2020-03-31
Search via diamond instead of MODELLER's in-house solution

ekiefl:
This is so fast its crazy and provides equal accuracy +-1 %


ekiefl:
If its not too late, I would like this to be in v6.3

meren:
Excellent timing! :) It will be in as I am planning to tag `v6.2` tomorrow morning.

PR #1394, opened 2020-04-03
Indel

ekiefl:
PR opens up several new doors into the anvi-profile world.

## indels can be stored in DB

With `--profile-indels`, one can store each indel. Here is the table structure:

image

## percentage ID filtering during anvi-profile

With `--min-percent-identity`, one can filter reads based on percentage identity.

### DIFF
--- a/sandbox/anvi-script-get-coverage-from-bam
+++ b/sandbox/anvi-script-get-coverage-from-bam
@@ -36,7 +36,7 @@ class ComputeCoverage(object):
         self.sanity_check()
         self.f = self.init_output()
 
-        self.bam = pysam.Samfile(self.args.bam_file, 'rb')
+        self.bam = bamops.BAMFileObject(self.args.bam_file, 'rb')
 
         # lengths is {contig_name: contig_length} of each contig in the bam
         self.lengths = dict(zip(self.bam.references, self.bam.lengths))
@@ -123,11 +123,12 @@ class ComputeCoverage(object):
         ==========
         contig_name : str
         """
+
         self.progress.increment()
         self.progress.update('%s (%d/%d)' % (contig_name, self.progress.progress_current_item, self.progress.progress_total_items))
 
         cov = bamops.Coverage()
-        cov.run(self.bam, contig_name, skip_coverage_stats=True)
+        cov.run(self.bam, contig_name, read_iterator='fetch', skip_coverage_stats=True)
 
         return cov.c
 
--- a/sandbox/anvi-script-get-coverage-from-bam
+++ b/sandbox/anvi-script-get-coverage-from-bam
@@ -166,6 +166,7 @@ class ComputeCoverage(object):
             self.process_collection_txt()
 
         self.progress.end()
+        self.f.close()
         self.run.info_single('Output file %s has been written' % self.args.output, nl_before=1)
 
 
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -376,6 +376,15 @@ D = {
                      "computational complexity, this feature comes 'off' by default. Using this flag you can rise against the "
                      "authority, as you always should, and make anvi'o profile codons."}
                 ),
+    'profile-indels': (
+            ['--profile-indels'],
+            {'default': False,
+             'action': 'store_true',
+             'help': "The alignment of a read to a reference genome/sequence can be imperfect, such that the read exhibits "
+                     "insertions or deletions relative to the reference. Anvi'o normally accounts for this information "
+                     "but does not _store_ it in the profile database. If you would like this information, you could use "
+                     "this flag. It comes at significant computational cost."}
+                ),
     'drop-previous-annotations': (
             ['--drop-previous-annotations'],
             {'default': False,
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -383,7 +383,8 @@ D = {
              'help': "The alignment of a read to a reference genome/sequence can be imperfect, such that the read exhibits "
                      "insertions or deletions relative to the reference. Anvi'o normally accounts for this information "
                      "but does not _store_ it in the profile database. If you would like this information, you could use "
-                     "this flag. It comes at significant computational cost."}
+                     "this flag. It comes at significant computational cost. If --skip-SNV-profiling is set, --profile-indels "
+                     "will be unset."}
                 ),
     'drop-previous-annotations': (
             ['--drop-previous-annotations'],
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -358,6 +358,16 @@ D = {
                      "will instruct profiler to skip that step. Please remember that parameters and flags must be "
                      "identical between different profiles using the same contigs database for them to merge properly."}
                 ),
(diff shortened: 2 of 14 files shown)


ekiefl:
# Checking accuracy of indels

Here are the first couple in my test dataset compared against IGV

image

Here are screenshots of the regions read from top to bottom.

image

image

image

image

image

image

image

image

image

image

## Verdict

Everything I have checked out so far has worked. There is obviously a shit-load we can do with this, but for now in this PR I simply want to make sure the information reported is correct and succinct.

sandbox/anvi-script-get-coverage-from-bam
```diff
@@ -36,7 +36,7 @@ class ComputeCoverage(object):
         self.sanity_check()
         self.f = self.init_output()
 
-        self.bam = pysam.Samfile(self.args.bam_file, 'rb')
+        self.bam = bamops.BAMFileObject(self.args.bam_file, 'rb')
```
meren, line 39: does this mean this previous version of this line will explode in tagged `v6.2.`? :) asking for a friend.

meren:
profile db version is one up, there is a migration script, I just tested on infant gut data and it seems to be working. what is the added time to profiling? you had the performance tests from the previous profiler revamping effort. did you re-run any of them?

one problem is that our online documentation fails to represent any of these additions. we will be fine, but this will impact others' ability to use or understand these improvements. I am not sure what is the best way to solve this :/

thank you Evan! great contribution, and I'm certain we will find great ways to exploit the information in the indels table.

meren:
@ekiefl, please update your branch and run mini-test :)

ekiefl:
> @ekiefl, please update your branch and run mini-test :)

```
✖ anvi-profile encountered an error after 0:22:57.205934
  File "/Users/evan/Software/anvio/bin/anvi-profile", line 98, in <module>
    main(args)
  File "/Users/evan/Software/anvio/anvio/terminal.py", line 748, in wrapper
  File "/Users/evan/Software/anvio/bin/anvi-profile", line 33, in main
    profiler.BAMProfiler(args)._run()
  File "/Users/evan/Software/anvio/anvio/profiler.py", line 283, in _run
    self.profile_single_thread()
  File "/Users/evan/Software/anvio/anvio/profiler.py", line 731, in profile_single_thread
    contig = self.process_contig(bam_file, contig_name, contig_length)
  File "/Users/evan/Software/anvio/anvio/profiler.py", line 685, in process_contig
    split.auxiliary.process(bam_file)
  File "/Users/evan/Software/anvio/anvio/contigops.py", line 210, in process
    self.run_SNVs_and_indels(bam)
  File "/Users/evan/Software/anvio/anvio/contigops.py", line 452, in run_SNVs_and_indels
    for read in read_iterator(self.split.parent, self.split.start, self.split.end, **kwargs):
  File "/Users/evan/Software/anvio/anvio/bamops.py", line 79, in fetch_and_trim
    read = Read(read)
  File "/Users/evan/Software/anvio/anvio/bamops.py", line 153, in __init__
    self.reference_sequence = np.frombuffer(read.get_reference_sequence().upper().encode('ascii'), np.uint8)
  File "pysam/libcalignedsegment.pyx", line 1841, in pysam.libcalignedsegment.AlignedSegment.get_reference_sequence
  File "pysam/libcalignedsegment.pyx", line 864, in pysam.libcalignedsegment.build_reference_sequence
ValueError: MD tag not present
```

What kind of diabolical .bam files are these? MD tags specify info about the reference bases at the aligned positions. I can create a pre-test in `profiler.py` that opens up a handful of reads from the BAM file and if none of them contain an MD tag, I will run a warning and remove `--min-percent-identity`. What do you think?

meren:
> I can create a pre-test in profiler.py that opens up a handful of reads from the BAM file and if none of them contain an MD tag, I will run a warning and remove --min-percent-identity.

I think this would be a great way to solve it. By removing it, I assumed you mean setting `--min-percent-identity` to `None` or `0.0` in cases where it can't be applicable.


> It would helpful if these BAM files were updated so --min-percent-identity could be tested in mini test.

To be honest I hate these BAM files because they are ancient. But updating them will require A LOT of work, so I've been being lazy about it. But I will put an issue now and this will be addressed sooner or later.

ekiefl:
> To be honest I hate these BAM files because they are ancient. But updating them will require A LOT of work, so I've been being lazy about it. But I will put an issue now and this will be addressed sooner or later.

Yeah, _sooooo_ much of our tests are reliant on these lil guys so I can imagine. We may hate them but they discovered an important bug :)

meren:
True. I'm actually sending some virtual hugs to those shitty-but-ours BAM files.

sandbox/anvi-script-get-coverage-from-bam
ekiefl: Fortunately, no. `Samfile` is a deprecated but functioning class as of pysam 0.15.4... 

But I'm realizing that the `requirements.txt` has 0.15.2.... What should we do?

ekiefl: Thanks for the catch. Fixed in 5f6632c

ekiefl:
Ok I measured the time taken and database sizes for the following cases.

1. `anvi-profile` in master branch

Time taken: `✓ anvi-profile took 0:30:27.722558`
DB size: 2.4GB

2. `anvi-profile` in indel branch

Time taken: `✓ anvi-profile took 0:31:46.993346`
DB size: 2.4GB

Things are just slower. Ok, fine. I think it's because of a new nested yield in `Coverage` and because `Read.__init__` has one more pysam attribute lookup and an `if` statement. Amazing that there are so many reads that this adds up to a minute.

2. `anvi-profile --profile-indels`

Time taken: `✓ anvi-profile took 0:31:49.627898`
DB size: 2.6GB

The cost is almost nothing to `--profile-indels`. And the DB size is only 200MB larger.

3. `anvi-profile --min-percent-identity 95`

Time taken: `✓ anvi-profile took 0:35:12.351169`
DB size: 1.5GB

The DB is smaller because there are less SNVs. The time cost for filtering is appreciable but not insane and probably not noticeable to the user.

---------------------------------------

Ok, based on these timings, I think we should get rid of `--profile-indels` and the INDEL calculation should piggy-back on `--skip-SNV-profiling`: If `--skip-SNV-profiling`, they don't get SNVs or INDELs. Otherwise they get both. Does this seem reasonable @meren? After this is settled I will merge.

meren:
> Ok, based on these timings, I think we should get rid of --profile-indels and the INDEL calculation should piggy-back on --skip-SNV-profiling: If --skip-SNV-profiling, they don't get SNVs or INDELs.

I agree with this! Probably INDEL time/space requirements will change as a function of the distance between environmental populations and the genome that recruits reads from them. But I like the idea to make it the default setting regardless. 

I also agree with `--skip-SNV-profiling` turning off INDEL profiling, but I think there should also be an independent `--skip-INDEL-profiling` flag. I think it is a better practice.

Other than this, I think it is ready to merge as far as I can see :)

ekiefl:
Ok here goes!

PR #1413, opened 2020-04-22
Metabolic reconstruction

ivagljiva:
Dear @meren, @ekiefl, @watsonar, @semiller10 (and anyone else who is interested),

The purpose of this pull request is primarily to merge the metabolic estimation code that I have so far, but also includes a couple of other related improvements to the codebase. I would be grateful if you could review it and provide some feedback. 

The major changes you will find here are:
- the addition of the Kegg class and 3 scripts for the metabolism estimation workflow as I described it in lab meeting on Monday
  - metabolism estimation output includes module completeness and several basic measures for module redundancy
  - there is version tracking for the MODULES.db as discussed in lab meeting
- performance modifications to `anvi-run-pfams` and `anvi-run-hmms`
  - there is now the option to use `hmmsearch` rather than `hmmscan`. For pfams, `hmmsearch` is now the default
  - `hmmpress` is run during pfams setup rather than unpacking the profiles into a temporary directory before doing the HMM search

Thank you very much!


anvio/parsers/hmmscan.py
```diff
@@ -49,20 +66,76 @@ def __init__(self, hmm_scan_hits_txt, alphabet='AA', context='GENE'):
...
+                    is going on. But for now, anvi'o is going to keep all hits to this profile. Consider those hits with a grain of salt, \
+                    as not all of them may be good." % hit['gene_name'])
+                    entry = {'entry_id': entry_id,
+                             'gene_name': hit['gene_name'],
+                             'gene_hmm_id': hit['gene_hmm_id'],
+                             'gene_callers_id': hit['gene_callers_id'],
+                             'e_value': hit['e_value']}
+
+                else:
+                    # but in Pfams, we don't care, we just keep all hits
```
meren, line 133: just to note, we keep all hits in this case because some HMMs carry around their own noise cutoff terms, and anvi'o enforces the user who defines an HMM profile directory to explicitly state how noise cutoff must be done before things come to this section of the code :) It is not specific to PFAMs.

anvio/parsers/hmmscan.py
```diff
@@ -49,20 +66,76 @@ def __init__(self, hmm_scan_hits_txt, alphabet='AA', context='GENE'):
         Parser.__init__(self, 'HMMScan', [hmm_scan_hits_txt], files_expected, files_structure)
 
 
-    def get_search_results(self):
+    def get_search_results(self, ko_list_dict = None):
```
meren, line 69: I don't think we should call this variable `ko_list_dict`. this basically is a dictionary that describes noise cutoff thresholds for each HMM individually. it can be used for KEGG KOs, but the concept is more general than KEGG (and this module doesn't know about KEGG, so we don't have to be too specific).

So the variable `ko_list_dict` can be renamed to `noise_cutoff_dict`, `knum` to `hmm_entry_name`, etc. Furthermore, error messages should be agnostic to KEGG / KO, too.

anvio/pfam.py
```diff
@@ -163,8 +189,32 @@ def __init__(self, args, run=run, progress=progress):
...
 
     def is_database_exists(self):
-        if not os.path.exists(os.path.join(self.pfam_data_dir, 'Pfam-A.hmm.gz')):
+        """
+        This function verifies that pfam_data_dir contains the Pfam hmm profiles and checks whether they are compressed or not.
+        If they are compressed, we decompress them and run hmmpress.
+
+        PARAMETERS: N/A
+
+        RETURNS: N/A
```
meren, line 198: there is no need for these since the fact that the function does not take any parameters is obvious from the header :)

anvio/utils.py
```diff
@@ -381,6 +382,25 @@ def gzip_decompress_file(input_file_path, output_file_path=None, keep_original=T
 
     return output_file_path
 
+def tar_extract_file(input_file_path, output_file_path=None, keep_original=True):
+    filesnpaths.is_file_exists(input_file_path)
+
+    if not tarfile.is_tarfile(input_file_path):
```
meren, line 388: this should go into `filesnpaths` module so it is more broadly accessible to the codebase.

meren: I mean `tarfile.is_tarfile` and the error message if not should be accessible as `filesnpaths.is_tar_file`

bin/anvi-setup-kegg-kofams
```diff
@@ -0,0 +1,45 @@
...
+from anvio.errors import ConfigError, FilesNPathsError
+from anvio.terminal import time_program
+
+__author__ = "Developers of anvi'o (see AUTHORS.txt)"
+__copyright__ = "Copyleft 2015-2020, the Meren Lab (http://merenlab.org/)"
+__license__ = "GPL 3.0"
+__version__ = anvio.__version__
+__maintainer__ = "Iva Veseli"
+__email__ = "iveseli@uchicago.edu"
+__provides__ = ["kofam-data", "kegg-modules-db",]
```
meren, line 18: I think it only provides `kegg-db` (`kegg-modules-db` is too wordy). in addition, as far as I can see, we can remove `kofam-data` from every provides / requires statements, and things wold make more sense :) ignore this if you don't agree. 

Plus, please run `anvi-script-gen-programs-network` in your branch.

anvio/kegg.py
```diff
@@ -0,0 +1,2346 @@
...
+
+        if os.path.exists(self.kofam_hmm_file_path):
+            raise ConfigError("It seems you already have KOfam HMM profiles installed in '%s', please use --reset flag if you want to re-download it." % self.kegg_data_dir)
+
+        if os.path.exists(self.kegg_module_file):
+            raise ConfigError("Interestingly, though KOfam HMM profiles are not installed on your system, KEGG module information seems to have been \
+            already downloaded in %s. Please use the --reset flag to re-download everything from scratch." % self.kegg_data_dir)
+
+        if os.path.exists(self.kegg_pathway_file):
+            raise ConfigError("Interestingly, though KOfam HMM profiles are not installed on your system, KEGG pathway information seems to have been \
```
meren, line 204: we don't care too much about line length in general, but these lines are exceptionally long. plus all these raise statements are missing alignments and use `\` for new lines. I also saw other `run.warnings` statements that suffered from similar things :) I will not point out everything, but please take another look at your modules.

anvio/kegg.py
```diff
@@ -0,0 +1,2346 @@
...
+        self.kegg_modules_db = KeggModulesDatabase(os.path.join(self.kegg_data_dir, "MODULES.db"), args=self.args)
+
+
+    def set_hash_in_contigs_db(self):
+        """Modify the contigs DB self table to indicate which MODULES.db has been used to annotate it"""
+        A = lambda x: self.args.__dict__[x] if x in self.args.__dict__ else None
+        self.contigs_db_path = A('contigs_db')
+
+        contigs_db = ContigsDatabase(self.contigs_db_path)
+        contigs_db.db.set_meta_value('modules_db_hash', self.kegg_modules_db.db.get_meta_value('hash'))
```
meren, line 612: what if there is an existing one?

meren:
Amazing work, @ivagljiva. I see a lot of very good programming practices in here. thank you for that and I can't wait for people to start playing with this new anvi'o toy!

I will send some relatively minor comments for things I saw as I went through the code in a second for your consideration.

ivagljiva:
Wow, @meren, thank you for reviewing so thoroughly and quickly! I'll get to work on fixing these things :)

anvio/drivers/hmmer.py
```diff
@@ -70,40 +99,24 @@ def run_hmmscan(self, source, alphabet, context, kind, domain, num_genes_in_mode
...
         self.run.info('Number of CPUs will be used for search', self.num_threads_to_use)
+        if alphabet in ['DNA', 'RNA']:
+            self.run.info('HMMer program used for search', 'nhmmscan')
+        else:
+            self.run.info('HMMer program used for search', self.program_to_use)
 
-        # we want to create hmm files in the same direcotry
         tmp_dir = os.path.dirname(self.target_files_dict[target][0])
-        log_file_path = os.path.join(tmp_dir, '00_log.txt')
+        log_file_path = os.path.join(tmp_dir, '*_log')
```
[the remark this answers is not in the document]
ivagljiva: I changed this because with multithreading, each hmmscan worker thread gets its own log. (and, even with just 1 thread there is never a log file called '00_log.txt'. Here is an example:
```
$ ls /var/folders/1n/2s6d_kq53pv9js812zwcljq80000gn/T/tmphz7r15hp/*log
/var/folders/1n/2s6d_kq53pv9js812zwcljq80000gn/T/tmphz7r15hp/AA_gene_sequences.fa.0_log /var/folders/1n/2s6d_kq53pv9js812zwcljq80000gn/T/tmphz7r15hp/AA_gene_sequences.fa.2_log
/var/folders/1n/2s6d_kq53pv9js812zwcljq80000gn/T/tmphz7r15hp/AA_gene_sequences.fa.1_log /var/folders/1n/2s6d_kq53pv9js812zwcljq80000gn/T/tmphz7r15hp/AA_gene_sequences.fa.3_log
```

This particular variable is used just for printing run info, whereas the actual log file names are generated later in the code (in the for loop that spawns the worker threads). So at the time it was easiest to just update `log_file_path`.

I changed it to have a literal `*` so that copying the file path into an `ls` command would allow you to list all the log files, but I do agree that it would be best to print the actual log file names for each thread. I will change that now.

anvio/drivers/hmmer.py
```diff
@@ -119,17 +132,28 @@ def run_hmmscan(self, source, alphabet, context, kind, domain, num_genes_in_mode
...
-                        '--tblout', shitty_file,
-                        hmm_file_path, part_file]
+            if noise_cutoff_terms:
+                cmd_line = ['nhmmscan' if alphabet in ['DNA', 'RNA'] else self.program_to_use,
+                            '-o', output_file, *noise_cutoff_terms.split(),
+                            '--cpu', cores_per_process,
+                            '--tblout', shitty_file,
+                            hmm, part_file]
+            else: # if we didn't pass any noise cutoff terms, here we don't include them in the command line
+                cmd_line = ['nhmmscan' if alphabet in ['DNA', 'RNA'] else self.program_to_use,
```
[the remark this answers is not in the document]
ivagljiva: Yes, totally. There is a warning for this in the code above. :)
```
if alphabet in ['DNA', 'RNA'] and self.program_to_use == 'hmmsearch':
            self.run.warning("You requested to use the program `%s`, but because you are working with %s sequences Anvi'o will use `nhmmscan` instead. "
                             "We hope that is alright." % (self.program_to_use, alphabet))
```

anvio/kegg.py
```diff
@@ -0,0 +1,2346 @@
...
+            if parens_level < 0:
+                return False
+        splits.append(d[last_split_index:len(d)])
+
+        if return_delims:
+            return splits, delim_list
+        return splits
+
+
+    def recursive_definition_unroller(self, step):
```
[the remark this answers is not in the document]
ivagljiva: yes. unfortunately that is going to be the case. sorry :P

bin/anvi-setup-kegg-kofams
ivagljiva: Meren, I ran `anvi-script-gen-programs-network` and got a `NETWORK.json` file. Is there a specific place I should put this file in the `anvio` or `web` repository? :)

meren: no :) I just was just wondering whether you'd get an error for missing `kegg-modules-db` description in programs.py

ivagljiva: I did before, but now I have fixed that :)

PR #1421, opened 2020-05-04
Update news button

isaacfink21:
This PR updates the news button on the interactive interface to attract more attention, by making it bigger and activated on mouse hover when unread.


meren:
Hey @isaacfink21, thanks for this :) I have a couple of suggestions.

One, I think you should implement this change in your branch so it is easier to know what to do to stop the button being annoying:

``` diff
diff --git a/anvio/data/interactive/index.html b/anvio/data/interactive/index.html
index f402b244..f832fb94 100644
--- a/anvio/data/interactive/index.html
+++ b/anvio/data/interactive/index.html
@@ -797,7 +797,10 @@
     </div>

     <div id="news-panel">
-        <h1 style="font-size: 22px; display: inline-block;">What's New</h1><a href="#" style="font-weight: bold; font-size: 14px; color: gray; margin-top: 36px; float: right;" onclick="newsMarkRead(); ">Mark everything as read&nbsp;</a>
+        <h1 style="font-size: 22px; display: inline-block;">What's New</h1>
+        <p style="border: 1px dashed #b8b8b8; background: #f0f8ff2e; border-radius: 4px; padding: 6px;">
+            <a href="#" style="font-weight: bold; color: red;" onclick="newsMarkRead();">Mark everything as read</a> to make the annoying news button to stop glowing in red.
+        </p>
         <div id="news-panel-inner"></div>
     </div>
     <!-- DIALOGS SECTION -->
``` 

After making that change, I think it would be best if that `p` only appeared when there were things unread (if someone clicks the news button when there is nothing unread, they shouldn't see the "Mark everything as read" box).

Finally, a small cosmetics suggestion. I think it now catches the eye for sure, but instead of a bigger glowing button, perhaps a blinking button (without the glow) would be more more elegant way to ask for attention. A CSS like this should give that to you (you can find examples of fade animation on objects online):

``` css
.fade {
  animation: fade 5s infinite;
}

@keyframes fade {
  25% {
    opacity: 0;
  }
  50% {
    opacity: 1;
  }
  75% {
    opacity: 0;
  }
  100% {
    opacity: 1;
  }
}
```

Thank you again! :)

isaacfink21:
Thanks for the suggestions! Here is an updated version with all the implementations :)

meren:
Great! Thank you :) Merging this to master, now.

PR #1425, opened 2020-05-12
Anvi'o online help on steroids

meren:
This PR introduces code into anvi'o that generates these web pages automatically:

http://merenlab.org/software/anvio/help/

To generate a static web page, run the following command in any directory:

```
anvi-script-gen-help-pages
```


meren:
(merging it now, but we can continue reviewing it)

PR #1428, opened 2020-05-15
Orf frame

ekiefl:
# The problem

When genes are not divisible by 3 (or when genes are partial), we don't know which codon frame to use. This is a serious problem. For example, one can use prodigal to call genes and end up with 32k genes, 5k of which are partial. When the gene calls are exported and then used as external genes, the partial genes cannot be translated--we are forced to pass the amino acid sequences during exportation. As another example, if an external gene call is passed where the sequence is not divisible by 3, it may be a relatively good gene call, but for whatever reason it is missing 1 or 2 bases. How can we recover from this?

# Solution

The hypothesis is that amino acid sequences have signatures that make them look like genuine, naturally occurring proteins. Aspects like secondary structure restrict amino acid sequences from being totally random. This ends up being a very true hypothesis. Therefore, if we can get a grasp on what a "normal looking" sequence looks like, we can translate all 3 possible frames and decide which one looks most normal (the one that has the highest probability of occurring).

To attribute a probability score to a sequence, I developed a Markov model that describes the probability of seeing strings of amino acids together. For example, you're reading the sequence and see 'A', what is the probability of seeing 'V' in the next position? The probability for the whole sequence is the product of transition probabilities.

The transition probabilities were built empirically from the uniprot50 protein sequence dataset (14.9 billion amino acids, 15GB fasta). These models are stored as `.npy` files in `data/seq_transition_models/AA`. 

# Testing accuracy

## Can it find recover AA seqs identified by prodigal for partial gene calls?

As a first test, I wanted to see whether the partial gene amino acid sequences identified by prodigal for the infant gut dataset could be recovered using anvi-gen-contigs-database, if the external gene calls did _not_ contain the `aa_sequence` column, so I ran the following workflow:

```bash
#! /usr/bin/env bash

rm -rf out
mkdir out

anvi-gen-contigs-database -f contigs.fa -o out/CONTIGS_0.db
anvi-export-gene-calls -c out/CONTIGS_0.db -o out/gene_calls_no_seq --gene-caller prodigal --skip

anvi-gen-contigs-database -f contigs.fa -o out/CONTIGS_1.db --external out/gene_calls_no_seq --ignore-internal --predict-frame

# Comparison betweeen these 2 gene calls
anvi-export-gene-calls -c out/CONTIGS_0.db -o out/gene_calls_original --gene-caller prodigal
anvi-export-gene-calls -c out/CONTIGS_1.db -o out/gene_calls_predicted_frame --gene-caller prodigal

# Create protein seq fastas
python create_fasta_from_gene_calls.py -g out/gene_calls_predicted_frame -o out/aa_seqs_predicted_frame.fa
python create_fasta_from_gene_calls.py -g out/gene_calls_original -o out/aa_seqs_original.fa

makeblastdb -dbtype prot -in out/aa_seqs_original.fa

# Each query sequence gets only 1 match--if perfect, `qseqid` should equate with `sseqid` for each hit
blastp -db out/aa_seqs_original.fa -query out/aa_seqs_predicted_frame.fa -max_target_seqs 1 -outfmt 6 -num_threads 40 > blast_results.txt
```

where `create_fasta_from_gene_calls.py` is:

```python
#! /usr/bin/env python

import pandas as pd
import argparse
import anvio.fastalib as u

ap = argparse.ArgumentParser()
ap.add_argument('-g', required=True, help="gene calls")
ap.add_argument('-o', required=True, help="output")
args = ap.parse_args()

output = u.FastaOutput(args.o)

gene_calls = pd.read_csv(args.g, sep='\t')
for idx, row in gene_calls.iterrows():
    output.write_id(row['gene_callers_id'])
    output.write_seq(row['aa_sequence'], split = False)

output.close()
```

Finally, we analyze the results:

```python
#! /usr/bin/env python

import pandas as pd

blast = pd.read_csv('blast_results.txt', sep='\t', header=None)
gene_calls = pd.read_csv('out/gene_calls_predicted_frame', sep='\t')
gene_calls_prodigal = pd.read_csv('out/gene_calls_original', sep='\t')

gene_calls = gene_calls[gene_calls['partial'] == 1]
gene_calls_prodigal = gene_calls_prodigal[gene_calls_prodigal['partial'] == 1]

blast = blast[blast[0].isin(gene_calls['gene_callers_id'])] # only consider partials
blast = blast[blast[0] == blast[1]].drop_duplicates(subset=(0, 1)) # did protein map to itself

fraction_correct = blast.shape[0] / gene_calls.shape[0]

inc = gene_calls[~gene_calls['gene_callers_id'].isin(blast[0])]
prod = gene_calls[~gene_calls['gene_callers_id'].isin(blast[0])]

print(f"Fraction of protein sequences predicted via markov model that blast to the protein sequences predicted by prodigal: {fraction_correct}")
print(f"Number agreeing with prodigal: {blast.shape[0]}")
print(f"Number disagreeing with prodigal: {gene_calls.shape[0] - blast.shape[0]}")
```

The output of the code is:

```
Fraction of protein sequences predicted via markov model that blast to the protein sequences predicted by prodigal: 0.99979883323275
Number agreeing with prodigal: 4970
Number disagreeing with prodigal: 1
```

If a protein blasts to its prodigal counterparts, this guarantees that the correct frame was chosen. Otherwise, the sequence would be frameshifted and contain no similarity in signal.

In conclusion, of the 4971 partial genes, only 1 did not work, and upon further inspect, even this one has matching frames (It has trouble blasting to itself because the sequence is `NGYFDDDSDSDSDSDSDSDSDSDSDSDSDSDSDSDSDSDSDSDSDS...`)


ekiefl:
# Parameters of effectiveness

I wrote a script to test what makes this thing accurate:

```python
#! /usr/bin/env python

import anvio

import numpy as np
import pandas as pd
import argparse
import anvio.fastalib as u
import anvio.constants as c
import anvio.utils as utils

ap = argparse.ArgumentParser()

ap.add_argument('-c', required=True, help="contigs")
ap.add_argument('-g', required=True, help="gene calls")
ap.add_argument('-m', required=True, help="transition matrix to use")
args = ap.parse_args()

matrix = np.load(args.m)
gene_calls = pd.read_csv(args.g, sep='\t')

random_offset = np.random.randint(0, 3, size=gene_calls.shape[0])
correct_frame = (3 - random_offset) % 3

gene_calls['start'] += random_offset

fasta_dict = {}
fasta = u.SequenceSource(args.c)
while next(fasta):
    fasta_dict[fasta.id] = fasta.seq

# -------------------------------------------------------------

report = {
    'prediction': [],
}

stop_probability = matrix.min()/1e3
null_probability = np.median(matrix)

print(f"min value: {matrix.min()}")

for i, row in gene_calls.iterrows():
    nt_seq = fasta_dict[row['contig']][row['start']:row['stop']]

    prediction, aa_seq = utils.get_most_likely_translation_frame(nt_seq, stop_prob=stop_probability, null_prob=null_probability, model=matrix)
    report['prediction'].append(prediction)

df = pd.DataFrame(report)

num_correct = (df['prediction'] == correct_frame).sum()
total = df.shape[0]
frac = num_correct / total

print(f"num_wrong: {total - num_correct}")
print(f"num_correct: {num_correct}")
print(f"frac_correct: {frac}")
```

## The probability value attributed to a stop codon

By far the most important classifying feature is the number of internal stop codons. These do not occur in the uniprot dataset so we need to _decide_ what they should be. Theoretically the probability is 0, however that leads to division complications, so we _must_ pick something.  Here are the values chosen:

| Prob(stop)            | Fraction Correct |
|-----------------------|------------------|
| median probability    | 0.9002           |
| min probability       | 0.9814           |
| min probability / 1e3 | 0.9920           |
| min probability / 1e6 | 0.9937           |

The results of this indicate that by far and large the biggest discriminant of the correct sequence from the incorrect sequences is the number of internal stop codons.

--------------------------------------------------

To determine whether the other factors have an effect at all, I set the stop codon transition probability to the median value, so to be the least informative. Then I varied the order of the markov model:

| Markov model order | Fraction Correct | .npy file size |
|--------------------|------------------|----------------|
| First order        | 0.8142           | 0 KB           |
| Second order       | 0.8637           | 1 KB           |
| Third order        | 0.8993           | 64 KB          |
| Fourth order       | 0.9175           | 1.3 MB         |
| Fifth order        | 0.9301           | 25 MB          |
| Sixth order        | 0.9378           | 489 MB         |

The results indicate that the composition of substrings is also a strong discriminant (since stop codons are normalized for).

Given file size, I have chosen the default to be the fourth order model, and the stop codon probability to be 1 millionth the minimum empirical probability

meren:
It is great to see it is working, @ekiefl! :) This is very exciting and of course, being able to recover 4,943 of 4,971 partial genes with such an elegant way without any heuristics is simply a major improvement.

Few comments:

- I think it would be best if there was a README file under `anvio/data/seq_transition_models/` that explains how to re-populate data in that directory from scratch. Something [like this](https://github.com/merenlab/anvio/blob/master/anvio/data/misc/SCGDOMAINCLASSIFIER.md), for instance. This would help anyone to maintain this directory with new genes, etc, etc.

- I think this calls for some usage examples [here](http://merenlab.org/software/anvio/help/programs/anvi-gen-contigs-database/) and perhaps a  small description [here](http://merenlab.org/software/anvio/help/artifacts/external-gene-calls/) citing this PR. We can then suggest people to read more in those pages under the ConfigError you raise in the case of `non_divisible_by_3_error` and `predict_frame == False`.

Most critically, though, I am curious to know what happens to ~30 proteins that don't match in your tests to amino acid sequences Prodigal predicted for partial genes? Without knowing which ones are incorrect, we are adding everything coming from the model to the DB, and now we have a bunch of genes that are not going to occur in any database known to humans. That is a risk I think we should be willing to take. But here are two quick points:

  * Are those genes particularly short? If they are, we may suggest a length cutoff and simply don't even try to translate them to avoid frameshift miscalls as much as possible (the rest of my comment will assume they are not too short for the sake of making another argument).
  * Which contigs they are coming from (if you send me the list I can tell you more about them)? The infant gut data includes eukaryotic organisms and viruses and I am curious whether partial genes we were unable to recover with this method are those that belong to non-bacterial origins.

But we may be wrongly assuming that the Markov model is doing worse than Prodigal. I wonder among those ~30 genes whether there are ones the amino acid sequences of which were predicted wrong by Prodigal. Which I believe is not unlikely since you are training on a much larger dataset of proteins (hopefully most of which are correct). If that happens (Prodigal screws up partial gene calls), this is big as it will have a big influence on metagenomic studies (when we study fragmented metagenomes, and then find proteins that are not ANYWHERE at all, some of this may be due to poor estimation of proper frames for partial amino acid sequences).

This opens a can of worms that I find worthy to investigate a bit deeper. To quantify whether the Markov model can better predict the frame in an impartial gene call compared to Prodigal in extreme cases, one can download a genome from the NCBI, use `anvi-script-process-genbank` to get a gold-standard gene calls file. Then fragment the genome with increasing number of contigs, generate contigs dbs and compare prodigal-predicted amino acid sequences for partial gene calls to those that are coming from the Markov model and compare them to the ground truth. I know this is a lot of work. We can describe this as a [small project](https://github.com/merenlab/anvio/issues?q=is%3Aissue+label%3A%22small+project%22+) and wait for a volunteer / interested party to show up (in fact I have a meeting with a potential intern, who can perhaps work with you on this).

In the absence of any insights into the efficacy of Prodigal vs Markov model in predicting AA seq for impartial genes, the most appropriate solution is to let people know that "_in some rare cases this approach can predict frames for impartial/inappropriate gene calls that may not agree with those that would have been predicted by Prodigal_".

All aside, this was a very good idea, came with a very good execution, and happened to work well :) Thank you!

ekiefl:
> I think it would be best if there was a README file under anvio/data/seq_transition_models/ that explains how to re-populate data in that directory from scratch. Something like this, for instance. This would help anyone to maintain this directory with new genes, etc, etc.

Done.

> I think this calls for some usage examples here and perhaps a small description here citing this PR. We can then suggest people to read more in those pages under the ConfigError you raise in the case of non_divisible_by_3_error and predict_frame == False.

Done, done, done.

ekiefl:
I made a log probability cutoff and tested the efficacy. Here are the results:

| Log prob cutoff | total | sequences predicted | correct | incorrect | % correct |
|-----------------|-------|---------------------|---------|-----------|-----------|
| 0               | 16453 | 16453               | 16340   | 113       | 99.3      |
| 1               | 16453 | 16403               | 16315   | 88        | 99.5      |
| 2               | 16453 | 16382               | 16305   | 77        | 99.5      |
| 3               | 16453 | 16360               | 16293   | 67        | 99.6      |
| 4               | 16453 | 16346               | 16287   | 59        | 99.6      |
| 5               | 16453 | 16335               | 16279   | 56        | 99.7      |
| 6               | 16453 | 16314               | 16264   | 50        | 99.7      |
| 7               | 16453 | 16295               | 16249   | 46        | 99.7      |

Based on this data, I think we should apply a log cutoff of 2. At that cutoff, half the sequences we get rid of were incorrect, which is higher than the 1/3 we would expect to cull by chance

Issue #1358, opened 2020-02-16
genes_in_contigs table should have a "type" column, which states whether the gene is coding or non-coding


ekiefl:
In the current design, it is very hacky/difficult to work with gene collections when you want codon sequences and values like 'codon_order_in_gene' and 'base_pos_in_codon', because noncoding genes do not have these attributes. This issue is magnified greatly when external gene calls are imported during anvi-gen-contigs-database

meren:
I fully agree. One solution is to literally add a new column to the genes table.

For instance, we can add a variable in constants, something like this:

```python
known_gene_types = set([1: 'CODING',
                        2: 'NONCODING',
                        3: 'UNKNOWN')
```

We would also require a new column in external gene calls file, `type`, that would have to have for each gene one of the known types. During import we can make sure everything checks out. If there is no column, we can set the type of each gene to `1` if we can get a translated DNA, and `2` if they have internal stop codons or `len(gene_seq) % 3 != 0` (of course when we do this, noncoding gene sequences that happened to satisfy the `% 3 != 0` criterion and lack 'internal stop codons' would erroneously be set as coding sequences, but it is still better than what's happening now).

Then we would change every function in anvi'o that reads gene calls from db to accept an optional `set` variable for gene type(s). So we can get all genes and work with only on those that match to a specific type (i.e., calculating SCVs only for coding genes) or request a specific subset of genes depending on the context (i.e., `1` for pangenomics). This way rRNA and tRNA HMMs could also add their genes with type `2` to avoid downstream confusions.

We have to do this for sure.

meren:
So the plan is to keep the `is_partial` flag, and implement `known_gene_types` as a new column.

ekiefl:
This is solved with the above-mentioned PR

PR #1429, opened 2020-05-18
A new feature for gene calls in contigs db: gene call types

meren:
This PR introduces the feature that was discussed in issue #1358.

I think the behavior makes much more sense. We now should be able to only work with genes that are type CODING in SCV and SAAV contexts. Thanks to Evan's Markov model we are golden with partial or non-partial external gene calls that come without an amino acid sequence. So we can rely on them in any analysis with confidence. The only problem I see is that when the user provides amino acid sequences in external genomes files, we are not testing whether the gene start/stop positions should be trimmed to match that amino acid sequence. This will probably explode at some point miserably because someone will put in start/stop positions that will not correspond to the amino acid sequence and/or the proper frame in DNA form.

Another thing this PR misses is a migration script for contigs db `v15` to `v16` :) Once it is in, we can merge it to `master` to start playing with it.


ekiefl:
Looks really comprehensive. Not too shabby for Sunday afternoon! I will add a migration script, and I will modify the profiler to make use of the new `coding_type` entry in gene_calls

meren:
Sounds great! Thank you.

Good catch with the snakemake. I can't think of anything else to add to the workflows.

ekiefl:
@meren I made the migration script as promised. Please take a close look to make sure I'm not doing anything too stupid. This is how I tested it:

Using master, I create a contigs database from infant gut dataset, then run HMMs to get Ribosomal genes. Then I run anvi-migrate and confirm that everything is `call_type = 1`, except for the Ribosomal genes at the bottom of the `genes_in_contigs` table, which are given `call_type = 2`.

If you think this branch has been tested enough, let's merge it. Otherwise, let me know if you have suggestions for further testing.

meren:
Thank you, @ekiefl. I will look into this after the class. I hope it is not too late.

ekiefl:
Upon further reflection, I realize the consequences of an error could be deadly, for example thinking about migrating matts assembly job

meren:
Hey @ekiefl. please see the changes I've made. I tried to make it more robust.

As far as I can see, it should work on every contigs db. I know that if these migration scripts fail, they usually fail so badly and ruin people's projects, so it takes a lot of courage can care to write them.

Perhaps we can store a copy of the db in tmp and then only delete that version after a successful migration. but if you are working with gigabytes of data, it is not practical either. A better way could have been to implement a `--safe-migration` or `--skip-safe-migration` flag to offer alternatives.

Perhaps a future weekend project.

ekiefl:
I am going to run the migration on a 13GB contigs DB. If it works, I will merge.

ekiefl:
Update.

Ok, I ran the migration on 2 large contigs DBs with no problems. But I discovered there is a new can of worms to open up: the `nt_position_info` compressed table. This stores the following info as an 8bit integer:

```python
        """This function compresses information regarding each nucleotide position in a given contig
           into a small int. Every nucleotide position is represented by four bits depending on whether
           they occur in a complete open reading frame, and which base they correspond to in a codon.

                0000
                ||||
                ||| \
                |||   Third codon?
                || \
                ||   Second codon?
                | \
                |   First codon?
                 \
                   Whether the position in a partial gene call

           8: int('1000', 2); nt position is in a partial gene call
           4: int('0100', 2); nt position is in a complete gene call, and is at the 1st position in the codon
           2: int('0010', 2); nt position is in a complete gene call, and is at the 2nd position in the codon
           1: int('0001', 2); nt position is in a complete gene call, and is at the 3rd position in the codon
        """
```

The way this compression works is, if the gene is partial, the codon positions are not logged. With these new changes we should be updating the rules of this compression to allow for codon positions to be logged even if the gene is partial. Instead, they should not be logged if the gene call type is NONCODING or UNKNOWN.

Then, there is a further complication because if they predict partial genes with prodigal, they are set as CODING but because we do not currently trim prodigal's partial gene calls, we are not able to say which codon position a nucleotide is in.

Here is what I propose to handle this. This may be a totally illegitimate path forward because I don't know the extent of contexts with which `nt_position_info` is used outside of `anvi-profile` (for example is this used in pangenomics?).

1. We  no longer store whether the gene is partial in `nt_position_info`. I reckon we don't need it?
2. We store `not_translated` in place of partial in `nt_position_info` which is 1 if `call_type` is either NONCODING or UNKNOWN
3. If prodigal gives us a partial gene prediction, we automatically trim it (this behavior would supersede the role of `--skip-predict-frame`.
4.  We do not attempt to propagate this behavior in the migration script. This will mean any genes that are partial will be flagged as `not_translated`. This is essentially how it was before.

What do you think @meren?

meren:
I'm glad you caught this, @ekiefl. The good news is that the only purpose of `nt_position_info` is to determine whether SNVs will be shown as green, red, or black in the interface. So that information is propagated by anvi-profile and used by anvi-interactive (and anvi-refine etc).

I think your proposal solves this and I agree with each item.

ekiefl:
With the latest commits, you can see that we properly store codon position info even when the gene is partial.

Before: 
image

After (look at far right):
image

meren:
pure beauty.

ekiefl:
Ok there were a lot of things that needed updating upon the switch from partial -> noncoding. See above commits for details.

I have done very basic tests on everything I have changed. Probably there exists errors somewhere in there and we will just have to discover those as they are reported.

I have to write a migration script for the profile database since now the variability table contains `is_in_noncoding_gene_call` and `is_in_coding_gene_call`. But before I write it, I want to point out that because the compressed nt info does not contain `is_in_partial_gene_call`, `is_in_complete_gene_call` info anymore, this info is no longer stored in the the nucleotide variability table. They have been replaced with `is_in_noncoding_gene_call` and `is_in_coding_gene_call`. In theory we can also keep these columns, but it will require some redesign. Do you think it is worth it @meren? Imo it is not.

meren:
> Ok there were a lot of things that needed updating upon the switch from partial -> noncoding. See above commits for details.

I can see multiple very important catches in the list. Thank you! Clearly we should increase the coverage of `run_all_tests` :/

> I have to write a migration script for the profile database since now the variability table contains is_in_noncoding_gene_call and is_in_coding_gene_call. But before I write it, I want to point out that because the compressed nt info does not contain is_in_partial_gene_call, is_in_complete_gene_call info anymore, this info is no longer stored in the the nucleotide variability table. They have been replaced with is_in_noncoding_gene_call and is_in_coding_gene_call. In theory we can also keep these columns, but it will require some redesign. Do you think it is worth it @meren? Imo it is not.

I agree. It is not worth it. We can get rid of them.

ekiefl:
@meren, I added a migration script, but I'm too scared to merge. Please do it if you feel good about it.

meren:
:)

Thank you, @ekiefl!

I will take a look at this tonight, and merge it once I am satisfied that enough eyes went though this.

meren:
(I couldn't because busy :( .. which continues to be the case)

meren:
OK. You take a look, and merge, @ekiefl. Everything looks good to me at this point.

Thank you very much!

anvio/migrations/profile/v32_to_v33.py
```diff
@@ -46,15 +46,9 @@ def migrate(db_path):
...
-        progress.update('Loading variability table...')
-        var = profile_db.get_table_as_dataframe(variable_nts_table_name, error_if_no_data=False)
-        var.rename(columns = {'in_partial_gene_call': 'in_noncoding_gene_call',
-                              'in_complete_gene_call': 'in_coding_gene_call'}, inplace=True)
-
-        progress.update('Rewriting variability table... Do not interrupt this process')
-        profile_db.drop_table(variable_nts_table_name)
-        profile_db.create_table(variable_nts_table_name, variable_nts_table_structure, variable_nts_table_types)
-        profile_db.insert_rows_from_dataframe(variable_nts_table_name, var)
+        progress.update('Renaming columns in variability table...')
```
ekiefl, line 46: omg... lol

ekiefl:
Ok. Looks good to me too!

PR #1437, opened 2020-06-08
Multithreaded prodigal

mooreryan:
Splits input fasta and runs Prodigal on each split in a different thread.

## Changes

* Add `AnviThread` class.  It's a subclass of `Thread`.  Only difference is that it tracks the return value of the of `Thread`'s `target` function/method.
* Add `CommandError` class.  A new `AnvioError` subclass that's a little more descriptive for commands that fail when run with something like `utils.run_command`.
* Add `--num-threads` command line option to `anvi-gen-contigs-database` script.
* Input fasta file is split into `num_threads`.  
  * Prodigal is run on each split it a new `Thread`.  
  * If the Prodigal command fails in a Thread it is retried a couple of times (currently hard-coded to 3 times).    I've found that sometimes running commands like this in a background thread, sometimes whatever bash command or command line program you're trying to run will fail for some weird reason.  Often the retries will "automagically" just fix it.
  * All the resulting peptide and gff gene files are concatenated into a single peptide file and a single gff file.
  * One potential issue with the gff files...prodigal starts the gene numbering in them at 1 for each instance.  So the numbering in the gff files will be different from the numbering in the peptide files. However, all info about genes is currently gotten from the peptide files so not sure it will be a problem....if the GFF files are used somewhere else down the line, it may be.

## Tests

* The `AnviThread` has a couple of unit tests.
* What kind of (automated) tests would y'all like to see for the Prodigal driver?  (Unit tests for Prodigal driver code?  Some changes in existing integration tests?)

## Profiling

Here's the profile info with 1 and 2 threads running the Infant Gut dataset.  One thing to note is that even when the user asks for a single thread, it is still run in "multithreaded" mode...i.e., it will still "split" the file and run Prodigal on that single split in another thread.  It seems like a wasteful thing, but at least on the Infant Gut dataset, it didn't add a noticeable overhead compared to the original code.  However, it may depending on the data.

### Single thread

profile-out-threads-1

### Two threads

profile-out-threads-2


meren:
Thank you, @mooreryan!

I can't wait to look into this :)

anvio/drivers/prodigal.py
```diff
@@ -21,11 +26,108 @@
...
+
+        # Return code of zero means success in the unix world.
+        if ret_val == 0:
+            return ret_val
+
+    # Failed to run the command a couple of times, so return an error.
+    return CommandError(f'Failed to run the command after {retries} tries. Final exit code was: {ret_val}')
+
+
+def split_fasta_wrapper(fasta_file_path, num_splits):
```
meren, line 62: I think it would be much better to use the `utils.split_fasta` directly here.

If you think checking whether splits exist is essential, you can change the header of the existing function from,

```python
def split_fasta(input_file_path, parts=1, prefix=None):
```

to

``` python
def split_fasta(input_file_path, parts=1, prefix=None, check_splits=False):
    (...)

    if check_splits:
        (check  splits here)

    (...)
```

to check for splits :) But I am not sure if it is really necessary. by the way, this will probably explode if the FASTA file contains number of sequences that are smaller than the number of threads. please test for that to make sure it is OK :)

anvio/drivers/prodigal.py
```diff
@@ -21,11 +26,108 @@
...
+    return paths
+
+
+def add_suffix(strings, suffix, separator='.'):
+    """Adds a suffix on to a list of `strings`.  Joins with `separator`."""
+    return [separator.join([string, suffix]) for string in strings]
+
+
+# Todo: Consider using shutil.copyfileobj instead
+def concatenate_files(intermediate_files, final_file):
```
meren, line 82: I think you should consider using `utils.concatenate_files` instead of this. the less the redundancy in the codebase the better for maintenance.

anvio/drivers/prodigal.py
```diff
@@ -21,11 +26,108 @@
...
+                    # Read line by line in case the logfile is large.
+                    for line in infile:
+                        outfile.write(line)
+            else:
+                raise CommandError('Something went wrong in the command, and we could not find the expected '
+                                   f'logfile: {file_path}')
+
+
+def run_commands_in_threads(commands, target, final_logfile, num_retries):
+    """Runs the given `commands` each in its own thread.
```
meren, line 100: Can't we turn this into another class in that simply takes in all this information and put it in `anvio/anvi_thread.py` so it is generalized in such a fashion, anyone can use this?

anvio/anvi_thread.py
```diff
@@ -0,0 +1,47 @@
+# coding: utf-8
+"""A lightly modified Thread sublcass.  Tracks the return values of Thread's target."""
+
```
meren, line 3: I think it would be much better to call this `anvio/threadingops.py` so we can expand it with more classes and functions :)

anvio/drivers/prodigal.py
```diff
@@ -21,11 +26,108 @@
 __maintainer__ = "A. Murat Eren"
 __email__ = "a.murat.eren@gmail.com"
 
-
 run = terminal.Run()
 progress = terminal.Progress()
 pp = terminal.pretty_print
 
+# How many times to retry the gene calling.
+NUM_RETRIES = 3
```
meren, line 34: I am not sure why is this necessary :) In the function where this variable is used the description says "Sometimes commands can fail for weird reasons so retrying often fixes things", but it should never happen. If retrying is solving something, there must be somewhere else where there is a problem (i.e., race conditions, missing file.close() calls, etc) and we should find them. I think this variable complicates things unnecessarily.

mooreryan: I think that's a fair point....I will remove the "retrying" code from the command runner.

anvio/drivers/prodigal.py
mooreryan: Hmm, I wouldn't say checking the splits is essential...I suspect that an exception would be raised in the `split_fasta` function if there were a problem writing the sequences and closing the file.  It was more of sprinkling in assertions to catch things I'm assuming to be true.  However, there are also many checks that files exist later, so it could definitely be considered redundant!

Long winded way to say that I'll switch it to not use the wrapper.

anvio/drivers/prodigal.py
mooreryan: Ah yep, that would be better...for some reason I did not check to see if that function already existed in the code base!

anvio/drivers/prodigal.py
mooreryan: Yep, that sounds like a good idea!  While looking at the hmmer driver, it also uses threading in a similar way to the prodigal driver (as i based it on that!), so really, it may be worth it at some point to abstract the whole "split up a fasta and run some program on each split then concatenate" pattern.

anvio/drivers/prodigal.py
mooreryan: > this will probably explode if the FASTA file contains number of sequences that are smaller than the number of threads. please test for that to make sure it is OK :)

Looking at the `split_fasta` code, it looks like it handles this case ...see: https://github.com/merenlab/anvio/blob/849f998c8be3b4eabebefb485f0bd0d79e67e6a0/anvio/utils.py#L748

This looks like a good candidate for some unit tests though...I can add some to be sure it's working that way.

meren: Thank you! :)

mooreryan:
Many things have been updated in this latest round of commits.

The main thing is that I took a stab at making an interface which abstracts the whole notion of split a file, run some commands on it, collate the results pipeline.

Check out the `ThreadedCommandRunner` abstract class for the interface, and `ThreadedProdigalRunner` for a concrete subclass.

I tried to leave tons of comments and lots of docstring help messages to make it clear how somebody could implement a subclass `ThreadedCommandRunner`.

Let me know what you think of the design of it!

Thanks!!

mooreryan:
Oh, and I forgot to mention, the new classes still need unit tests, but I wanted to wait to get some feedback on the interface before I went through the process of writing a bunch of tests for it.  It does work on the infant gut test dataset at least though.

meren:
> @meren I think the changes in commit bfdc8ba may have been pushed to the wrong branch....

I realized it too late, but I `cherry-pick`ed it to `master` :) Sorry!

anvio/threadingops.py
```diff
@@ -0,0 +1,551 @@
...
+        finally:
+            # Avoid a refcycle if the thread is running a function with
+            # an argument that has a member that points to the thread.
+            del self._target, self._args, self._kwargs
+
+
+# todo fix all the self.logger checks for None
+
+
+class ThreadedCommandRunner(abc.ABC):
```
meren, line 98: I haven't tested it yet, but this is a beautiful class! I really like the overall approach :)

The docstring is very clear, and the `ThreadedProdigalRunner` looks tidy, which is great!

anvio/drivers/prodigal.py
```diff
@@ -94,78 +99,56 @@ def check_version(self):
...
-
-        utils.run_command(cmd_line, log_file_path)
-
-        if not os.path.exists(self.amino_acid_sequences_in_contigs):
-            self.progress.end()
-            raise ConfigError("Something went wrong with prodigal, and it failed to generate the "
-                              "expected output :/ Fortunately, this log file should tell you what "
-                              "might be the problem: '%s'. Please do not forget to include this "
-                              "file if you were to ask for help." % log_file_path)
+        self.progress.update('Identifying ORFs using %s' % (('%d threads.' % self.num_threads) if self.num_threads > 1 else 'a single thread (boo).'))
```
meren, line 130: I wonder if there is an easy way to show the number of active threads. Because some threads will be done with their files quicker than others.

anvio/threadingops.py
```diff
@@ -0,0 +1,551 @@
...
+
+        return gene_calls_dict, amino_acid_sequences_dict
+
+    # Check the gff files and append them into the single outfile.
+    def _process_gff_files(self, gff_paths: Collection[str]) -> None:
+        """Check the gff files and combine them.
+
+        Question:  Is genes_in_contigs (the gff file) ever actually used?...each of the threads will start renumbering
+        sequences from 1 in each subfile.  Will this break something downstream?
+        """
```
meren, line 547: No, GFF files are not used for anything (if we did, numbering that reset in each file would have been a big problem though :)).

anvio/threadingops.py
```diff
@@ -0,0 +1,551 @@
...
+        Raises `ValueError` if `self.number_of_splits < 0.
+
+        See Superclass `_split_input_file` for more info.
+        """
+        # Todo: probably should move this check into the constructor.
+        if self.number_of_splits <= 0:
+            ValueError(f'number_of_splits muts be > 0.  Got {self.number_of_splits}')
+
+        # Todo are there errors to catch here?
+        self.input_file_splits = utils.split_fasta(self.input_file_path, parts=self.number_of_splits)
```
meren, line 409: I think `utils.split_fasta` needs a `randomize=False` flag and when it is set to `True` it should split the input FASTA file in such a way, resulting split files do not have the sequences in the same order.

What I mean is this: Currently when we split a FASTA file into two splits, the first split contains the first half of contigs, and the second split contains the second half. Assemblers often report contigs in an ordered fashion, where longest contigs appear at the beginning of the file. So if you are splitting an assembly output into 5 files, the first file may contain more nucleotides than the last four combined. Which kinda makes the multithreading a little less effective.

I realized this while testing this code on the infant gut dataset. I gave 6 threads, and 5 of them finished very quickly and there was only a single thread for more than 70% of the total runtime.

ekiefl:
Damn, this `threadingops` module is looking awesome, @mooreryan. Good job

anvio/threadingops.py
mooreryan: Check out https://github.com/merenlab/anvio/pull/1448 for one solution to this.

meren:
When should we merge this, @mooreryan? :)

anvio/drivers/prodigal.py
mooreryan: The number of active threads is available in `threading.enumerate()`.  This will include the calling thread and any other threads that may have been spawned.  First thought I had was just to check for `AnviThread`s as that is what this class will spawn.  But it's conceivable that something else could spawn an `AnviThread`.  Then I thought, just give the threading ops threads specific names and search against that.  Meh, kind of annoying to do that.

Then I thought of something like this:

```python
        living_worker_threads = [thread for thread in self.threads if thread.is_alive()]
        while len(living_worker_threads) > 0:
            if self.logger:
                self.logger.run.info('Info', f'living_worker_threads: {len(living_worker_threads)}')

            time.sleep(1)
            living_worker_threads = [thread for thread in self.threads if thread.is_alive()]
```

This would basically be a hacky version of calling `join()` on all threads in `self.threads` a `for` loop, except that it will print out info about how many threads are currently running.  On the other hand maybe it's kind of annoying to print a message like that every second...what do you think?

(Of course the `[thread for thread in self.threads if thread.is_alive()]` bit could become a class method so it's easier to call.)

anvio/threadingops.py
```diff
@@ -0,0 +1,551 @@
...
+from typing import Callable
+from typing import Collection
+from typing import Dict
+from typing import Mapping
+from typing import MutableMapping
+from typing import MutableSequence
+from typing import Optional
+from typing import Sequence
+from typing import Tuple
+from typing import Union
```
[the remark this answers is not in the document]
mooreryan: I can remove the typing.  As a newish python programmer the typing helped catch a few bugs while I was working on it, but removing the type annotations at this stage doesn't bother me at all.

anvio/threadingops.py
```diff
@@ -0,0 +1,551 @@
...
+                 input_file_path: str,
+                 collated_output_file_paths: Mapping[str, str],
+                 number_of_splits: int,
+                 log_file_path: str,
+                 logger: Optional[Logger],
+                 installed_version: str,
+                 parser: Callable[[str], Dict[str, Any]],
+                 # while it will take str or int, it will be saved as a str
+                 translation_table: Optional[Union[str, int]]
+                 ) -> None:
```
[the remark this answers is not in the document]
mooreryan: Did you want the `ThreadedCommandRunner` to also take `args`, or just the `ThreadedProdigalRunner`?

mooreryan:
Oh shoot...I really messed up the version control on this branch......I'm going to close this pull request and the git issues.

meren:
Ouch :) Sorry, @mooreryan. I hate it when it happens. I hope resolving git issues goes smoothly.

PR #1441, opened 2020-06-13
add anvi-run-kegg-kofams to contigs workflow

ivagljiva:
As the title implies, this PR now allows us to run `anvi-run-kegg-kofams` on multiple contigs databases using the contigs workflow :)


meren:
looks great. thank you, Iva!

Issue #1438, opened 2020-06-09
anvi-merge is slow/hangs for large profile databases


ekiefl:
Trying to merge ~300 profile databases that are collectively 1TB seems too challenging for `anvi-merge`. `Merging split coverage data` has been stuck on one sample for over 24 hours. The job has not exceeded the allotted memory, and is still "running".

Presumably, it is due to some non-ideal operations that do not scale well to this data size. I will rerun the job to try and make sure it is not due to issues unrelated to anvi'o, like the job has been cancelled but still remains in the queue, etc.

ekiefl:
I noticed a potential problem here:

```python
        # fill coverages in from all samples
        for input_profile_db_path in self.profile_dbs_info_dict:
            self.progress.update(input_profile_db_path)
            sample_split_coverage_values = auxiliarydataops.AuxiliaryDataForSplitCoverages(AUX(input_profile_db_path), self.contigs_db_hash)

            for split_name in self.split_names:
                coverages_dict = sample_split_coverage_values.get(split_name)
                for sample_name in coverages_dict:
                    merged_split_coverage_values.append(split_name, sample_name, coverages_dict[sample_name])


            sample_split_coverage_values.close()
        merged_split_coverage_values.store()
        merged_split_coverage_values.close()
```

`merged_split_coverage_values` has a buffer which is only stored at the very end, after all split coverage data has been processed. By moving the `store` inside the loop, I hoped this would fix things. However, here are the timings for each sample..... It still grows exponentially:

![image](https://user-images.githubusercontent.com/8688665/84336713-787d3080-ab5d-11ea-9d42-d39f4c8a690b.png)

Each time is the time it takes to process a sample.

ekiefl:
^ The above fix is an improvement in memory management, but the speed is still slow. I've determined what grows exponentially is

```python
merged_split_coverage_values.store()
```

Down the rabbit hole we end at

```python
233     def _exec_many(self, sql_query, values):$
  1         chunk_counter = 0$
  2         for chunk in get_list_in_chunks(values):$
  3             if anvio.DEBUG:$
  4                 self.progress.reset()$
  5                 self.run.info_single("Adding the chunk %d with %d entries of %d total is being added to the db with "$
  6                                      "the SQL command '%s'." \$
  7                                     % (chunk_counter, len(chunk), len(values), sql_query), nl_before=1)$
  8 $
  9             self.cursor.executemany(sql_query, chunk)$
 10 $
[...]

PR #1442, opened 2020-06-14
Merge fixes

ekiefl:
This PR addresses https://github.com/merenlab/anvio/issues/1438

# Changes

The main fixes are as follows:

1. Concatenated split coverage data to be stored in the merged DB is written and emptied from memory after processing each sample instead of being held in memory until all samples have been processed. This saves memory but does not speed up the code.

2. `CREATE INDEX` used to be called during creation of `split_coverages` table. Herein lies the problem. `INSERT` operations are very expensive on indexed tables. Yet indexed tables are important for fast table accession. The solution proposed in this PR moves index creation to `auxiliarydataops.AuxiliaryDataForSplit.close` to be added as a teardown operation. See the docstring for more info.

3. `merger.MultipleRuns.merge_split_coverage_data` was using the `split_coverages` API to fetch each split's coverage numpy array by converting from binary format, and then converting it back to binary format for storage in the merged DB. Instead, the contents of each `split_coverages` table are just appended to the merged `split_coverages` table using `db.DB.copy_paste`. No conversion to and from binary formats.

# Results

- The memory saved from (1) is unmeasured.

- (2) represents the biggest time save. Before, the table had to be reindexed after each sample's coverage data was appended. This created an asymptotic-like time complexity, which is why I couldn't merge 1TB of profile DBs. Now, `INSERT` operations are independent of table size, so the time complexity is linear with respect to total data size.

- (3) For my dataset (500k contigs), about 2 minutes is spent converting to and from binary formats per sample. Therefore about 12 hours is saved over 400 samples.

# Testing

Split coverages appear normal when viewed through the inspection page, which gives me confidence that shifting to the `copy_paste` method has not influenced anything.

I am currently running the aforementioned merge job on this dataset.

What else should we test?


anvio/db.py
```diff
@@ -496,6 +496,9 @@ def get_table_as_dict(self, table_name, table_structure=None, string_the_key=Fal
         # new ones to avoid key information to not be overwritten due to the lack of unique entry ids which become keys for
         # the data dictionary). in other cases there are no ways to fix it, such as for HMM tables.. The ACTUAL SOLUTION to\
         # this is to remove `entry_id` columns from every table in anvi'o, and using SQLite indexes as entry ids.
+        #
+        # NOTE from the future
+        # Every SQLite table has an implicit column called ROWID. Does this solve our problem?
```
meren, line 501: yes, but unfortunately I learned about it a bit too late :/

meren: it would have made things MUCH more simpler if I had used this instead of `entry_id` monstrosity..

anvio/merger.py
```diff
@@ -454,11 +381,36 @@ def _concatenate_single_profile_tables(self, merged_db, table_name, is_auxiliary
...
 
+        if 'entry_id' in merged_db.get_table_structure(table_name):
+            # entry_ids need to be made unique
+            self.progress.new("Merging '%s' tables" % table_name)
+            self.progress.update("Updating entry_ids")
+            merged_db.reset_entry_id(table_name)
+            self.progress.end()
+
+
+    def merge_variant_tables(self, table_name):
```
meren, line 397: beautiful solution :)

anvio/db.py
```diff
@@ -180,6 +180,21 @@ def copy_paste(self, table_name, source_db_path, append=False):
         self._exec('''DETACH DATABASE "source_db"''')
 
 
+    def reset_entry_id(self, table_name):
```
meren, line 183: this will be a life saver.

meren:
These all look very good, @ekiefl! Thank you very much.

We must mention this in `v7` release notes :)

ekiefl:
# More changes

- Instead of loading data frame SQL into python, then writing back into python, now all split coverages, SNVs, SCVs, and indels are concatenated from within SQL for big speed gains (meren you've already reviewed all these changes)

# Testing

Unfortunate news... Things are fast, but since SNVs, SCVs, and indels have `entry_id`s, you cannot simply concatenate and be done. You have to overwrite `entry_id` with a new set of sequential integers. The solution I used is this:

```python
    def reset_entry_id(self, table_name):
        """Rewrite entry_id column as sequential integers. Complain if entry_id doesn't exist

        Notes
        =====
        - Every SQLite table contains an implicit column, ROWID, that is the row number of the
          entry. We take advantage of that here.
        """

        if 'entry_id' not in self.get_table_structure(table_name):
            raise ConfigError("DB.reset_entry_id :: %s does not have entry_id as a column" % table_name)

        self._exec('''UPDATE %s SET entry_id = ROWID-1''' % table_name)
```

This `ROWID` is actually a hidden column present in all SQL databases. But the problem is that this is an indetermined level of slow. Creating the merged SNV table took 2 hours, and it's been stuck on this step for 10 hours.

So it seems `entry_id` is coming back to haunt us yet again. @meren, I fear it is time to approach "one of the most critical design mistakes that remain in anvi'o". How difficult do you anticipate this would be?

meren:
I think the solution is to get rid of `entry_id` columns and change functions in `db.py` module to accept a parameter, i.e., `use_row_ids_as_entry_ids=False` to report every dict currently expecting `entry_id`s as keys to be reported with them.

I don't think anywhere in the code we are dependent on the actual values in any of the `entry_id` columns. the entire purpose of that is to have a unique element in tables so when we are adding contents to a dictionary entries in the table do not overwrite each other.

This would clean up A LOT OF mess and simplify data insertion and recovery methods we have.

meren:
> So it seems entry_id is coming back to haunt us yet again.

Entry ids are no more following #1443. I hope this improves things. When you have an idea please share an overall final performance gain numbers for `anvi-merge`, @ekiefl.

ekiefl:
With a collective size of 1TB, 364 single profiles were merged in about 20 hours. This job used to be impossible, so the performance gain in speed is infinite.

PR #1444, opened 2020-06-19
Fix modules db bug

ivagljiva:
Briefly, this PR fixes a bug in the MODULES.db setup done in `anvi-setup-kegg-kofams`. Prior to this fix, some KOs that are components of enzyme complexes and on the same ORTHOLOGY line in a modules file were not separated when parsing the line. These KOs ended up in the same ORTHOLOGY row in the MODULES.db:
image

And since they were multiple KOs in the same ORTHOLOGY line, these KOs could not be found when searched for individually. 

After this fix, the KOs are separated so that they each get their own individual ORTHOLOGY line in the DB:
image
So now searching works for these individual KOs, they get marked as present in their respective modules, and the module completion estimates are now correct.

### DIFF
--- a/anvio/kegg.py
+++ b/anvio/kegg.py
@@ -2961,7 +2961,8 @@ class KeggModulesDatabase(KeggContext):
         data_types_to_split = ["ORTHOLOGY","REACTION"] # lines that fall under these categories need to have data_vals split on comma
         if current_data_name in data_types_to_split:
             # here we should NOT split on any commas within parentheses
-            for val in re.split(',(?!.*\))', data_vals):
+            vals = [x for x in re.split('\(|\)|,|\+|-', data_vals) if x]
+            for val in vals:
                 line_entries.append((current_data_name, val, data_def, line_num))
         else:
             line_entries.append((current_data_name, data_vals, data_def, line_num))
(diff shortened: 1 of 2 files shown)


ivagljiva:
@meren As this is a simple yet critical fix, I will merge this now so that we can start re-running metabolism estimation to get correct output in various projects. The reason I made a PR just for this one commit is so that it is well-documented, as there will likely be a big difference in estimation outputs before and after this fix.

If you look at this later and think it was not a good idea, we can revert it :)

meren:
Thanks, Iva!

ekiefl:
Thanks for documenting this. I will rerun on my dbs

ivagljiva:
@ekiefl I have just finished updating the KEGG data directory on Midway, so you can rerun KOfams/estimation whenever you are ready :)

UriaMorP:
Right on time! Thanks @ivagljiva

UriaMorP:
Hey @ivagljiva since your last commit, I'm struggling to get the `gene_callers_id` for KOfams.
When I use `anvi-export-functions`, I'm getting this output:

image

Where previously, the table contained a `gene_callers_id` column:
image

I'm using this feature in order to construct a non-redundant gene catalog out of the MAGs I have containing only sequences that hit some KOfam. That is, I'm iterating through the fasta file used to create the contigs.db and only take entries that hit a KO.

Can you think of a way to do that?

meren:
I don't see how this could be related to anything Iva did. It was probably me :( I will look into this immediately and thank you for catching it!

UriaMorP:
Thanks, I'll let you know as soon as I get to check it!

Issue #1431, opened 2020-05-26
anvi-gen-contigs-database is slow


ekiefl:
Hi!

[`anvi-gen-contigs-database`](http://merenlab.org/software/anvio/help/programs/anvi-gen-contigs-database/), one of the key programs in [anvi'o](http://merenlab.org/software/anvio), takes a lot of time running, but its speed can be increased greatly with some of the following suggestions (ordered based on potential gain):

### 1. Refactor `get_kmer_frequency`.

This function currently spends 76% of `anvi-gen-contigs-database`'s time.

![image](https://user-images.githubusercontent.com/8688665/82952628-f2fb5d00-9f6e-11ea-98d0-0b77e9c4e5c4.png)

One issue is that it appears to be called twice. Finding out more about the call structure is important. If second call is unnecessary, we can see a 2-fold speed gain. More importantly, the code can be re-written using `numba` (see [here and below](https://github.com/merenlab/anvio/blob/master/anvio/bamops.py#L1153) in `bamops.py` for `numba`-written functions) or in Cython. I estimate these speed gains would be 40-200x faster, making [`get_kmer_frequency`](https://github.com/merenlab/anvio/blob/master/anvio/kmers.py#L56), which currently has a very naive Python implementation, almost instant.

### 2. Parallelize "The South Loop"

["The South Loop" in `dbops.py`](https://github.com/merenlab/anvio/blob/master/anvio/dbops.py#L3890) takes up most of the time in `anvi-gen-contigs-database` and processes contigs independently, and therefore could be parallelized for nearly linear speed gains as a function of number of threads.

### 3. Parallelize Prodigal

Prodigal is a gene caller that does not support multithreading. But one can implement a workaround to split input FASTA files into multiple pieces and running Prodigal in parallel just to combine results later. This has been suggested before (https://github.com/merenlab/anvio/issues/1344), although it seems like the most amount of work for the least amount of gain (but you're the boss).

---

If you are interested in addressing any of these suggestions, you can [install anvi'o to follow the active codebase](http://merenlab.org/2016/06/26/installation-v2/#following-the-active-codebase-youre-a-wizard-arry) and start playing with some FASTA files using this command:

```
anvi-gen-contigs-database -f FASTSA.fa -o CONTIGS.db
```

Let us know here if you have any questions.

Thanks!

rbartelme:
[...]

PR #1445, opened 2020-06-19
Speed up get_split_start_and_stop_with_gene_calls

mooreryan:
References  #1431.  See also [this comment](https://github.com/merenlab/anvio/issues/1431#issuecomment-646344564).

# Overview

Rather than starting with a `set` of all positions in a contig and then going through all coding regions and removing them by converting `range` to `set` then using set `difference` function, start with an empty `list` of coding regions.  Then `append` all coding regions to that list.  Finally, subtract all those coding positions from the set of all positions in the contig to create a set of all non-coding positions.

# Timing
Here is the original timing:

original_get_split_start_and_stop_profile

And here is the timing after the change:

fast_get_split_start_and_stop_profile

It's a simple enough change, but it makes a *big* difference in runtime.

# Testing

I used this little bash script to test the splits in the contigs db from the original master branch, and the updated fast function.

```bash
#!/bin/bash

db1="${1}"
db2="${2}"

diff -y <(sqlite3 "${db1}" 'select split, start, end from splits_basic_info;' -separator $'\t' -header) \
        <(sqlite3 "${db2}" 'select split, start, end from splits_basic_info;' -separator $'\t' -header)
```

Splits are exactly the same.

Also the `run_all_tests.sh` script ran without errors.

### DIFF
--- a/anvio/utils.py
+++ b/anvio/utils.py
@@ -21,6 +21,7 @@ import tracemalloc
 import configparser
 import urllib.request, urllib.error, urllib.parse
 
+import itertools as it
 import numpy as np
 import pandas as pd
 import Bio.PDB as PDB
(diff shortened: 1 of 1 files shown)


meren:
Ah, awesome :)

> It's a simple enough change, but it makes a big difference in runtime.

That is the definition of brilliance.

meren:
Looks great, Ryan. Thank you *very* much! :)

PR #1448, opened 2020-06-21
Add shuffle option to utils.split_fasta()

mooreryan:
See this comment for why this is necessary:  https://github.com/merenlab/anvio/pull/1437#discussion_r440516194.

Basically, you would like to be able to split up a fasta file in such a way that in cases where sequences are sorted by length, not all the super long sequences are all clumped together in one of the splits.

To do this, I've added a `shuffle` option to `utils.split_fasta`.  When set to `True` it will sort of do the opposite of interleaving the sequences (i.e., the first sequence goes to the first outfile, the second to the second, and so on).

Also includes unit tests for the `split_fasta` function. 

Oh and a couple of places where fasta files weren't closed after reading or writing if there was an error were also fixed.

### DIFF
--- a/anvio/fastalib.py
+++ b/anvio/fastalib.py
@@ -100,6 +100,7 @@ class SequenceSource():
             self.file_pointer = io.open(self.fasta_file_path, 'rU', newline='')
 
         if not self.file_pointer.read(1) == '>':
+            self.file_pointer.close()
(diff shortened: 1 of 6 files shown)


meren:
Yay :) This looks great to me :) I think this will maximize the performance gain!

meren:
Thank you, Ryan (also for the unittests)! Merged to master.

PR #1449, opened 2020-06-23
Migrate modules db

ivagljiva:
This PR contains the very first migration script for a modules DB (which was rendered necessary by my last PR #1444 ). What the migration script does is fetch all the ORTHOLOGY entries that have multiple KOs, removes those entries, splits each multi-KO entry into entries with individual KOs, and adds those individual entries back into the modules table.

This PR also officially adds the 'version' metadata to the modules DB self table, and as an added bonus, there is a small bug fix in `anvi-estimate-metabolism` where before we were double-counting some modules as complete (we don't do that anymore).

## How to test
To test the migration script, this is what I did:
1) Set up an early version of the KEGG modules DB from an archive file (this particular archive file will be uploaded to figshare soon; if you want it before that happens I can send it to you)
```
anvi-setup-kegg-kofams --kegg-archive KEGG_build_2020-04-27.tar.gz --kegg-data-dir KEGG_ARCHIVE
```
2) Run the migration script 
```
anvi-migrate ./KEGG_ARCHIVE/MODULES.db --migrate-dbs-safely
```

Before the migration, this is what the self table looks like:
Screen Shot 2020-06-23 at 00 45 05
And here is how it looks like when you manually search for multi-KO entries:
Screen Shot 2020-06-23 at 00 45 30

After the migration, the self table looks like:
Screen Shot 2020-06-23 at 00 46 25
And we can no longer find any multiple entries:
Screen Shot 2020-06-23 at 00 46 16
Plus, we can confirm that the KOs were split properly by looking at some examples from module M00567:
image

### DIFF
--- a/anvio/migrations/modules/v1_to_v2.py
+++ b/anvio/migrations/modules/v1_to_v2.py
@@ -6,7 +6,6 @@ import argparse
 import re
 
 import anvio.db as db
-import anvio.tables as t
 import anvio.utils as utils
 import anvio.terminal as terminal
 
--- a/anvio/migrations/modules/v1_to_v2.py
+++ b/anvio/migrations/modules/v1_to_v2.py
@@ -3,6 +3,7 @@
 
 import sys
 import argparse
+import re
 
 import anvio.db as db
 import anvio.tables as t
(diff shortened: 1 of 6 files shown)


anvio/tables/__init__.py
```diff
@@ -30,7 +30,8 @@
                          'pan': pan_db_version,
                          'genomestorage': genomes_storage_vesion,
                          'auxiliary data for coverages': auxiliary_data_version,
-                         'config': workflow_config_version}
+                         'config': workflow_config_version,
+                         'modules': kegg_modules_db_version}
```
meren, line 34: I think this should be called `metabolic_modules_db_version`. if we switch to MetaCyc one day, or implement something that can keep both KEGG and MetaCyc, we don't have to change an essential db name :)

ivagljiva:
You can find the archive file used to test this migration script [here](https://figshare.com/articles/KEGG_build_2020-04-27_tar_gz/12545975).

anvio/migrations/modules/v1_to_v2.py
```diff
@@ -0,0 +1,75 @@
...
+        if orth != "ORTHOLOGY":
+            raise ConfigError("Serious SQL error here. We asked for ORTHOLOGY but we got %s" % orth)
+        split_kos = re.split('\+|\-', kos)
+        for ko in split_kos:
+            if len(ko) != 6 or ko[0] != 'K':
+                raise ConfigError("Uh oh. We split a KO that doesn't seem to have the right format. It looks like this: %s, "
+                                  "and the string that it came from is %s." % (ko, kos))
+            good_rows.append((mod, orth, ko, definition, line))
+
+    modules_db.insert_many(t.module_table_name, entries=good_rows)
```
meren, line 51: Ah, sorry! I've missed this totally.

Migration scripts should never use table descriptions in the tables module. The reason to this is this: imagine someone run this script a year from now, and `t.module_table_name` in `anvio.tables` is different in that anvi'o version than what it is now. This will crash.

You should always use explicit names, and ideally never import `anvio.tables`.

meren:
Sorry, I just realized I missed something critical :)

anvio/migrations/modules/v1_to_v2.py
ivagljiva: No worries! I will fix this :)

ivagljiva:
@meren Does it look good now?

anvio/migrations/modules/v1_to_v2.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+
+import sys
+import argparse
+import re
+
+import anvio.db as db
+import anvio.tables as t
```
meren, line 9: I guess this is no longer necessary :)

meren:
Looks good @ivagljiva! Thank you :)

ivagljiva:
Thanks @meren! I will merge now. 🚀

Issue #1452, opened 2020-06-26
Contigs DB names with "." creates error in contigs snakemake workflow


mschecht:
Hi @meren and @ekiefl,

Is it possible to have contigs DB names that contain periods?

I am currently working with a set of genomes that I would like to convert to contigs DBs but the genome names contain periods:

e.g. file name: RS_GCF_xxxxxxxx.1_genomic.fna.gz; contigsDB name: RS_GCF_xxxxxxxx.1

Unfortunately anvio complains:

```
Config Error: Well, the fasta.txt entry name contains characters that anvi'o does not like :/
              Please limit the characters to ASCII letters, digits, and the underscore ('_')
              character.
```

If it is not possible for anvio to work with names with non-ASCII characters, I can make a lookup index and convert the genome names to an anvio friendly format.

Thanks for the help!

ekiefl:
I don't know if this feature request is likely to be fulfilled @mschecht. The codebase very likely makes assumptions all throughout the codebase that there contains no `.` characters. meren would know better.

By the way, `.` is an ASCII character, but not an ASCII letter (a-zA-Z)

mschecht:
Thanks @ekiefl, maybe anvio could internally store a lookup index? For example, there could be a script called, `anvi-make-names-anvio-friendly` that would store a separate table in the sqlite db that has the lookup to the original names and the corresponding names that anvio can work with.

ekiefl:
In fact, that's exactly what `anvi-script-reformat-fasta` does if `--simplify-deflines` is provided. The results are not stored in the DB, but output as a file. Something generalized to snakemake inputs could be useful, but I'm not sure this is a worthwhile project. I think it would be nearly equal effort to allow anvi'o to work with strange characters.

meren:
We have a section in contigs database `self` table to describe the project name:

```
(...)
  -n PROJECT_NAME, --project-name PROJECT_NAME
                        Name of the project. Please choose a short but
                        descriptive name (so anvi'o can use it whenever she
                        needs to name an output file, or add a new table in a
                        database, or name her first born).
(...)
```

We could add `fasta.txt` description a new optional column, `project_name`: if it is present and the contents of it is not null, the snakemake workflow would set the content there as the project name. But what concerns me here is the part of the help message that says "anvi'o can use it [to] add a new table in a database". Because that's when these characters become a problem.

Let me double check this and write back.

meren:
OK. That doesn't look like a very good direction. I will propose another solution soon.

PR #1453, opened 2020-06-26
Allow contigs db names with dots

meren:
This is related to #1452. 

I tested it with the pangenomic workflow, and everything was fine.


meren:
Tested this. I think it is good to go (famous last words?).

PR #1466, opened 2020-07-13
Overhaul Inspect Splits Page

isaacfink21:
This PR updates the inspect splits page with several additions, including:
- A nucleotide/amino acid display
- Coloring genes by COG, KEGG metabolic categories, or by gene ID
- Visualization of insertions/deletions
- Highlight bars for nucleotides


isaacfink21:
It should be fixed now - I must have committed a broad part of the anvio directory so I'll make sure to commit specific files in the future :) 

I also updated the amino acid sequence to work with genes in the reverse direction and added a button on the sidebar to toggle the nucleotide display.

meren:
Hey @isaacfink21, in `master` we are able to click on SNVs to see this panel:

image

But it seems SNVs are no longer 'clickable' :)

isaacfink21:
Thanks for pointing this out! I had disabled pointer-events on SVG text to prevent the selection of the sequences or x-scale - the latest commit should fix this and has a better implementation :)

PR #1468, opened 2020-07-19
Add ThreadedCommandRunner and ThreadedProdigalRunner

mooreryan:
See https://github.com/merenlab/anvio/pull/1437 for more information.

Unfortunately, i messed up the version control in that pull request and had to close it.  This one has all the stuff I added (I think at least...that previous pull request now has over 100 files changed so tough to make sure *everything* is there :/)  Very sorry for the confusion & mix up!

Here are a couple of relevant discussions from the previous pull request:

- Type annotations: https://github.com/merenlab/anvio/pull/1437#discussion_r440510077
  - I removed all type annotations in this version.
- Taking args: https://github.com/merenlab/anvio/pull/1437#discussion_r440504923
  - `MultithreadedProdigalRunner` takes `argparse.Namespace`
  - Should the base class `MultithreadedCommandRunner` also take `Namespace`?
- GFF file usage: https://github.com/merenlab/anvio/pull/1437#discussion_r440514864
- Shuffling fasta files: https://github.com/merenlab/anvio/pull/1437#discussion_r440516194

### DIFF
--- a/anvio/drivers/prodigal.py
+++ b/anvio/drivers/prodigal.py
@@ -1,17 +1,21 @@
 # coding: utf-8
 """Interface to Prodigal."""
 
+# Note: The Prodigal command will be run in a "mutlithreaded" way even if the user selects only one thread.  In other
+#   words, it will still "split" the input fasta into a single split, and run Prodigal in a new thread.  It seems
+#   wasteful, but with the Infant Gut test dataset, it only added a couple of seconds of overhead vs. the original
+#   single threaded code.  So it seems to not be a major performance issue as of now.
+import argparse
 import os
 
 import anvio
 import anvio.utils as utils
-import anvio.fastalib as fastalib
 import anvio.terminal as terminal
 import anvio.constants as constants
-import anvio.filesnpaths as filesnpaths
 
 from anvio.errors import ConfigError
 
+from anvio.threadingops import ThreadedProdigalRunner
 
 __author__ = "Developers of anvi'o (see AUTHORS.txt)"
 __copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
@@ -20,7 +24,6 @@ __license__ = "GPL 3.0"
 __version__ = anvio.__version__
 __maintainer__ = "A. Murat Eren"
 __email__ = "a.murat.eren@gmail.com"
-
 
 run = terminal.Run()
 progress = terminal.Progress()
@@ -35,6 +38,9 @@ class Prodigal:
 
         A = lambda x: (args.__dict__[x] if x in args.__dict__ else None) if args else None
         self.prodigal_translation_table = A('prodigal_translation_table')
+        self.num_threads = A('num_threads')
+
+        self.run.info('Num threads for gene calling', self.num_threads)
 
         self.parser = None
         self.installed_version = None
@@ -43,7 +49,6 @@ class Prodigal:
                                   'v2.60': self.__parser_1}
 
         self.check_version()
-
 
     def __parser_1(self, defline):
         """parses this:
@@ -66,7 +71,7 @@ class Prodigal:
 
         if not len(fields) != 8 or fields[6] not in ['1', '-1']:
             raise ConfigError('Somethings is wrong with this prodigal output :( The parser for the '
-                               'version %s is failing to make sense of it.' % (self.installed_version))
+                              'version %s is failing to make sense of it.' % (self.installed_version))
 
         hit['direction'] = 'f' if fields[6] == '1' else 'r'
         hit['start'] = int(fields[2]) - 1
@@ -83,7 +88,6 @@ class Prodigal:
 
         return hit
 
-
     def check_version(self):
         """checks the installed version of prodigal, sets the parser"""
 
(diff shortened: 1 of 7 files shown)


mooreryan:
There were a couple of more comments that have disappeared from the last pull request (I think they got marked as outdated), but one in particular was @meren mentioning it would be cool to have some logging about current number of gene caller worker threads running.

Off the top of my head, you could do something like this:

```python
active_worker_threads = [thread for thread in self.threads if thread.is_alive()]
while len(active_worker_threads) > 0:
    # log something about number of active worker threads
 
    # then sleep for some amount of time

    # then update active_worker_threads
```

Not sure how annoying it would be to ping every second with the active worker threads counts though.

meren:
Amazing work, @mooreryan. Thank you very much! I am merging this to `master` so it can be available to anyone who is tracking the active branch.

Do you have any statistics regarding the total performance improvement? Because not only you have revamped the gene calling step with multithreading support, but also improved how splits were identified, etc. It would have been great to have a summary of those data somewhere so we could add them to the release notes :)

mooreryan:
> Do you have any statistics regarding the total performance improvement? Because not only you have revamped the gene calling step with multithreading support, but also improved how splits were identified, etc. It would have been great to have a summary of those data somewhere so we could add them to the release notes :)

Pretty sure I have the original profiles kept.  Let me check on that!

mooreryan:
@meren Here are the runtimes/profiles you asked for.

Both of these are using the Infant Gut tutorial contigs (4,189 contigs, 35,766,167 nucleotides, ~8500 nuc/contig).

## Before

This one was from a commit before any of the work was done:

anvi-gen-contigs-db-profile-before

That's **273 seconds**.  Since it's single threaded, wall time and CPU time would be about that.

## After

With two CPUs for prodigal, plus all the other optimizations of different parts of the pipeline (eg `get_split_starts_and_stops`).

This is timed using commit 5b9facaa.

anvi-gen-contigs-db-profile-after-5b9facaa

Ignore the CPU time there...that's only reporting anvio's python cpu time and ignoring prodigal's cpu time.

The **total time was 159 seconds** (~1.7 times faster than the original).

And since it was multicore, here is the Unix `time` output...

```
real	2m40.598s
user	3m54.674s
sys	0m1.186s
```

Sure there's a bit of overhead from timing the pyinstrument run rather than without it, but it is small.

meren:
I tried with another assembly from [this study](https://www.biorxiv.org/content/10.1101/2020.04.29.069278v1) that had 23,877 contigs and 135,900,299 nucleotides.

``` bash
wget https://ndownloader.figshare.com/files/22467302 -O P-A-F.tar.gz
tar -zxvf P-A-F.tar.gz && cd P-A-F

# switch to v6.2
cd $anvio; git checkout v6.2; cd -

anvi-export-contigs -c CONTIGS.db -o CONTIGS.fa

##########################################################
#     on v6.2 with a single thread
##########################################################
rm -rf test.db; time anvi-gen-contigs-database -f CONTIGS.fa -o test.db

# (...)
# 
# Result .......................................: Prodigal (v2.6.3) has identified 130704 genes.
# 
# Contigs with at least one gene call ..........: 22612 of 23877 (94.7%)
# Contigs database .............................: A new database, test.db, has been created.
# Number of contigs ............................: 23,877
# Number of splits .............................: 24,257
# Total number of nucleotides ..................: 135,900,299
# Gene calling step skipped ....................: False
# Splits broke genes (non-mindful mode) ........: False
# Desired split length (what the user wanted) ..: 20,000
# Average split length (wnat anvi'o gave back) .: 22,985
# 
# real    17m5.803s
# user    14m15.357s
# sys     0m6.895s

##########################################################
#     on master with five threads
##########################################################
cd $anvio; git checkout master; git pull; cd -
rm -rf test.db; time anvi-gen-contigs-database -f CONTIGS.fa -o test.db -T 4

# (...)
# 
# Result .......................................: Prodigal (v2.6.3) has identified 130704 genes.
# 
# CONTIGS DB CREATE REPORT
# ===============================================
# Split Length .................................: 20,000
# K-mer size ...................................: 4
# Skip gene calling? ...........................: False
# External gene calls provided? ................: False
# Ignoring internal stop codons? ...............: False
# Splitting pays attention to gene calls? ......: True
# Contigs with at least one gene call ..........: 22612 of 23877 (94.7%)
# Contigs database .............................: A new database, test.db, has been created.
# Number of contigs ............................: 23,877
# Number of splits .............................: 24,257
# Total number of nucleotides ..................: 135,900,299
# Gene calling step skipped ....................: False
# Splits broke genes (non-mindful mode) ........: False
# Desired split length (what the user wanted) ..: 20,000
# Average split length (what anvi'o gave back) .: 22,985
# 
# real    8m43.918s
# user    18m0.312s
# sys     0m8.202s
```

It is 1026 seconds vs 523 seconds. Almost two times faster! A significant amount of time was spent outside of gene calling and mindful splitting.

ekiefl:
> A significant amount of time was spent outside of gene calling and mindful splitting.

I have a promise I have yet to uphold with @mooreryan, and that is using numba for kmer counting. Let's keep this test around for when I do that.

mooreryan:
That's a pretty nice result!  Kmer counting was the next big bottleneck in the infant gut dataset...probably a similar story in that larger set you tried

PR #1472, opened 2020-07-20
InteracDome

ekiefl:
This PR implements InteracDome into anvi'o.

Starting this early to potentially receive feedback. As things get closer I will add more context :)


anvio/parsers/hmmer.py
```diff
@@ -0,0 +1,732 @@
...
+            ('f',               str), # from (env coord)
+            ('f',               str), # to (env coord)
+            ('mean_post_prob',  str), # acc
+        ]
+
+        return list(zip(*col_info))
+
+
+    def get_search_results(self, noise_cutoff_dict = None):
+        """Goes through the hits provided by `hmmscan` and generates an annotation dictionary with the relevant information about each hit.
```
ivagljiva, line 642: Evan, I just wanted to let you know that I made some changes to this function in master last night (so that it could return a dictionary of bitscores when requested; this is related to our desire to compute new KEGG score thresholds). It was basically a crappy patch to get the functionality that I wanted out of the current hmmscan parser design.

That functionality will need to be incorporated somehow into your HMMER overhaul (which is looking very nice, by the way, and was much-needed :). Probably this is a good opportunity to turn it from a crappy patch into a robust design enabling bitscore retrieval. I also agree with you that the pre-processing of the hmmer output needs to be moved out of the driver class, so perhaps we could do that at the same time. 

Do you want to have a meeting about it at some point?

ekiefl: @ivagljiva thanks for your message. I will try and merge master into this branch and resolve merge conflicts. Once I do that, could you please just double check that I did not break the KEGG stuff?

I'm glad you agree that we should not preprocess the HMMER output. I think it would be worth discussing how to do that, however I think we should do it in a different branch and perhaps different time. Having spent a lot of time getting that code to do what I want, I would prefer stepping away from it for the time being because I found it very frustrating to work with.

ivagljiva: Sure! I'll take a look.

And that is completely understandable. You've spent heaps of effort on this already :)

ekiefl:
Here is the rationale behind this branch in blog form: http://merenlab.org/2020/07/22/interacdome/

meren:
The mother of all PRs,

image

has conflicts with not one, not two, but three files,

image

as it should.

ekiefl:
I think this is ready to merge, but I am tentative because these two issues on master (https://github.com/merenlab/anvio/issues/1477 @ivagljiva, https://github.com/merenlab/anvio/issues/1476 @meren) potentially prevent other bugs from revealing themselves during self tests. When you guys get a chance, could you please check these out?

ivagljiva:
> but I am tentative because these two issues on master (#1477 @ivagljiva, #1476 @meren) potentially prevent other bugs from revealing themselves

I will check it this evening :)

ekiefl:
> this commit message honestly had me so curious that I had to look up what 'don't mess with the zohan' was referring to :)

Lmao yes it is Adam Sandler's worst film. But they are all his worst film, which is what is so curious about his career. I have no idea why it popped into my brain.

Issue #1481, opened 2020-08-06
[BUG] anvi-script-visualize-split-coverages cannot make a fill chart using both --xlim flag and a samples.txt


watsonar:
## Short description of the problem

When making a fill chart (the default chart type) with `anvi-script-visualize-split-coverages`, you can either use the `--xlim` flag or a `samples.txt`, but not both at the same time without encountering an error. Adding `--chart-type line` fixes the issue.

## anvi'o version

```
Anvi'o version ...............................: esther (v6.2-master)
Profile DB version ...........................: 34
Contigs DB version ...........................: 18
Pan DB version ...............................: 14
Genome data storage version ..................: 7
Auxiliary data storage version ...............: 2
Structure DB version .........................: 2
Kegg Modules DB version ......................: 2
```

## System info

MacOS Catalina, and I installed anvi'o via the [youre-a-wizard-arry method](http://merenlab.org/2016/06/26/installation-v2/#following-the-active-codebase-youre-a-wizard-arry
) on the anvi'o installation page.

## Detailed description of the issue

When I run this command:
```
anvi-script-visualize-split-coverages -i coverages.txt \
                                      -o coverages.pdf \
                                      -d variability.txt \
                                      -w 30 \
                                      -s samples.txt \
                                      --snv-marker-transparency 1 \
                                      --snv-wobble-pos-color=#749637 \
                                      --snv-non-wobble-pos-color=#e33e33 \
                                      --snv-intergenic-color=#999999 \
                                      --snv-marker-width=0.1 \
                                      --free-y-scale
```
I get the desired output, split and colored according to my specifications in the `samples.txt` file, and no errors.

But when I run this command (adding the --xlim flag):
```
anvi-script-visualize-split-coverages -i coverages.txt \
                                      -o coverages.pdf \
                                      -d variability.txt \
                                      -w 30 \
                                      -s samples.txt \
                                      --snv-marker-transparency 1 \
                                      --snv-wobble-pos-color=#749637 \
                                      --snv-non-wobble-pos-color=#e33e33 \
                                      --snv-intergenic-color=#999999 \
                                      --snv-marker-width=0.1 \
                                      --free-y-scale \
                                      -x 2287,4232
```
I get this output and error message:
```
I, [2020-08-05 17:11:24.720491 #95741] INFO -- Checking options
I, [2020-08-05 17:11:24.742883 #95741] INFO -- Parsing input data
[...]

PR #1482, opened 2020-08-06
Changed xlim to coord_cartesian

watsonar:
Related to #1481. Changed `anvi-script-visualize-split-coverages` so that `--xlim` flag uses coord_cartesian rather than xlim.


ekiefl:
Looks perfect! I will merge right away!

PR #1485, opened 2020-08-11
Pretty printing of long run info lines

ivagljiva:
This PR introduces a small change to the prevalent `run.info` method allowing longer info text to be broken up and nicely aligned across different lines in the terminal. **This feature is now the default behavior of `run.info`, meaning that if you don't want lines to be broken up, you must explicitly turn the `align_long_values` flag off.**

To demonstrate, this is what the run info lines for `anvi-estimate-metabolism` output headers looked like before this change:
image

And this is what they look like afterwards:
image

A couple of notes:
- I turned off the `break_long_words` and `break_on_hyphens` flags so that file paths would not be broken across different lines (for convenient copy-paste)
- I also explicitly turned off this behavior for the run info lines with terminal commands for the same reason

### DIFF
--- a/anvio/kegg.py
+++ b/anvio/kegg.py
@@ -2006,7 +2006,7 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
         self.run.info("Module completion threshold", self.module_completion_threshold)
         self.run.info("Number of complete modules", metabolism_dict_for_list_of_splits["num_complete_modules"])
         if complete_mods:
-            self.run.info("Complete modules", ", ".join(complete_mods))
+            self.run.info("Complete modules", ", ".join(complete_mods), align_long_values=True)
 
         return metabolism_dict_for_list_of_splits
 
--- a/anvio/kegg.py
+++ b/anvio/kegg.py
@@ -1348,7 +1348,7 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
         run.warning(None, header="AVAILABLE OUTPUT MODES", lc="green")
 
         for mode, mode_meta in self.available_modes.items():
-            self.run.info(mode, mode_meta['description'], align_long_values=True)
+            self.run.info(mode, mode_meta['description'])
 
 
     def list_output_headers(self):
@@ -1359,7 +1359,7 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
             desc_str = header_meta['description']
             type_str = header_meta['mode_type']
             mode_str = "output modes" if header_meta['mode_type'] == 'all' else "output mode"
-            self.run.info(header, f"{desc_str} [{type_str} {mode_str}]", align_long_values=True)
+            self.run.info(header, f"{desc_str} [{type_str} {mode_str}]")
 
 
     def init_hits_and_splits(self):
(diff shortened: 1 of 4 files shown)


ivagljiva:
Oh, and I forgot to mention, I tested this by running `anvi-self-test --suite mini` and saw no issues. :)

anvio/terminal.py
```diff
@@ -351,6 +384,17 @@ def info(self, key, value, quiet=False, display_only=False, overwrite_verbose=Fa
         info_line = "%s%s %s: %s\n%s" % ('\n' * nl_before, c(label, lc),
                                          '.' * (self.width - len(label)),
                                          c(str(value), mc), '\n' * nl_after)
+        if align_long_values:
+            terminal_width = os.get_terminal_size()[0]
```
meren, line 388: there is already a global `get_terminal_size` function in `terminal.py` :)

I think it would be better if we stick with a single source of information for this.

anvio/kegg.py
```diff
@@ -2678,7 +2678,7 @@ def list_output_headers(self):
             desc_str = header_meta['description']
             type_str = header_meta['mode_type']
             mode_str = "output modes" if header_meta['mode_type'] == 'all' else "output mode"
-            self.run.info(header, f"{desc_str} [{type_str} {mode_str}]")
+            self.run.info(header, f"{desc_str} [{type_str} {mode_str}]", align_long_values=True)
```
meren, line 2681: these seem to be unnecessary since the default value is set to `True` in the function header. Unless there is another reason I am missing to be explicit here.

anvio/terminal.py
```diff
@@ -335,12 +335,45 @@ def write(self, line, quiet=False, overwrite_verbose=False):
                 sys.stderr.write(line.encode('utf-8'))
 
 
-    def info(self, key, value, quiet=False, display_only=False, overwrite_verbose=False, nl_before=0, nl_after=0, lc='cyan', mc='yellow', progress=None):
+    def info(self, key, value, quiet=False, display_only=False, overwrite_verbose=False, nl_before=0, nl_after=0, lc='cyan',
+             mc='yellow', progress=None, align_long_values=True):
+        """
+        This function prints information nicely to the terminal in the form:
+        key ........: value
```
meren, line 342: thank you for adding the docstring :)))

anvio/terminal.py
ivagljiva: oops, I missed that :) this is now fixed.

anvio/kegg.py
ivagljiva: no reason! I put this there before I changed the default to `True`, and forgot to change it back. This is also now fixed.

ivagljiva:
I ran more of the `anvi-self-test` suites and have found a few more bugs. I am fixing these now.

meren:
Great! :) Please feel free to merge after fixing those bugs so we can start testing it altogether.

ivagljiva:
I believe I have fixed everything, but please test it some more to make sure :)
I ran all of anvi'o's self tests. Only the `full` suite and `alons-classifier` suite threw errors, but I don't believe they are related to this PR because I get those same errors in `master` as well. Perhaps it is a good idea to resolve those errors before we launch `v7`?

I will merge now.

PR #1486, opened 2020-08-11
New parameter to takes best hit from cogs

mschecht:
Hi Meren,

Here is my solution for taking the best COG hit when multiple are provided. For example, when the user provides the new flag, annotations like `COG123!!!COG456` will become `COG123`.

Cheers,
Matt

### DIFF
--- a/anvio/synteny.py
+++ b/anvio/synteny.py
@@ -71,7 +71,7 @@ class NGram(object):
         self.skip_init_functions = A('skip_init_functions')
         self.genome_names_to_focus = A('genome_names')
         self.ngram_source = A("ngram_source")
-        self.take_first_cog = A("take_first_cog")
+        self.first_functional_hit_only = A("first_functional_hit_only")
         self.annotation_source_dict = {}
 
         self.pan_db_path = A('pan_db')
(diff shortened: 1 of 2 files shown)


anvio/synteny.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, args, run=terminal.Run(), progress=terminal.Progress(), skip_
         self.skip_init_functions = A('skip_init_functions')
         self.genome_names_to_focus = A('genome_names')
         self.ngram_source = A("ngram_source")
-
+        self.take_first_cog = A("take_first_cog")
```
meren, line 74: this would be better if it was `--fist-functional-hit-only`. why focus on COG? this will work for any functional annotation source :)

anvio/synteny.py
```diff
@@ -323,8 +324,25 @@ def annotate_window(self, window):
         # Annotate window based on user input
         gene_annotation_dict = {}
         for annotation_source, annotations_dict in self.annotation_source_dict.items():
-            ngram_annotation = tuple([annotations_dict[g] for g in window])
-            gene_annotation_dict[annotation_source] = ngram_annotation 
+            if self.take_first_cog:
+                if annotation_source == 'COG_FUNCTION':
```
meren, line 328: you don't need to test if the annotation_source is COG here. if you remove this `if` statement (and fix the indentation, of course), the rest of the code should work even if the user chose `KOFams` as `--ngram-source`.

bin/anvi-analyze-synteny
```diff
@@ -62,6 +62,8 @@ if __name__ == '__main__':
                         help="Provide this flag if you want anvi-analyze-synteny to report "
                               'Ngrams that contain gene calls that have no annotation.')
     groupC.add_argument(*anvio.A('genomes-names'), **anvio.K('genomes-names', {"required": False}))
+    groupC.add_argument('--take-first-cog',action="store_true",
+                        help="When TRUE, this flag will only take the best COG annotation")
```
meren, line 66: not necessary to mention COG here, either.

But it would be better to put a more helpful help (that clarifies in some cases a given gene may have multiple annotation, this flag will ensure blah blah).

anvio/synteny.py
mschecht: Good idea, I was thinking too specific.

mschecht:
Thanks for the suggestions Meren, I addressed them all in the most recent commit.

mschecht:
@meren do you think this is ready to be merged?

meren:
Ah, I'm very sorry that I forgot to get back to this! Thanks for the ping.

anvio/synteny.py
```diff
@@ -307,7 +307,8 @@ def annotate_window(self, window):
         """This method annotates a gene-callers-id window
 
         This method will annotate a gene-callers-id window based using annotation sources provided
-        by the user (e.g. COGs, pan_db)
+        by the user (e.g. COGs, pan_db). If the user provided the `take_first_cog` flag, the COG
```
meren, line 310: I think the docstring here needs an update since that flag does not exist anymore ;)

other than that, I think this is good to go. please merge it whenever you like, @mschecht. thank you very much!

mschecht: Thanks!

PR #1490, opened 2020-08-12
Prevent naughty characters from ever entering contigs database

ekiefl:
- Prevents database generation what incoming contigs fasta contains characters that are not A, C, T, G, N, a, c, t, g, or n.
- Adds new `--seq-type` parameter to `anvi-script-reformat-fasta` which fixes contig fastas that suffer from this ailment

### DIFF
--- a/anvio/dbops.py
+++ b/anvio/dbops.py
@@ -5,6 +5,7 @@
 """
 
 import os
+import re
 import sys
 import time
 import copy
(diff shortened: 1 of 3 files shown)


meren:
this looks great!

I like the idea of not moving forward and sending people to the reformat FASTA script if they wish.

ekiefl:
Great. Also, the timings are in. on the IGD, this check takes 0.18s compared to the 1m7s spent on kmer counts.

This solution takes 1.11s:

```python
for s in fasta.seq:
    if s not in allowed:
        raise ConfigError("asdf")
```

and this solution takes 0.34s:

```python
if set(fasta.seq) > allowed:
    raise ConfigError("asdf")
```

I know that regex is kind of crazy to be putting in our codebase, but it seems to be the fastest.

meren:
Thanks for checking :)

> I know that regex is kind of crazy to be putting in our codebase, but it seems to be the fastest.

Regex is so ugly, it very well knows that the only way it can ensure its survival is by showing everyone how neatly implemented it is.

PR #1493, opened 2020-08-21
init commit

mschecht:
Hi Meren, 

I added a flag to `anvi-script-reformat-fasta` that will allow a user to filter out sequences with a gaps percentage greater than the 3rd quartile of the percent gap distribution in an alignment. I unfortunately could not find another way than to load the fasta file twice which is probably not best practice. Please let me know if there is a better approach.

Cheers,
Matt

### DIFF
--- a/sandbox/anvi-script-reformat-fasta
+++ b/sandbox/anvi-script-reformat-fasta
@@ -119,7 +119,7 @@ def reformat_FASTA(args):
 
         num_gaps = fasta.seq.count('-')
         if args.export_gap_counts_table:
-            gaps_info_list.append([fasta.id, num_gaps, fasta.seq])
+            gaps_info_list.append([fasta.id, num_gaps])
 
         percentage_of_gaps = num_gaps * 100.0 / l
         if percentage_of_gaps >= args.max_percentage_gaps:
@@ -147,7 +147,7 @@ def reformat_FASTA(args):
             output.store(fasta, split = False)
 
     if args.export_gap_counts_table:
-        df = pd.DataFrame(gaps_info_list, columns=["header", "num_gaps", "sequence"])
+        df = pd.DataFrame(gaps_info_list, columns=["header", "num_gaps"])
         df.to_csv(args.export_gap_counts_table + ".tsv", sep='\t', index=False)
 
     if report_file:
--- a/sandbox/anvi-script-reformat-fasta
+++ b/sandbox/anvi-script-reformat-fasta
@@ -10,6 +10,8 @@ import anvio.terminal as terminal
 import anvio.constants as constants
 import anvio.filesnpaths as filesnpaths
 
+import statistics
+
 from anvio.errors import ConfigError, FilesNPathsError
 
 __author__ = "Developers of anvi'o (see AUTHORS.txt)"
@@ -81,8 +83,20 @@ def reformat_FASTA(args):
     total_num_nucleotides_modified = 0
     total_num_contigs_removed = 0
 
+    # Get stats for arg.filter_third_quartile
+    percent_of_gaps_list = []
     while next(fasta):
+        l = len(fasta.seq)
+        percentage_of_gaps = fasta.seq.count('-') * 100.0 / l
+        percent_of_gaps_list.append(percentage_of_gaps)
+    standard_deviation = statistics.pstdev(percent_of_gaps_list)
+    average = statistics.mean(percent_of_gaps_list)
+    third_quartile = average + standard_deviation
 
+    # Reopen fasta
+    fasta = u.SequenceSource(args.contigs_fasta)
+
+    while next(fasta):
         l = len(fasta.seq)
 
         total_num_nucleotides += l
@@ -118,6 +132,12 @@ def reformat_FASTA(args):
             total_num_nucleotides_removed += l
             total_num_contigs_removed += 1
             continue
+
+        if args.filter_third_quartile_gaps == True:
+            if percentage_of_gaps >= third_quartile:
+                total_num_nucleotides_removed += l
+                total_num_contigs_removed += 1
+                continue
 
         if args.simplify_names:
             if prefix:
@@ -160,6 +180,10 @@ if __name__ == '__main__':
                         help="Maximum fraction of gaps in a sequence (any sequence with \
                               more gaps will be removed from the output FASTA file). The \
                               default is %(default)f.")
+    parser.add_argument('--filter-third-quartile-gaps', type=bool, default=False, metavar='PERCENTAGE',
+                        help="Sequences with a percentage of gaps greater than the third  \
+                              quartile of the gap percentage distribution will be removed \
+                              more gaps will be removed from the output FASTA file (Boolean).")
     parser.add_argument('-i', '--exclude-ids', required=False, metavar='TXT FILE',
                         help="IDs to remove from the FASTA file. You cannot provide both\
                               --keep-ids and --exclude-ids.")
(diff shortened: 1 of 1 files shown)


ekiefl:
@mschecht From our offline conversation, it seems that quartiles may not be suitable because you are guaranteed to throw away 25% of the data, even if none deserve it. I would recommend looking into mean absolute deviation, it may do what you want. We already use this in our codebase in `utils`:

```
def get_list_of_outliers(values, threshold=None, zeros_are_outliers=False, median=None):
    """Return boolean array of whether values are outliers (True means outlier)

    Modified from Joe Kington's (https://stackoverflow.com/users/325565/joe-kington)
    implementation computing absolute deviation around the median.

    Parameters
    ==========
    values : array-like
        A num_observations by num_dimensions array of observations.

    threshold : number, None
        The modified z-score to use as a thresholdold. Observations with
        a modified z-score (based on the median absolute deviation) greater
        than this value will be classified as outliers.

    median : array-like, None
        Pass median of values if you already calculated it to save time.

    Returns
    =======
    mask : numpy array (dtype=bool)
        A num_observations-length boolean array. True means outlier

    Examples
    ========

    Create an array with 5 manually created outliers:

    >>> import numpy as np
    >>> import anvio.utils as utils
    >>> array = 10*np.ones(30) + np.random.rand(30)
    >>> array[5] = -10
    >>> array[9] = -10
    >>> array[12] = -10
    >>> array[15] = -10
    >>> array[23] = -10
    >>> mask = utils.get_list_of_outliers(array, threshold=5)
    >>> print(mask)
    [False False False False False  True False False False  True False False
     True False False  True False False False False False False False  True
     False False False False False False]

    As can be seen, mask returns a numpy array of True/False values, where True corresponds to
    outlier values.

    >>> print(outlier_indices)
    >>> outlier_indices = np.where(mask == True)[0]
    [ 5  9 12 15 23]

    The `True` values indeed occur at the indices where the values hand been manually changed to
    represent outliers.

    References
    ==========
    Boris Iglewicz and David Hoaglin (1993), "Volume 16: How to Detect and
    Handle Outliers", The ASQC Basic References in Quality Control:
    Statistical Techniques, Edward F. Mykytka, Ph.D., Editor.

    http://www.sciencedirect.com/science/article/pii/S0022103113000668
    """

    if threshold is None:
        threshold = 1.5

    if len(values.shape) == 1:
        values = values[:, None]

    if not median: median = np.median(values, axis=0)

    diff = np.sum((values - median) ** 2, axis=-1)
    diff = np.sqrt(diff)
    median_absolute_deviation = np.median(diff)

    if not median_absolute_deviation:
       if values[0] == 0:
            # A vector of all zeros is considered "all outliers"
            return np.array([True] * values.size)
       else:
            # A vector of uniform non-zero values is "all non-outliers"
            # This could be important for silly cases (like in megahit) in which there is a maximum value for coverage
            return np.array([False] * values.size)

    modified_z_score = 0.6745 * diff / median_absolute_deviation
    non_outliers = modified_z_score > threshold

    if not zeros_are_outliers:
        return non_outliers
    else:
        zero_positions = [x for x in range(len(values)) if values[x] == 0]
        for i in zero_positions:
            non_outliers[i] = True
        return non_outliers
```

Something to think about.

mschecht:
Thank you @ekiefl and @meren for supervising this branch. Per offline conversations, I have simplified this branch to add only a `--max-gaps` flag which will allow the user to filter out sequences that have more gaps then the defined threshold and a `--export-gap-counts-table` flag which exports a tsv with the gap counts per sequence. This will be useful for the user to explore the distribution of gap counts in their alignment. 

> What happens when you declare that flag when sequences have no gaps? Or one sequence with gaps?

I tested these cases and it appears to check out.

sandbox/anvi-script-reformat-fasta
```diff
@@ -113,12 +116,21 @@ def reformat_FASTA(args):
             total_num_contigs_removed += 1
             continue
 
-        percentage_of_gaps = fasta.seq.count('-') * 100.0 / l
+        num_gaps = fasta.seq.count('-')
+        sequence_num_gaps_list = [fasta.id, num_gaps, fasta.seq]
+        num_gaps_list.append(sequence_num_gaps_list)
+        percentage_of_gaps = num_gaps * 100.0 / l
```
meren, line 122: `sequence_num_gaps_list` does not look like a necessary variable. This change would remove the need for it completely:

``` python
num_gaps_list.append([fasta.id, num_gaps, fasta.seq], )
```

That list variable looks more like `gaps_info_list` than `num_gaps_list` to me.

Another note, that list is populated even if `args.export_gap_counts_table` is False, but it is only used for anything if `args.export_gap_counts_table` is True down below.

mschecht: Thanks Meren, I just fixed these issues in the most recent commit.

sandbox/anvi-script-reformat-fasta
```diff
@@ -160,10 +177,15 @@ if __name__ == '__main__':
                         help="Maximum fraction of gaps in a sequence (any sequence with \
                               more gaps will be removed from the output FASTA file). The \
                               default is %(default)f.")
+    parser.add_argument('-M','--max-gaps', type=float, default=1000000,
+                        help="Maximum amount of gaps allowed per sequence in the alignment.")
```
ekiefl, line 181: How about adding something like this: "Don't know what to pick? Use the --export-gap-counts-table to decide an appropriate number"

meren:
If it is working for you feel free to merge after considering @ekiefl's last suggestion for the help text, @mschecht. I think it has reached to a level that is ready for `master` testing.

sandbox/anvi-script-reformat-fasta
```diff
@@ -133,6 +146,10 @@ def reformat_FASTA(args):
         else:
             output.store(fasta, split = False)
 
+    if args.export_gap_counts_table:
+        df = pd.DataFrame(gaps_info_list, columns=["header", "num_gaps", "sequence"])
+        df.to_csv(args.export_gap_counts_table + ".tsv", sep='\t', index=False)
+
```
ekiefl, line 152: Users you want this info probably don't want a reformatted FASTA, they first just want this table so they can decide the `--max-gaps` they want to use. Am I understanding right?

mschecht: Yeah it's just to quickly examine the distribution of gap counts per sequence. I actually left the sequences column in by accident because I was double-checking in R that the gap counts were correct, but left it because I thought it might be useful for the user. Should I take out the sequence column?

sandbox/anvi-script-reformat-fasta
mschecht: Gread idea!

sandbox/anvi-script-reformat-fasta
```diff
@@ -81,8 +83,10 @@ def reformat_FASTA(args):
     total_num_nucleotides_modified = 0
     total_num_contigs_removed = 0
 
-    while next(fasta):
+    if args.export_gap_counts_table:
+        gaps_info_list = []
```
ekiefl, line 87: This should go above with all of the other `if/else` logic.

sandbox/anvi-script-reformat-fasta
ekiefl: Well, it is totally up to you. But keep in mind that you don't control how big the user's sequences are. Then again it is an MSA so it can't be **that** crazy big.

ekiefl:
Looks good to me, @mschecht

sandbox/anvi-script-reformat-fasta
mschecht: That's I good point, I removed the column.

PR #1504, opened 2020-09-13
anvi-cluster-contigs += --log-file

meren:
Anvi'o automatically assigns a temp file path to store the STDOUT and STDERR while running clustering algorithms, however, this causes issues when the run fails on HPCs and clusters where the temporary file path is not accessible from the head node.

The new parameter `--log-file` enables the user to pass an explicit file path (so it is accessible form everywhere).


meren:
(this is a follow up from Charles Xu's report on anvi'o Slack)

PR #1537, opened 2020-10-23
Leeway parameter

mschecht:
Hi Meren,

I added a `flank-length` parameter to `anvi-get-sequences-for-gene-calls`. This will allow a user to extend the nucleotide gene calls with flanking regions past the start and stop position.

Cheer,
Matt


mschecht:
@meren I took a stab at exporting the external-gene-calls file along with the flanked sequences. Let me know if this is the right direction when you have a chance.

meren:
I'm taking a look at this. Give me 20 minutes without changing the code please :)

meren:
None of these are right so far since the external gene calls file should have start / stop positions for the new sequences, not for the contigs. I will take a look at this after the lab meeting.

PR #1540, opened 2020-10-27
Read only DB

ekiefl:
Currently, all DB instances must pass a `check_if_db_is_writable` method, even if the database is used for read-only purposes. This PR addresses this by introducing a new parameter called `read_only`, conditionally running `check_if_db_is_writable` based on `read_only`'s value, and decorating all methods with write operations with a gatekeeper, that prevents the method being called if `read_only` is set to `True`. In other words, this PR represents a procrastination from paper writing.

@meren, @ivagljiva, @mschecht, @semiller10, tagging you all so you know about this new feature.


ekiefl:
It passes `run_all_tests.sh`

meren:
Looks great to me!

meren (approved):

PR #1543, opened 2020-11-02
Contig level metagenome mode

ivagljiva:
Hello everyone, and happy Feature 🧊 Freeze 🧊 Day :D 

This PR contains the final changes to `anvi-estimate-metabolism` that I want to be included in `v7`. The changes are primarily concerned with metagenome mode of `anvi-estimate-metabolism`, which now estimates module completeness for each contig in a metagenome individually (before, we were essentially handling metagenome the same way we did isolate genomes, which over-estimated module completeness).

It passes all tests in `anvi-self-test --suite metabolism`.

## Overview of changes
Here are the major changes to the codebase that were necessary to make this new metagenome mode possible:
- metabolism outputs no longer report modules with 0 completeness score unless explicitly requested with the `--include-zeros` parameter. This is to avoid excessively long output files in metagenome mode (in which most contigs are likely to have 0 completeness score for most modules).
- the `filesnpaths` module now contains an `AppendableFile` class that enables our programs to repeatedly append output to the same file. We use this to incrementally add data to the output files while estimation is running rather than creating an output file all at once after estimation is finished.
  - initialization opens the file handle, an `append()` function can be called to add data to the file, and a `close()` function is used to explicitly close the file handle
  - the `append()` function currently supports the `str` and `dict` (specifically, dictionary of dictionaries) datatypes but can be expanded to support more datatypes (ie dataframes) as needed. Strings are appended as new lines in the file. In the `dict` case, the first dictionary to be added to the file determines the header of the file and all subsequent dictionaries that are appended must have the same keys as are in the header. 
  - the `append()` function accepts keyword arguments so that it can play nice with other anvi'o functions. For example, appending dictionaries relies upon the `utils.store_dict_as_TAB_delimited_file()` method, and users can use keyword arguments to send any needed parameters for that function through `append()`
- `utils.store_dict_as_TAB_delimited_file()` function now has the option to keep the file handle open if you pass `do_not_close_file_obj=True`. `AppendableFile` uses this option so that it can keep appending to the file after the initial dictionary is added.

A brief demonstration of how to use the `AppendableFile` class:
```
import anvio.filesnpaths as filesnpaths

## Simplest Use Case
file = filesnpaths.AppendableFile(output_path) # this opens a file handle to the provided path
file.append(some_data_to_append)  # the data is added to the file and the I/O buffer is flushed
file.close()  # this closes the file handle

## More Complex Initialization
## append_type parameter: requires all data given to the append() function to be that datatype
## fail_if_file_exists=False: do not throw error if we are trying to append to an existing file
## overwrite_if_file_exists=True: if the file already exists when we initialize the AppendableFile, overwrite it (instead of appending to the existing file)
file = filesnpaths.AppendableFile(output_path, append_type=dict, fail_if_file_exists=False, overwrite_if_file_exists=True)

## Using kwargs with append()
file.append(output_dictionary, key_header="unique_id", headers=header_list)

file.close()
```

Small note for people who are interested in speed gains: I originally had the AppendableFile class open and close a file handle to the same path each time append() was called. Changing this to keeping one file handle open from initialization to `close()` improved metagenome mode processing on the Infant Gut dataset from `0:30:07.014325` to `0:23:21.809579`. 

Finally, this PR also includes the following miscellaneous changes:
- The output of `anvi-estimate-metabolism is now much less verbose when estimating for bins in a collection and for metagenome mode. Instead of showing similar warnings and output for every bin or contig, there is now simply a progress bar. 
- a couple of fixed bugs and an additional sanity check against requesting `kofam_hits` output mode with JSON input

## Things that could still be improved
There are two remaining issues that I did not have time to fix. I do not believe they are critical, but if anyone thinks otherwise please let me know 🙂 
1. Better memory management with the metabolism 'super-dictionaries'.  Using `AppendableFile` allows us to incrementally write the output files one bin/contig/multi-mode entity at a time, but after this is done we do not release the memory for the data that was put in the file. As of right now, we still build up two large super-dictionaries containing all the metabolism estimation information for all genomes/bins/contigs (one dictionary for KEGG modules and one for KOfam hits), because some applications (`--matrix-format` and `anvi-display-metabolism`) rely on the metabolism estimation functions to return all the data. It would be excellent if we had some way to release this memory so that the memory consumption of `anvi-estimate-metabolism` does not get too large for very large datasets. However, in order to do this we will probably need to adjust how the matrices are created when the `--matrix-format` option is used. 
2. The `unique_id` column is problematic when we are appending data to multiple output files at once. This ID is established when we convert the multi-level super-dictionaries of data into a simpler two-level dictionary for writing output (the IDs are the keys of the outer dictionary). IDeally - see what I did there? I'm here all night folks -  each output file would have a `unique_id` column that starts at 0 and increases by 1 each row. This is what happens when we have just one output file. However, when we have multiple output files that we are appending to incrementally after each set of splits, we get 'jumps' in the value of the `unique_id` within the same file because they all use the same `self.unique_id` variable. (Restarting the `unique_id` at 0 each time we convert a dictionary for printing results in repeating `unique_id` values within the same file, which is a much worse problem than the 'jumps'.

I welcome any feedback and ideas on how to address these remaining issues. :) I have some ideas for (1) but I am not sure how long it will take to implement so I have left it alone for now. 

@semiller10 I have a specific request for you to review the two commits that I made when you helped me fix the rogue trnaseq file name changes. Can you make sure these commits will not mess up your code? Thank you!


ekiefl:
> It would be excellent if we had some way to release this memory so that the memory consumption of anvi-estimate-metabolism does not get too large for very large datasets. However, in order to do this we will probably need to adjust how the matrices are created when the --matrix-format option is used.

Do you know that memory is an issue?

anvio/filesnpaths.py
```diff
@@ -368,3 +368,192 @@ def get_name_from_file_path(file_path, postfix_separator="."):
...
+
+    Parameters
+    ==========
+    file_path : string
+        the path to the file
+    append_type : data type, None
+        If you provide a data type here, then only data of that type will be allowed to be appended to the file
+    fail_if_file_exists : boolean
+        if True, throw an error upon initialization if the file already exists. If False,
+        start appending to the end of the file.
```
ekiefl, line 384: I think `fail_if_file_exists` should be `False` by default. I think this because the class name is `AppendableFile`, so programmers who use it are expecting to append to a file that already exists, not be complained at because the file exists.

anvio/filesnpaths.py
```diff
@@ -368,3 +368,192 @@ def get_name_from_file_path(file_path, postfix_separator="."):
...
+    overwrite_if_file_exists : boolean
+        if True (and fail_if_file_exists is False), the existing file will be removed before adding new data.
```
ekiefl, line 386: I would argue this functionality lies outside the responsibility of a class called `AppendableFile`

ekiefl:
@ivagljiva Now `AppendableFile` is a context manager, so it can be used like so:

```
#! /usr/bin/env python

from anvio.filesnpaths import AppendableFile

filename = 'testing'
with AppendableFile(filename, fail_if_file_exists=False) as f:
    f.append("hi")
``` 

Let me know what you think.

ivagljiva:
@ekiefl, 

> Now AppendableFile is a context manager

First, this is great, thank you for adding it :) And also thank you for reviewing this PR, it was an oversight on my part not to add you to the reviewer list. 

Second,

> Do you know that memory is an issue?

To be totally honest, no, I have not explicitly confirmed this with tests on large datasets. Memory issues have been in the back of my mind since before this PR - there is a documented example on a large dataset [here](https://github.com/merenlab/lab-fiesta/issues/955#issuecomment-663638130). While this PR fixes the issue associated with generating the output files, I assumed that very large datasets may still use excessive memory since I do not release the memory for the superdictionary after processing each set of splits. 

But like I said, I haven't actually tested this yet. I am going to submit a large job now and get back to you.

anvio/filesnpaths.py
ivagljiva: Okay, this makes sense. I will change it.

anvio/filesnpaths.py
ivagljiva: I see your point. I wanted this functionality because when I re-run `anvi-estimate-metabolism` with the same output file prefix, I did not want those output files to be continuously appended to, I wanted to clear out the old data and then start appending. But I realize now that I can control this from the estimation function that opens the output files instead of doing it from the `AppendableFile` class. I think I will do that instead.

(🤔 this also just made me realize that not overwriting existing files will allow us to combine data from different runs of `anvi-estimate-metabolism` into the same file, which could be useful for some people. I will implement that as well, thanks for the inspiration!)

ivagljiva:
I finished running some memory profiling and wanted to document the memory issue better. I attempted to run `anvi-estimate-metabolism --metagenome-mode` on a group of 11 large metagenomes using multi-mode. Here is the graph of memory usage over time:
memory_plot_20201104150048

As you can see, memory increases linearly over time (as more data is added to the super-dictionaries that pass the data around).

I will note that this job did not finish. It did not even finish processing the first metagenome, whose database is 927 MB in size. I allocated 200 GB of memory to the job, and the job was killed when it got to 200635.800781 MiB or about ~210 GB of memory usage. 

The solution for this will be to free the memory in between each set of splits that we process. Since the data is appended to any long-format output files as we process, this will not be a problem for long-format output. Matrix mode needs to be re-done so that no longer relies on keeping the entire data dictionary in memory until all splits have been processed. The only other thing to keep in mind is that if we want to eventually be able to estimate metabolism on-the-fly while binning, doing so will probably require keeping the entire data dictionary in memory, which means we may need to leave that option open for use in certain cases. 

For now, before the memory issue is fixed, users who want to process many large metagenomes with `--metagenome-mode` would be better off running metabolism estimation on each one separately and combining the output files afterwards. (Unless they are blessed with oodles of RAM.)

PR #1549, opened 2020-11-06
anvi-display-contigs-stats update

meren:
This PR is primarily to change this section of the output interface,

image

to show only the distribution of SCGs within a metagenome like this:

image

Because only the distribution of hits for HMMs of `singlecopy` type are really relevant for this section.


meren:
@FlorianTrigodet, please take a look at the code to see the changes I've made, and merge it to main when you are satisfied :)

PR #1558, opened 2020-11-16
Indel filter

ekiefl:
Alright, so in this I nerfed `--min-indel-fraction` and replaced it with a coverage-dependent threshold that must be exceeded--identical to the one used in SNV identification. This will help keep down indels with low `count` values that sneak into the database because the `coverage` is low.

In my test dataset, here are the differences:

| value                   | Before | After  |
|-------------------------|--------|--------|
| indel counts            | 6288   | 1111   |
| mean coverage of indels | 23\.15 | 48\.26 |
| mean count of indels    | 2\.96  | 10\.12 |
| count/coverage          | 0\.128 | 0\.210 |

These all seem expected to me.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2085,7 +2085,7 @@ D = {
                      "may result in very large output files and millions of reports, but you are the "
                      "boss). Do not forget to take a look at '--min-coverage-for-variability' parameter. "
                      "Also note that this flag controls indel reporting: normally '--min-coverage-for-variability' "
-                     "and '--min-indel-fraction' control whether or not indels should be reported, but with this "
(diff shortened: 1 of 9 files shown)


ekiefl:
In this branch, `min_indel_fraction` is no longer an entry of the `self` table. Is it important to remove this with a migration script?

meren:
This all looks very good to me, @ekiefl. Thank you very much for going after this.

> In this branch, min_indel_fraction is no longer an entry of the self table. Is it important to remove this with a migration script?

I don't think it will cause any issues and we can leave it as is. It will continue to live in a small fraction of profile databases generated by the users of the main branch without influencing anything.

Issue #1567, opened 2020-11-24
[FEATURE] New COG database version 2020


Sirbius:
## The need
Hi there,
A new version of the COG database has been released in September 2020 and since there are good improvement, as announced [here](https://www.ncbi.nlm.nih.gov/research/cog), I was wondering if we could import it on anvi'o.

## The solution
Ideally, just few changes in the anvi-setup-ncbi-cogs script where it points at the [ftp page](https://www.ncbi.nlm.nih.gov/research/cog-project/) should do the job (?). 

## Beneficiaries
Everyone who is working with COG annotation and is tired to read so many undefined proteins.

Not sure if it's the right place too, but besides the COG function, the COG category in the enrichment file would also be great for downstream analysis and fancy plots.

Thank you guys! You work is amazing!! 

Silvia

xvazquezc:
I'd suggest to keep some backwards compatibility if somebody needs to use the older (current) version. I think there is something alike in the scg data used from the GTDB

meren:
Thanks for the report, @Sirbius. I will look into this (although a very preliminary investigation showed that they changed everything, so it will take some effort to get it into anvi'o).

> I'd suggest to keep some backwards compatibility if somebody needs to use the older (current) version

I agree, that would be ideal. But I am not sure what would be the best way to do it. Thinking out loud here, the best way I can think of would be the following:

* Write a migration script to rename all occurrences of `COG_CATEGORY` and `COG_FUNCTION` in existing contigs databases to `COG14_CATEGORY` and `COG14_FUNCTION`.
* Add a --cog-version flag to `anvi-setup-ncbi-cogs` and `anvi-run-ncbi-cogs` (the default would be 2020, and would annotate contigs databases with `COG20_CATEGORY` and `COG20_FUNCTION`, but the user can change it to 2014 to setup databases and run them).


Best,

meren:
Regarding this part of your comment @Sirbius, 

> Not sure if it's the right place too, but besides the COG function, the COG category in the enrichment file would also be great for downstream analysis and fancy plots.

@ivagljiva is working on generalizing @adw96's method to compute enrichment statistic, which will accept any functional source to calculate enrichment scores. Since we store `COG_FUNCTION` and `COG_CATEGORY` as distinct functional sources per gene, in theory you will be able to do enrichment analyses on `COG_CATEGORY` seamlessly :)

ivagljiva:
[...]

PR #1570, opened 2020-11-26
Restructure COGs module to make room for NCBI's 2020 release

meren:
NCBI released a new version of COGs (the last release was in 2014), and given how often we use this resource, we had to update anvi'o so it can work with it.

This PR does a few things:

* Turns the COGs module into one that can work with multiple releases (both to setup data from different releases, and to run different releases on the same contigs database .. it really was fun).
* Automatically salvages the existing COGs data on the user's system and puts it in the `COG14` folder under the COGs base directory.
* Updates existing contigs databases so COG_* becomes COG14_* if there are any COG annotations.

So there is no hassle on the file system side, no data loss in contigs databases, and the opportunity to be able to work with multiple releases of COGs (for science, I guess).

I thank @Sirbius for bringing this up to our attention via #1567, and @xvazquezc for pushing me over the fence by explicitly asking to keep the existing COG data (which resulted in a much better design, actually).

I tried multiple different scenarios and it seems everything is working.


ivagljiva (approved):
Very impressive, Meren :) 

I have tested this on my own system and it is working well for me. I fixed a small bug while I was at it so I just wanted to bring it to your attention. COGs was not installed for me so I tried installing the 2014 version first. I got the following warning from `anvi-setup-ncbi-cogs --cog-version COG14`:
```
WARNING
4631 of 4631 COG IDs that appear in the list of orthology domains file (which
links protein IDs to COG names), are missing from the COG names file (which
links COG IDs to function names and categoires). Because clearly even the files
that are distributed together should not be expected to be fully compatible.
Anvi'o thanks everyone for their contributions.
```

Downstream, `anvi-run-ncbi-cogs -c TEST_CONTIGS.db -T 4 --cog-version COG14` finished successfully but I also saw the following warning:
image

And it seems that in the database none of the COG annotations had associated functions.
image

The reason for this bug was that this line
```
self.cogs_found_in_cog_names_file.add(COG)
```
was only being run for COG20, not for COG14. I fixed it and now I do not get the above warnings, and my database has been properly annotated. :)

Please double check that I have not inadvertently broken something else while fixing this.

meren:
Brilliant save, @ivagljiva! Thank you very much for looking into this 😊

Issue #1556, opened 2020-11-12
[FEATURE] --help should include `__resources__` tags


ekiefl:
## The need

`anvi-help` produces great info about resources. So do the anvi'o help docs. It would be nice if when `--help` is used in a command, the information in the `__resources__` variables was displayed.

![image](https://user-images.githubusercontent.com/8688665/98947044-4e3bf000-24ba-11eb-83d8-c4cc0c8ff95a.png)

## The solution

`argparse` gymnastics

## Beneficiaries

Everyone.

meren:
Also it would be great to provide a link to the online resource for a given program (unless we want to implement a terminal Markdown viewer which we certainly should not) 😂

ekiefl:
> Also it would be great to provide a link to the online resource for a given program 

This is the feature request. For example, right now we have:

```
▶▶ anvi-compute-gene-cluster-homogeneity -h
usage: anvi-compute-gene-cluster-homogeneity [-h] -p PAN_DB
                                             [-g GENOMES_STORAGE]
                                             [-o FILE_PATH] [--store-in-db]
                                             [--gene-cluster-id GENE_CLUSTER_ID]
                                             [--gene-cluster-ids-file FILE_PATH]
                                             [-C COLLECTION_NAME]
                                             [-b BIN_NAME]
                                             [--quick-homogeneity]
                                             [-T NUM_THREADS] [--just-do-it]

Compute homogeneity for gene clusters

RESOURCES:

The role of gene cluster
homogeneity described in the Anvi'o pangenomics tutorial:
http://merenlab.org/2016/11/08/pangenomics-v2/#inferring-the-homogeneity-of-
gene-clusters

optional arguments:
  -h, --help            show this help message and exit

INPUT FILES:
  Input files from the pangenome analysis.

  -p PAN_DB, --pan-db PAN_DB
                        Anvi'o pan database
  -g GENOMES_STORAGE, --genomes-storage GENOMES_STORAGE
                        Anvi'o genomes storage file

REPORTING:
  How do you want results to be reported? Anvi'o can produce a TAB-delimited
  output file for you (for which you would have to provide an output file
  name). Or the results can be stored in the pan database directly, for
  which you would have to explicitly ask for it. You can get both as well in
  case you are a fan of redundancy and poor data analysis practices. Anvi'o
  does not judge.

  -o FILE_PATH, --output-file FILE_PATH
                        File path to store results.
  --store-in-db         Store analysis results into the database directly.

SELECTION:
[...]

PR #1572, opened 2020-11-30
Connecting the help menu for anvi'o programs with online help

meren:
This PR overloads the `argparse` module to add additional information at the end of the help menu for anvi'o programs and scripts IF there is an entry for them in the `docs` module.

This is related to #1556 and the `epiloge` can be extended with additional information.

----

Here is an example for `anvi-profile`:

image

Here is an example for `anvi-script-gen-CPR-classifier` (for which here is no entry in `docs`):

image

You tell me if it is crazy useful or crazy useless.


meren:
(it learns the version the user is using, and constructs the URL according to that).

anvio/argparse.py
```diff
@@ -0,0 +1,118 @@
...
+        return help_text + "\n"
+
+
+    def get_args(self, parser):
+        """A helper function to parse args anvi'o way.
+
+           This function allows us to make sure some ad hoc parameters such as `--debug`
+           can be used with any anvi'o program spontaneously even if they are not explicitly
+           defined as an accepted argument, yet flags (or parameters) anvi'o does not expect
+           to see can still be sorted out.
```
ekiefl, line 112: Wrong indentation

ekiefl:
This is fantastic. I'm positive people will use this.

ekiefl:
I agree with leaving out resource links. People will click the online help, which has the resource links.

ivagljiva (approved):
Awesome! This will be crazy useful, Meren :) 🎉

meren:
With the last commit this is how the very end of `anvi-interactive -h` looks like in dark terminal:

image

versus in a light terminal:

image

If a program doesn't have an entry under `anvio/docs/programs` this is how it looks then:

image

ivagljiva:
can I just say that I LOVE how it goes "can consume (CHEESE), can provide (PIZZA), and you can learn more about how it does that with (FIRE)". 

it makes so much sense. Anvi'o is making pizza for me.

PR #1579, opened 2020-12-07
Artifacts relating to anvi'o Snakemake workflows may need to be added

mschecht:
Hi Everyone,

With @ekiefl 's recent [commit](https://github.com/merenlab/anvio/commit/cfcda91d72fcabc9dc67cd14319681f3f79985c7) I saw that `anvi-migrate` can now migrate our snakemake config files. This inspired me to add artifacts related to snakemake. 

Please let me know if I am heading in the right direction.

Cheers,
Matt

### DIFF
--- a/anvio/docs/__init__.py
+++ b/anvio/docs/__init__.py
@@ -88,6 +88,12 @@ ANVIO_ARTIFACTS ={
         "provided_by_anvio": False,
         "provided_by_user":True
     },
+    "fasta-txt": {
+        "name": "FASTA TXT",
+        "type": "TXT",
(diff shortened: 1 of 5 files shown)


meren:
It is a good idea. But I would prefer `workflow-config` rather than `snakemake-config` for accuracy, and type `JSON` rather than `TXT` because that's what it is :)

meren:
Also it would have been very useful if there was a link to the anvi'o snakemake workflows post from within the artifact description :)

mschecht:
Thanks for the ideas, just addressed them!

PR #1582, opened 2020-12-08
anvi-run-workflow --get-default-config error

mschecht:
Hey @ekiefl I was getting an error making a default config on main. You can reproduce it on the main branch as follows:

```
anvi-run-workflow -w contigs --get-default-config config.json
```
Screen Shot 2020-12-08 at 11 14 21 AM

I added a conditional in this commit to skip checking the version of the config if the user is requesting the default version of a workflow config with `anvi-run-workflow --get-default-config error`.

### DIFF
--- a/anvio/workflows/__init__.py
+++ b/anvio/workflows/__init__.py
@@ -76,12 +76,9 @@ class WorkflowSuperClass:
             filesnpaths.is_file_json_formatted(self.config_file)
             self.config = json.load(open(self.config_file))
 
-        # Check if config version is current unless getting default
-        # config which should already be the current version
-        if self.default_config_output_path:
-           pass
-        else:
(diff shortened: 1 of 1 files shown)


mschecht:
Thanks @ekiefl !

ekiefl:
Thank you for identifying and fixing the problem. Based on your PR I modified things slightly and I'll explain here why.

You identified the problem occurs when a default config file is being generated, so you said if the default config file name is not `None`, don't bother with this check. But the error occurs for a more general reason: `self.config `was `None`. In principle there are many reasons why the class is instantiated without a config file, not just when creating a default config file. So your fix wouldn't address such instances.

mschecht:
Thanks for the explanation @ekiefl, great way for me to learn!

meren:
🎊

PR #1585, opened 2020-12-09
getting help broken link

matthewlawrenceklein:
in top-left additional options dropdown menu

// also added sample-newick-tree.txt to testing sandbox :)


matthewlawrenceklein:
@ivagljiva !

ivagljiva (approved):
The link works for me, @matthewlawrenceklein! IMO you can merge this if you want to.

I have a piece of advice for you. :) When committing, it is usually best to make them as atomic as possible, so that your commit message describes everything that you have added. In this example, I would have put the newick tree in a separate commit that says something like 'add newick tree'. If you add multiple things at a time and then want to split them into different commits, you can use `git add -p` to go through each chunk of code and commit them separately (I use this all the time because I am always forgetting to commit between individual changes). This is just good practice so that later if you need to go through the commit history or revert some changes it is easy to see what changed where. But it is not a big issue to forget to do this, so do not worry about it here :) 

(Also I am sorry, I think that you were asking me about this earlier, but I misunderstood and thought you were asking about whether to put the tree in the PR description. My bad!)

Anyhow, congrats on your first PR!

meren:
> When committing, it is usually best to make them as atomic as possible

For that, I think GitHub Desktop is also very very useful (to see what you're about to commit and/or push). It also simplifies the management of branches :)

ekiefl:
> > When committing, it is usually best to make them as atomic as possible
> 
> For that, I think GitHub Desktop is also very very useful (to see what you're about to commit and/or push). It also simplifies the management of branches :)

Or if you use develop in vim, https://github.com/tpope/vim-fugitive is an amazing git plugin for vim where you can pick and choose each line you want to be in which commit

meren:
I think this is good to go now :) Thank you, @matthewlawrenceklein!

PR #1590, opened 2020-12-10
High GC leads to low quality ORF prediction: research

ekiefl:
This is a new way of doing things for me, and I wonder if it will work out. Basically, this pull request will never be merged into master. In fact, right after the is posted, I will close it. But ideally, this will serve as a comprehensive resource for what I did and why I did it, so that when I submit my branch that _will_ end up being merged, all of the design decisions can be learned from this janky version of anvi'o that I unforgivably hack.

# The problem

In this PR https://github.com/merenlab/anvio/pull/1428 I added a frame prediction so that if people use external genomes without a `aa_sequence` column, we try and predict the best frame for each gene and forgive the user for internal stop codons and non-divisible by 3 errors.

But in this issue https://github.com/merenlab/lab-fiesta/issues/1144 I determined that frame prediction is very inaccurate for genomes with GC content > 50%. This is because the Markov model I used is based on the uniref50 database, which presumably is full of human, yeast, and ecoli, all of which have GC content < 50%.

Here I try and find a solution to the high GC orf prediction problem.

### DIFF
--- a/anvio/tables/genecalls.py
+++ b/anvio/tables/genecalls.py
@@ -322,7 +322,6 @@ class TablesForGeneCalls(Table):
 
         from collections import Counter
         frame_count = Counter()
-        correct = []
         quality = []
 
         # the main loop to go through all the gene calls.
@@ -363,10 +362,9 @@ class TablesForGeneCalls(Table):
             elif predict_frame:
                 # no amino acid sequence is provided, BUT USER WANTS FRAME TO BE PREDICTED
                 # we may be good, if we can try to predict one for it.
-                frame, amino_acid_sequence, qual = utils.get_most_likely_translation_frame(sequence, model=model)
+                frame, amino_acid_sequence, d = utils.get_most_likely_translation_frame(sequence, model=model)
                 frame_count[frame] += 1
-                correct.append('correct' if frame == 0 else 'incorrect')
-                quality.append(qual)
+                quality.append(d)
 
                 if frame is None:
                     # we not good because we couldn't find a frame for it. because this gene call has no predicted frame,
@@ -431,9 +429,11 @@ class TablesForGeneCalls(Table):
 
             amino_acid_sequences[gene_callers_id] = amino_acid_sequence
 
-        print(frame_count)
         import pandas as pd
-        pd.DataFrame({'outcome': correct, 'quality': quality}).to_csv('results.txt', sep='\t', index=False)
+        df = pd.concat(quality)
+        print(df.outcome_stop.value_counts())
+        print(df.outcome_prob.value_counts())
+        df.to_csv('results.txt', sep='\t', index=False)
 
         # reporting time
         self.run.warning(None, header="EXTERNAL GENE CALLS PARSER REPORT", lc="cyan")
--- a/anvio/tables/genecalls.py
+++ b/anvio/tables/genecalls.py
@@ -233,7 +233,7 @@ class TablesForGeneCalls(Table):
         """Recover amino acid sequences for gene calls in a gene_calls_dict.
 
         If 'aa_sequence' exists as keys in the gene_calls_dict[<key>] objects, this function will take
-        those seqeunces into consideration and use them without trying to predict frames even if the gene
+        those sequences into consideration and use them without trying to predict frames even if the gene
         call is partial. So user-defined aa sequences in `aa_sequence` column will have priority.
 
         Please note this FIXME: By reading all contig sequences into memory, anvi'o does a pretty bad job
@@ -280,7 +280,7 @@ class TablesForGeneCalls(Table):
                     estimated_length_for_aa = gene_length / 3
                     user_provided_aa_length = len(amino_acid_sequences[gene_callers_id])
 
-                    # there is already a sanity check for htis, but one can't be too careful
+                    # there is already a sanity check for hits, but one can't be too careful
                     if gene_calls_dict[gene_callers_id]['call_type'] != constants.gene_call_types['CODING'] and user_provided_aa_length:
                         raise ConfigError("You have provided an amino acid sequence for at least one gene call in your external gene calls "
                                            "(%d) file that was not marked as CODING type :(" % gene_callers_id)
@@ -320,6 +320,11 @@ class TablesForGeneCalls(Table):
                   "num_partial_genes_with_internal_stops": 0,
                   "num_genes_with_internal_stops": 0}
 
+        from collections import Counter
+        frame_count = Counter()
+        correct = []
+        quality = []
+
         # the main loop to go through all the gene calls.
         for gene_callers_id in gene_calls_dict:
             gene_call = gene_calls_dict[gene_callers_id]
@@ -358,7 +363,10 @@ class TablesForGeneCalls(Table):
             elif predict_frame:
                 # no amino acid sequence is provided, BUT USER WANTS FRAME TO BE PREDICTED
                 # we may be good, if we can try to predict one for it.
-                frame, amino_acid_sequence = utils.get_most_likely_translation_frame(sequence, model=model)
+                frame, amino_acid_sequence, qual = utils.get_most_likely_translation_frame(sequence, model=model)
+                frame_count[frame] += 1
+                correct.append('correct' if frame == 0 else 'incorrect')
+                quality.append(qual)
 
                 if frame is None:
                     # we not good because we couldn't find a frame for it. because this gene call has no predicted frame,
@@ -422,6 +430,10 @@ class TablesForGeneCalls(Table):
                                       "sense as a gene call: '%s'." % (str(gene_callers_id), sequence, amino_acid_sequence))
 
             amino_acid_sequences[gene_callers_id] = amino_acid_sequence
+
+        print(frame_count)
+        import pandas as pd
+        pd.DataFrame({'outcome': correct, 'quality': quality}).to_csv('results.txt', sep='\t', index=False)
 
         # reporting time
         self.run.warning(None, header="EXTERNAL GENE CALLS PARSER REPORT", lc="cyan")
--- a/anvio/tables/genecalls.py
+++ b/anvio/tables/genecalls.py
@@ -323,6 +323,8 @@ class TablesForGeneCalls(Table):
         from collections import Counter
         frame_count = Counter()
         quality = []
+        correct = 0
+        incorrect = 0
 
         # the main loop to go through all the gene calls.
         for gene_callers_id in gene_calls_dict:
@@ -363,7 +365,11 @@ class TablesForGeneCalls(Table):
                 # no amino acid sequence is provided, BUT USER WANTS FRAME TO BE PREDICTED
                 # we may be good, if we can try to predict one for it.
                 frame, amino_acid_sequence, d = utils.get_most_likely_translation_frame(sequence, model=model)
-                frame_count[frame] += 1
+                if frame == 0:
+                    correct += 1
+                else:
+                    incorrect += 1
+                frame_count['correct' if frame == 0 else 'incorrect'] += 1
                 quality.append(d)
 
                 if frame is None:
@@ -431,6 +437,18 @@ class TablesForGeneCalls(Table):
 
         import pandas as pd
         df = pd.concat(quality)
+        from pathlib import Path
+
+        name = anvio.dbops.ContigsDatabase(self.db_path).get_meta_value('project_name')
+        gc = sum(utils.get_GC_content_for_FASTA_entries().values())/len(utils.get_GC_content_for_FASTA_entries(name + '.fa'))
+
+        print(correct)
+        print(incorrect)
+        print(gc)
+        print(name)
+        with open('many_genomes.txt', 'a') as f:
+            f.write(f"{name}\t{total}\t{correct}\t{incorrect}\t{GC}")
+        #print(frame_count)
         print(df.outcome_stop.value_counts())
         print(df.outcome_prob.value_counts())
         df.to_csv('results.txt', sep='\t', index=False)
--- a/anvio/tables/genecalls.py
+++ b/anvio/tables/genecalls.py
@@ -447,6 +447,9 @@ class TablesForGeneCalls(Table):
             gc = 0
         total = correct + incorrect
 
+        output = pd.concat(quality)
+        output.to_csv("results.txt", sep='\t', index=False)
+
         print(correct)
         print(incorrect)
         print(gc)
--- a/anvio/tables/genecalls.py
+++ b/anvio/tables/genecalls.py
@@ -261,10 +261,7 @@ class TablesForGeneCalls(Table):
                 raise ConfigError("The task at hand calls for the use of the anvi'o Markov model to predict proper open reading "
                                   "frames for external gene calls when necessary, but the model does not seem to be in the right "
                                   "place in the anvi'o codebase. FAILING BIG HERE.")
-
             model = numpy.load(model_path)
-            null_prob = numpy.median(model)
-            stop_prob = model.min()/1e6
 
         gene_caller_ids_with_user_provided_amino_acid_sequences = set([])
 
@@ -361,7 +358,7 @@ class TablesForGeneCalls(Table):
             elif predict_frame:
                 # no amino acid sequence is provided, BUT USER WANTS FRAME TO BE PREDICTED
                 # we may be good, if we can try to predict one for it.
-                frame, amino_acid_sequence = utils.get_most_likely_translation_frame(sequence, model=model, stop_prob=stop_prob, null_prob=null_prob)
+                frame, amino_acid_sequence = utils.get_most_likely_translation_frame(sequence, model=model)
 
                 if frame is None:
                     # we not good because we couldn't find a frame for it. because this gene call has no predicted frame,
--- a/anvio/tables/genecalls.py
+++ b/anvio/tables/genecalls.py
@@ -369,7 +369,7 @@ class TablesForGeneCalls(Table):
                     correct += 1
                 else:
                     incorrect += 1
-                frame_count['correct' if frame == 0 else 'incorrect'] += 1
+                frame_count[frame] += 1
                 quality.append(d)
 
                 if frame is None:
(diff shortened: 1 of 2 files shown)


ekiefl:
Before fixing the problem, I need to better understand the problem. Here is what I think is going one during mis-predicted frames. There are 3 possible frames each gene could be in. The Markov model produces probability values for each of them (since the values are all very small, I log-transform them). The one with the highest probability gets picked. But my thinking is that the quality of the prediction should basically be defined by how well the best one "stands out" from the other 2. For this, I designed a metric:

```
quality = 1 - (p2 - p3)/(p1 - p3)
```

where p1, p2, and p3 are the log probabilities for the 1st, 2nd, and 3rd most likely frames. p1 "standing out" would be if p1 = 1e-3 and p2 = p3 = 1e-6. In that case, `quality=1`. Conversely, p1 not "standing out" would be like p1 = p2 = 1e-3 and p3 = 1e-6. In that case, `quality=0`.

So in 13065a4, I took a the following genome: [A_shaii-contigs.fa.zip](https://github.com/merenlab/anvio/files/5675210/A_shaii-contigs.fa.zip)

and ran the following code:

```
#! /usr/bin/env bash

anvi-gen-contigs-database -f A_shaii-contigs.fa -T 6 -o A_shaii.db
anvi-export-gene-calls --gene-caller prodigal -c A_shaii.db -o A_shaii_gene_calls --skip-sequence-reporting

anvi-gen-contigs-database -f A_shaii-contigs.fa -o A_shaii.db --external-gene-call
```

First, a contigs database is created in which prodigal predicts the genes. I treat prodigal as ground truth. Then, export the gene calls, skipping sequence reporting. Then, I use those gene calls to create a new contigs database. Since sequences aren't in the gene calls, frames will be predicted. If the predicted frame is `0`, then it predicts the same frame that prodigal did, and consider it "correct". If the predicted frame is `1`, or `2`, it is incorrect. During this process, I store whether the frame is correct and the quality for each prediction in a dataframe (`results.txt`) that looks like this:

```
╒═══════════╤════════════╕
│ outcome   │    quality │
╞═══════════╪════════════╡
│ incorrect │   0.136439 │
├───────────┼────────────┤
│ incorrect │ 0.00221783 │
├───────────┼────────────┤
│ correct   │   0.577716 │
├───────────┼────────────┤
│ correct   │   0.909472 │
├───────────┼────────────┤
│ incorrect │     0.1257 │
├───────────┼────────────┤
│ correct   │   0.269286 │
├───────────┼────────────┤
│ incorrect │   0.112254 │
├───────────┼────────────┤
│ correct   │   0.508966 │
├───────────┼────────────┤
│ incorrect │   0.015858 │
├───────────┼────────────┤
│ incorrect │   0.162888 │
├───────────┼────────────┤
│ incorrect │   0.144929 │
├───────────┼────────────┤
│ incorrect │   0.196087 │
├───────────┼────────────┤
│ incorrect │  0.0481753 │
├───────────┼────────────┤
│ incorrect │  0.0219251 │
├───────────┼────────────┤
│ incorrect │   0.218957 │
├───────────┼────────────┤
│ correct   │  0.0864383 │
├───────────┼────────────┤
│ incorrect │ 0.00244503 │
├───────────┼────────────┤
│ incorrect │  0.0864946 │
├───────────┼────────────┤
│ correct   │   0.411179 │
╘═══════════╧════════════╛
```

Then, I compare the distributions of quality scores for correct and incorrect predictions with this Rscript:

``` R
#! /usr/bin/env Rscript

library(tidyverse)

df <- read_tsv("results.txt")

p <- ggplot() +
    geom_histogram(data=df %>% filter(outcome == "incorrect"), aes(x=quality), fill='red', alpha=0.5) +
    geom_histogram(data=df %>% filter(outcome == "correct"),   aes(x=quality), fill='green', alpha=0.5)
print(p)


```

Which makes this plot:

image

Ideally, these distributions would be completely separated, because then the a quality score cutoff could define whether a gene is bad or not. It does not, but the results are not too bad either. `quality` scores above 0.5 are very likely to be correct.

I think the line of attack should be to start counting internal stop codons for all predictions with `quality < 0.5`.

ekiefl:
NOTE: must remain open for new commits to appear

### Can number of stop codons serve better prediction capabilities than the Markov model?

Maybe we should say screw markov models and just use the number of stop codons as the predictor. In 48134ce I compare the predictions of these two methods via

```
rm -rf A_shaii_ext.db; anvi-gen-contigs-database -f A_shaii-contigs.fa -o A_shaii_ext.db --external-gene-calls A_shaii_gene_calls --ignore-internal
```

This outputs the following results (and a `results.txt`) for further analysis:

```
correct      2565
incorrect     576
Name: outcome_stop, dtype: int64
correct      2413
incorrect     728
Name: outcome_prob, dtype: int64
```

Wow! just counting stop codons produces more correct results.

### Do these methods generally get the same predictions wrong or do each have their own strengths?

Even if you pick the best prediction in each case, we still get over 439 incorrectly predicted frames:

```
library(tidyverse)

df <- read_tsv("results.txt")
print(df %>% mutate(outcome_either = ifelse(outcome_prob == 'correct' | outcome_stop == 'correct', 'correct', 'false')) %>% select(outcome_either) %>% table())
```

Yields

```
correct   false 
   2702     439 
```

So there are differences between the predictions, but they generally fail for the same genes.

ekiefl:
### Something very promising

In _every_ case where counting stop codons produces the incorrect prediction, the difference in # of stop codons between the best and second best options is either 0 or 1:

```
#! /usr/bin/env Rscript

library(tidyverse)

df <- read_tsv("results.txt")

df %>%
    mutate(id=row_number()) %>%
    filter(outcome_stop == 'incorrect') %>%
    select(id, stop_count_0, stop_count_1, stop_count_2) %>%
    pivot_longer(cols=c(stop_count_0, stop_count_1, stop_count_2)) %>%
    group_by(id) %>%
    filter(value != max(value)) %>%
    mutate(diff = max(value)-min(value)) %>%
    select(id, diff) %>%
    distinct() %>%
    .$diff %>%
    table() %>%
    print()
```

Given this, I have 2 ideas. Count stop codons. If the best and second best differ by 0 or 1 codons, 

1. Run the markov model and pick the best. But **critically**, do not penalize stop codons. Instead, give them null probabilities via `stop_prob=np.median(model)`
2. Award points if it starts with methionine. Award points if it ends in a stop codon.

ekiefl:
### Proposed solution

First, I decided to look at option (2) since its easier, and it seems like it would work very well. To test this I created 23bebba and

ekiefl:
### Proposed solution

First, I decided to look at option (2) since its easier, and it seems like it would work very well. To test this I created 23bebba and ran the command:

```
rm -rf A_shaii_ext.db; anvi-gen-contigs-database -f A_shaii-contigs.fa -o A_shaii_ext.db --external-gene-calls A_shaii_gene_calls --ignore-internal
```

which produced a `results.txt` file. It seems that in cases where counting stop codons fails, the correct frame starts with `M` in 524 of 576 cases and ends with `*` in 576 of 576 cases:

```R
#! /usr/bin/env Rscript

library(tidyverse)

df <- read_tsv("results.txt")
df %>% filter(outcome_stop == 'incorrect') %>% .$starts_m_0 %>% table() %>% print()
df %>% filter(outcome_stop == 'incorrect') %>% .$ends_stop_0 %>% table() %>% print()
```

This produces

```
.
FALSE  TRUE 
   52   524 

.
TRUE 
 576 
```

Based on this, I think it may be in our best interests to scrap the the Markov model. I think in my next commit I will implement some heuristics to look at for the M at the start and * at the stop, **in cases where the best and second best frames differ by 0 or 1 stop codons** and see accurate the predictions are.

ekiefl:
### Proposed solution 2

After chatting with meren we are thinking the following.

1. Find the number of internal stop codons in each
2. If the best and second best frames differ in their ISCs by greater than 1, we are done. Lowest wins.
3. If not, we run a markov with median probabilities assigned to stop codons, so that they do not bias the model
4. If quality score of markov prediction > X, we are done. highest probability wins.
5. If the quality score of the markov prediction is bad, we see if any of the frames have terminal stop codons. If only 1 does, it wins. NOTE: due to the nature of the  canonical stop codons (TAA, TGA, TAG), at most one of the frames can end in a stop codon.
6. If none of them have terminal stops, look at for ATG at the start. If any of them have it, they win.

Something like this. To test if this works, I do some tests in ________

ekiefl:
### Proposed solution 2

After chatting with meren we are thinking the following.

1. Find the number of internal stop codons in each
2. If the best and second best frames differ in their ISCs by greater than 1, we are done. Lowest wins.
3. If not, we run a markov with median probabilities assigned to stop codons, so that they do not bias the model
4. If quality score of markov prediction > X, we are done. highest probability wins.
5. If the quality score of the markov prediction is bad, we see if any of the frames have terminal stop codons. If only 1 does, it wins. NOTE: due to the nature of the  canonical stop codons (TAA, TGA, TAG), at most one of the frames can end in a stop codon.
6. If none of them have terminal stops, look at for ATG at the start. If any of them have it, they win.

Something like this. To test if this works, I do some tests in c7e4bee and show that this plan sucks because the markov model is intrinsically bad for high GC content genomes, and so this whole idea falls apart.

Running `rm -rf A_shaii_ext.db; anvi-gen-contigs-database -f A_shaii-contigs.fa -o A_shaii_ext.db --external-gene-calls A_shaii_gene_calls --ignore-internal` as per usual, I then run

```R
library(tidyverse)

df <- read_tsv("results.txt")

# Of those in step 3, the markov model with median-value stop probabilities scores only 433 correct,
# and 1254 wrong. This false positivity rate is so high that quality scores often exceed 0.5,
# despite being wrong.
df %>% filter(stop_qual %in% c(0,1)) %>% .$outcome_prob %>% table() %>% print()
```

Using the markov model in its current state just won't work. And I really really hate relying on terminal stop codons and methionines because this totally falls apart for partial genes. Only using internal stop codons gets every fifth codon wrong.

ekiefl:
### Proposed solution 3

So the markov model sucks right? No! It does an _amazing_ job for low GC content:

image

In this proposed solution, its no holds barred: make several markov models for different GC content. Hold tight.

ekiefl:
Ok so the plan is to build the Markov models for different GC ranges. Since for uniref50 (1) I cannot find corresponding DNA sequences for the protein sequences in uniref50 and (2) seems to bias lower GC sequences, I am using a different database: ~66% of GTDB v89. From a bunch of contigs dbs, I created a single fasta file (27GB) of all the protein sequences defined by prodigal. In each header contains the genome-averaged GC content that the protein came from. This was done with the following script, which takes a file of contigs db paths:

```python
#! /usr/bin/env python

import anvio
import anvio.db as db
import anvio.fastalib as u

import argparse

from pathlib import Path

ap = argparse.ArgumentParser()
ap.add_argument("-c", "--contigs-db-paths")
ap.add_argument("-b", "--bins")
ap.add_argument("-o", "--output-fasta")

args = ap.parse_args()

# ---------------------------------------

output = u.FastaOutput(args.output_fasta)
contigs_db_paths = [x.strip() for x in open(args.contigs_db_paths, 'r').readlines()]
gc_dist = []

count = 0
for contigs_db_path in contigs_db_paths:
    print(count)

    try:
        cdb = db.DB(contigs_db_path, None, ignore_version=True)
        name = Path(contigs_db_path).stem

        basic_info = cdb.get_table_as_dataframe('contigs_basic_info')
        gc = (basic_info['gc_content'] * basic_info['length'] / basic_info['length'].sum()).sum()
        gc_dist.append(gc)

        seqs = cdb.get_table_as_dataframe('gene_amino_acid_sequences')
    except:
        count += 1
        continue

    for i, row in seqs.iterrows():
        if row['sequence'] is not '':
            output.write_id(f"{name}|{row['gene_callers_id']}|{gc:.4f}")
            output.write_seq(row['sequence'], split = False)

    count += 1

output.close()

with open('gc_distribution', 'w') as f:
    f.write('\n'.join([str(gc) for gc in gc_dist]))
```

I ran it with `python create_aa_seq_db.py -c db_paths -o aa_seq_db.fa`, where `db_paths` is 23k db paths, one per line. The whole thing took 3 hours. This also outputs a file `gc_distribution`, that has genome-averaged GC content for each genome. Visualizing the distribution with the following script:

```R
#! /usr/bin/env Rscript

library(tidyverse)

df <- read_tsv('gc_distribution', col_names=F)

g <- ggplot(data=df) +
    geom_histogram(aes(X1))
print(g)
```

yields the following distribution:

image

I am surprised to see a bimodal distribution. Regardless, I think its reasonable to have 5 different models, and I want them each to have equal representation. That means my bins should be

```R
cut_number(df$X1, 5) %>% unique()
```

or 

```
Levels: [0,0.387] (0.387,0.46] (0.46,0.561] (0.561,0.645] (0.645,1.00]
```

This way, each model is defined by 4692 genomes each. With ~3000 genes per genome, and ~330 aas per gene, the empirical transition probabilities are defined by around ~5 billion data points each. This should be plenty.

ekiefl:
### Using a bacteria-built Markov model is more important than accounting for GC content

Let's get back to A_shaii. Running all 5 of these models yields the following prediction accuracies:

| name      | correct | incorrect |
|-----------|---------|-----------|
| uniprot50 | 2413    | 728       |
| GC_0-39   | 3104    | 37        |
| GC_39-46  | 3076    | 65        |
| GC_46-56  | 2865    | 276       |
| GC_56-64  | 2255    | 886       |
| GC_64-100 | 1789    | 1352      |

First of all, go GC_0-39. That's the kind of accuracy I'm talking about. I haven't looked at log likelihood differences between frames but maybe those 37 have poor quality scores and can avoid being false positives. Second of all, why is the low GC content superior for a high GC model? I checked for 5 genomes and all are best predicted by GC_0-39. It must have something to do with wobble base pairs or something crazy like that. Regardless, this accuracy exceeds how the uniprot50 markov model performed even on low GC genomes. Tagging you @meren for any potential input before I code this up.

meren:
So weird. This must have something to do with the way genetic code is organized. I was wondering whether A_shaii GC-content would be around 40%, but it is not the case at all.

I find it kinda scary to rely on a model that is trained exclusively using `GC<.40` that performed well in your tests with only 5 genomes. But I am fine with this solution since you will have other levels of scrutiny before things come here :)

ekiefl:
Hey @meren given how good this model is performing, the highest accuracy approach I can think of  is applying this model at the onset. It unreasonably out performs everything else. Counting internal stop codons yields unresolvable conflict in 1200 of 3000 a shaii genes--this isn't nearly robust enough to use as a first line of defense. I will do a large scale analysis in the pull request to hopefully put my money where my mouth is.

xvazquezc:
Hi @meren and @ekiefl ,
just a couple of things to consider, if you don't mind...

Please don't forget that the metagenomic HMM profile is also meant to call archaeal genes. Not sure if it will have a lot influence in the HMM, but keep them in mind

Have you tested a model generated from an even sampling of the GTDB (both bacteria and archaea) and compare to the performance of the uniref50? Seems to me like a more objective start point to improve the overall model. Nonetheless, in no case there will be a better model than the ones that are genome-specific, which could be another thing to think about - something like gene re-calling, at least for MAG/genome based workflows

Changes is the assignment of stop codons are more taxa/phylogeny-dependent than related to the GC% of the genome. Also, in ref to solution 2, ATG as start codon is not universal. This is a bit old (2008 https://www.tandfonline.com/doi/pdf/10.1080/13102818.2008.10817533) and very few genomes were analysed based on today's terms (158), but it shows ATG to be start codon in 'only' 78% of genes

ekiefl:
Hey @xvazquezc thanks for following this discussion.

> Please don't forget that the metagenomic HMM profile is also meant to call archaeal genes. Not sure if it will have a lot influence in the HMM, but keep them in mind

This is a great point (by the way, this is not a hidden markov model, it is simply a markov model). I am currently testing `MM_GC_0-39` on archaeal and bacterial genomes from the GTDB and will update the performance improvements here. 

> Have you tested a model generated from an even sampling of the GTDB (both bacteria and archaea) and compare to the performance of the uniref50?

No, but this is a totally reasonable suggestion. I agree it is the most objective sampling. Though with the results I will post soon, I don't think it is absolutely necessary.

> Nonetheless, in no case there will be a better model than the ones that are genome-specific, which could be another thing to think about - something like gene re-calling, at least for MAG/genome based workflows

If I understand you correctly, you are suggesting that we figure out the taxonomy of the contig, and then apply a model that has been trained for that species? I don't see this being at all feasible but interested to hear more.

> Changes is the assignment of stop codons are more taxa/phylogeny-dependent than related to the GC% of the genome. 
Also, in ref to solution 2, ATG as start codon is not universal. This is a bit old (2008 https://www.tandfonline.com/doi/pdf/10.1080/13102818.2008.10817533) and very few genomes were analysed based on today's terms (158), but it shows ATG to be start codon in 'only' 78% of genes

I completely agree. This is why I don't want to use start/stop heuristics _at all_. That's why I want to use a MM trained on amino acid sequences, and it seems that `MM_GC_0-39` or perhaps a more objective model including archaeal genomes will fit the bill nicely (comparison to the frame predicted by prodigal is incoming).

Unfortunately, this model will not work for genomes that employ very non-standard translation tables. But if you have any experience working with such genomes, it would be interesting how this model fairs, even though it works on the assumption of the standard translation table. If you have any human-curated gene calls for such genomes (that are anvi'o formatted), please send them my way.

Cheers.

ekiefl:
Alright so here are 2000 bacterial genomes from the GTDB, with gene calls defined by Prodigal. In the below plot, each point is a genome, where the x-axis is the average GC content of the genome, and the y axis is the fraction of genes that had the same predicted frame as Prodigal. In red shows the uniref50 MM, and in green shows the GTDB bacterial genome (GC < 39%) MM. Clearly, there is a massive divergence from Prodigal using the uniref50 MM. The GTDB MM massively improves, but is subject to the same phenomenon, of diverging at high GC content.

image

The very, very worst is 94.5% accuracy. This is definitely still a worrying amount--several hundred if there are 10k+ genes. But on average, this model gives 99.55% consistency to Prodigal. For the 190 archaeal genomes, the story is much the same: worst consistency is 96.4, with an average of 99.54%.

xvazquezc:
> If I understand you correctly, you are suggesting that we figure out the taxonomy of the contig, and then apply a model that has been trained for that species? I don't see this being at all feasible but interested to hear more.

No, just run Prodigal in normal mode (`-p normal`) as you do when doing gene-calling in a single genome. Prodigal, by default, trains itself on the genome provided to create a gene model subsequently used to perform the gene calls. This is actually the basic behaviour of all standard *de novo* gene calling programs, both prokaryotic and eukaryotic. Of course, this requires a reasonable completeness/quality of the genome/MAG to work at its best.

> This is interesting to know. This is why I will not be using ATG/stop codon heuristics at all, as was previously suggested in this thread. Instead, I want to use a MM trained on amino acid sequences, and it seems that `MM_GC_0-39` or perhaps a more objective model including archaeal genomes will fit the bill nicely (comparison to the frame predicted by prodigal is incoming).
> 
> Of course, the effectiveness of this approach is completely dependent on the assumption that genomes use standard translation tables. If they don't, then this model will mistranslate. That said, if someone is studying a genome that does not use the standard genetic code and have their own gene calls, they should _really_ be supplying the corresponding amino acid sequences, in which case this model would never be ran by anvi'o. In theory we could add yet another column to the gene calls file with a string that specifies which translation table anvi'o should use for that gene. But this is already such a niche case that it does not feel worth it.

I was only taking into account those using translation table 11 (bacteria and archaea)... taking into account others would be... well, not even sure if possible.

Issue #1554, opened 2020-11-11
[Technical Problem] Config Error: List index out of range during anvi-setup-kegg-kofams


jzrapp:
## Short description of the problem

Dear anvio team, @ivagljiva and @meren,

I'm trying to play with your new estimate-metabolism functionality, and in order to do so setup kegg kofams. However, the setup keeps getting interrupted when it gets to Parsing KEGG Module M00006.

## anvi'o version
```
anvi-self-test --version
```

Anvi'o version ...............................: esther (v6.2-master)
Profile DB version ...........................: 35
Contigs DB version ...........................: 19
Pan DB version ...............................: 14
Genome data storage version ..................: 7
Auxiliary data storage version ...............: 2
Structure DB version .........................: 2
Kegg Modules DB version ......................: 2
tRNA-seq DB version ..........................: 1

## System info
I'm working on a remote server and followed the instructions for a conda setup for Linux/Windows users to install the anvio master branch

```
conda create -y --name anvio-master python=3.6
conda activate anvio-master

conda install -y -c conda-forge -c bioconda prodigal \
                 mcl \
                 muscle \
                 hmmer \
                 diamond==0.9.14 \
                 blast \
                 megahit \
                 bowtie2 \
                 bwa \
                 samtools \
                 centrifuge \
                 trimal \
                 iqtree \
                 fastani \
                 fasttree \
                 r-base \
                 r-stringi \
                 r-tidyverse \
                 r-magrittr \
                 r-optparse \
                 bioconductor-qvalue \
                 trnascan-se


# setup the code directory and get the anvi'o codebase
mkdir -p ~/github && cd ~/github/
git clone --recursive https://github.com/meren/anvio.git

cd ~/github/anvio/
pip install -r requirements.txt

#while still in anvio env, copy this:

cat <<EOF >${CONDA_PREFIX}/etc/conda/activate.d/anvio.sh
export PYTHONPATH=\$PYTHONPATH:~/github/anvio/
export PATH=\$PATH:~/github/anvio/bin:~/github/anvio/sandbox
echo -e "\033[1;34mUpdating from the master repository in GitHub \033[0;31m(press CTRL+C to cancel)\033[0m ..."
cd ~/github/anvio && git pull && cd -
EOF


echo -e "\n# >>> ANVI'O STUFF >>>" >> ~/.bash_profile
echo 'alias anvi-activate-master="conda activate anvio-master"' >> ~/.bash_profile
[...]

PR #1599, opened 2020-12-16
Issue 1554 - hmmer multithread fix

ivagljiva:
This PR addresses the issue in #1554 where an exception occurring in one HMMER worker thread did not kill the whole process, resulting in incomplete output. 

As @ekiefl suggested, I copied his approach from `anvi-profile` and converted the HMMER driver class from using threads to using the `multiprocessing` class. Output and exceptions from the worker processes are passed back to the main process using a shared Queue. The main process goes through the Queue to merge the output from all workers and if necessary, terminate all processes when an exception is thrown by a worker. 

I tested that this approach gives us the same output as our current approach in `main`, and it does. Meren Lab members can find the relevant details for that [here](https://github.com/merenlab/lab-fiesta/issues/1223).


meren:
Excellent! :) Thank you, Iva!

PR #1607, opened 2020-12-24
Metabolism code now works with KEGG snapshot by default

ivagljiva:
This PR addresses the issue we've been having dealing with frequent KEGG updates (which is that each time you run `anvi-setup-kegg-kofams`, you get new data from KEGG and need to re-annotate all your databases to make them compatible with the new version before metabolism estimation will work). Now, by default, `anvi-setup-kegg-kofams` instead pulls a specific version of the KEGG data directory from figshare and sets that up on the user's computer. What this means is:
- the default KEGG data is now tied to an anvi'o release (the first one will be `v7`'s KEGG data)
- people will not always have the most up-to-date KEGG data available to use when annotating with KOfams/estimating metabolism. This will hopefully not a problem for most.
- but this will eliminate most problems with reproducibility and data sharing. yay us.

The way I have implemented this is that it is essentially a special case of the `--kegg-archive` option (which still exists, by the way) - we download a specific archive and set that up rather than setting up an archive already on the user's computer.

I did not eliminate the (previously default) option to download data directly from KEGG, so if people want a more updated version of KEGG data they can get it by running `anvi-setup-kegg-kofams -D`, at the risk of having later trouble with sharing said data. I have described the possible downstream version issues and ways to share non-default KEGG data in the help pages. 

**How to update the default KEGG archive, for future versions of anvi'o**
Next time we want to make a release, we should do the following to make sure the default KEGG data is as up-to-date as possible:

1. Run `anvi-setup-kegg-kofams -D --kegg-data-dir ./KEGG` to download the latest KEGG/KOfam data and set it up in an anvi'o-compatible KEGG data directory in your current working directory.
2. Run `anvi-db-info ./KEGG/MODULES.db` to get the hash value of the contents of the MODULES database.
3. Run `tar -czvf KEGG_build_YYYY-MM-DD_HASH.tar.gz ./KEGG` to archive the directory, replacing YYYY-MM-DD with the current date and replacing HASH with the MODULES.db hash value obtained in step 2.
4. Upload the `.tar.gz` archive to figshare and get the download link.
5. In `kegg.py`, replace the link in `self.default_kegg_data_url = "https://ndownloader.figshare.com/files/25878342"` with the download link for the new KEGG archive.
6. Test, merge, document what you did, etc and then go enjoy life.

**Details of `v7` default KEGG data**
I set this data up yesterday (Dec 23, 2020), and unless we want to update it one more time before the release, this is what it is going to be for everyone using `v7`.

Its hash value is `45b7cc2e4fdc` and its DOI is `https://doi.org/10.6084/m9.figshare.13224395.v1`.

Here are some relevant details about the MODULES.db, provided by the setup script:
```
Bad line splitting (usually due to rogue or missing spaces) : 40
Bad KEGG code format (usually not correctable) : 0
Modules database .............................: A new database, ./KEGG/MODULES.db, has been created.
Number of KEGG modules .......................: 443
Number of entries ............................: 13,720
Number of parsing errors (corrected) .........: 40
Number of parsing errors (uncorrected) .......: 0
```

I think that is everything. 

🎄 🤶 🎁 🎄


meren:
This is excellent, Iva. I really like your solution.

One minor point regarding this:

> In kegg.py, replace the link in self.default_kegg_data_url = "https://ndownloader.figshare.com/files/25878342" with the download link for the new KEGG archive.

I wonder if we could have a solution like the one in `anvio/data/misc/GTDB-RELEASES.yaml`, which keeps track of what is available, where is the exact file path, etc. Which enables a solution like this (so earlier versions can also be downloaded and set up on the latest codebase as we would be keeping track of everything in a structured fashion (and if `MODULES.db`s are no longer up-to-date for a given version, migration scripts would take care of it):

```python
(...)

        self.target_database_release = database_release or 'v95.0'

        self.target_database_releases_yaml = os.path.join(os.path.dirname(anvio.__file__), 'data/misc/GTDB-RELEASES.yaml')
        self.target_database_releases = utils.get_yaml_as_dict(self.target_database_releases_yaml)

        if self.target_database_release not in self.target_database_releases:
            raise ConfigError(f"Anvi'o doesn't know a release called {self.target_database_release}. It knows about "
                              f"{', '.join(self.target_database_releases)}, though.")

        self.target_database = self.target_database_releases[self.target_database_release]

(...)
```

Thank you very much for documenting everything meticulously.

ivagljiva:
Okay @meren, this should now be good to go :) 

The yaml solution is now implemented and tested, and as a bonus I also fixed a strange bug that was happening whenever the user provides a directory name that doesn't look like a path to `--kegg-data-dir`. I was getting an error like this:
```
$ anvi-setup-kegg-kofams --kegg-data-dir KEGG_TEST_OLD --kegg-snapshot v2020-04-27

File/Path Error: No output directory path is declared...
```
whenever I tried to pass a folder in the current directory, but the error did not happen if there were slashes in the directory name (as in `--kegg-data-dir KEGG_TEST_OLD/`). A small annoyance, but it doesn't happen anymore because now we add a '/' to the end of the directory name if there isn't already one. 

One last issue that I couldn't figure out: for some reason, the progress bars do not show up during 'setup from archive'. There are `self.progress` lines in the `setup_from_archive()` function and I could find no reason why they should not be executed, yet the progress info does not appear while the program is running. :/ It is a minor problem, but if you have any insight into why this could be happening I would really appreciate it.

meren:
Yay :))) This is great, @ivagljiva! I am merging this now, and it will be in the Docker image I will build soon for testing. Thank you very much for your clean and well-documented code.

> One last issue that I couldn't figure out: for some reason, the progress bars do not show up during 'setup from archive'. 

This is interesting. I will look into this after the merge :)

PR #1608, opened 2020-12-25
A script to fix INDELs associated with homopolymers using a reference

meren:
This is to help with genomes generated by ONT (i.e., minION) long-reads, which often result in a tremendous number of INDEL errors largely due to homopolymers.

The script aims to ameliorate these kinds of errors when there is a high-quality reference genome.

The PR also includes docs to clarify its use.


meren:
Anvi'o says "3.6 roentgen, not great, not terrible".

This is not the future for correcting long-read garbage.

PR #1611, opened 2020-12-29
Issue 1588: fix --cluster-contigs param in snakemake

### LINKED ISSUE #1588: [BUG] Cannot set --cluster-contigs to false for anvi-profile in metagenomics workflow

I am trying to produce single profiles without clustering. I have set the `--cluster-contigs` flag to `false` in my config file for the metagenomics workflow, but this flag is still being added to the `anvi-profile` rule. 

## anvi'o version

```
Anvi'o version ...............................: esther (v6.2-master)
Profile DB version ...........................: 35
Contigs DB version ...........................: 20
Pan DB version ...............................: 14
Genome data storage version ..................: 7
Auxiliary data storage version ...............: 2
Structure DB version .........................: 2
Kegg Modules DB version ......................: 2
tRNA-seq DB version ..........................: 1
```

## System info

Midway

See screenshots below:
image
image

I believe the issue is in the `get_cluster_contigs_param` function in the `anvio/anvio/workflows/metagenomics/Snakefile` file. The conditional statement will be false if you set `--cluster-contigs` to `false`, so the else clause always runs, and that clause uses the `--cluster-contigs` flag if you are working with single profiles (I am).

ivagljiva:
This PR resolves #1588, in which 1) we could not set a value of `false` for the `--cluster-contigs` param in `anvi-profile`, and 2) the default behavior was opposite that of the recommendation (ie, clustering was run for single profiles by default, but not for profiles that would need to be merged, which is not what `anvi-profile -h` recommends as @ekiefl pointed out).

The solution to 1) was to accept parameters that are not `None`, and the solution to 2) was to flip the logic so that by default clustering is not run for single profiles, but it was run for samples in groups with more than 1 member. The warning message that prints when you explicitly set `--cluster-contigs` to something has been updated to reflect the new defaults.

**How did I test this?**
The datapack from the [snakemake workflows tutorial](https://merenlab.org/2018/07/09/anvio-snakemake-workflows/#fastatxt) was perfect for this, as it had a simple `fasta.txt` with two references, one of which has two samples mapping to it and one of which has just one sample mapping to it. So right there we have a test case for both merged and single profiles.

I downloaded the datapack, set it up, got a default config file, and did a dry run to see the command lines for `anvi-profile`:
```
anvi-run-workflow -w metagenomics -c config.json -A -n
```
Which, as expected, yielded the correct defaults of `--cluster-contigs` for profiles that would be merged (G02 group) and no clustering for the single profile (G01 group):
image

I then made another config in which I explicitly set the value of `--cluster-contigs` to `false` (this was not working for me before), and this time in the dry run all `anvi-profile` jobs skipped the clustering:
```
anvi-run-workflow -w metagenomics -c config_set_cluster_contigs.json -A -n
```

image
(I am only showing the first two samples in this screenshot)

Setting the parameter value to `true` provides the opposite behavior, as expected.

### DIFF
--- a/anvio/workflows/metagenomics/Snakefile
+++ b/anvio/workflows/metagenomics/Snakefile
@@ -783,7 +783,7 @@ if sample_name != M.default_config['anvi_profile']['--sample-name'] and sample_n
 
 def get_cluster_contigs_param(wildcards):
     """ helper function to sort out whether to cluster contigs for a single profile database"""
-    if M.get_param_value_from_config(['anvi_profile', '--cluster-contigs']):
+    if M.get_param_value_from_config(['anvi_profile', '--cluster-contigs'] is not None):
         run.warning('You chose to set the value for --cluster-contigs as %s. You can do that if you '
                     'choose to, but just so you know, if you don\'t provide a value then '
                     'the workflow would automatically cluster contigs for profile databases '
--- a/anvio/workflows/metagenomics/Snakefile
+++ b/anvio/workflows/metagenomics/Snakefile
@@ -783,7 +783,7 @@ if sample_name != M.default_config['anvi_profile']['--sample-name'] and sample_n
 
 def get_cluster_contigs_param(wildcards):
     """ helper function to sort out whether to cluster contigs for a single profile database"""
-    if M.get_param_value_from_config(['anvi_profile', '--cluster-contigs'] is not None):
+    if M.get_param_value_from_config(['anvi_profile', '--cluster-contigs']) is not None:
         run.warning('You chose to set the value for --cluster-contigs as %s. You can do that if you '
                     'choose to, but just so you know, if you don\'t provide a value then '
                     'the workflow would automatically cluster contigs for profile databases '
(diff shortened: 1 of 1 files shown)


anvio/workflows/metagenomics/Snakefile
```diff
@@ -783,7 +783,7 @@ if sample_name != M.default_config['anvi_profile']['--sample-name'] and sample_n
 
 def get_cluster_contigs_param(wildcards):
     """ helper function to sort out whether to cluster contigs for a single profile database"""
-    if M.get_param_value_from_config(['anvi_profile', '--cluster-contigs'] is not None):
+    if M.get_param_value_from_config(['anvi_profile', '--cluster-contigs']) is not None:
```
meren, line 786: 😂

meren:
Thank you SO MUCH for addressing this, @ivagljiva, and for your input, @ekiefl :)

It looks to me it is good to go.

anvio/workflows/metagenomics/Snakefile
ivagljiva: yeah, it is embarrassing 😂

meren: No bug that is caught is embarrassing! Embarrassing bugs are those that linger for ages (at which I excel).

Issue #1562, opened 2020-11-17
[FEATURE] Sort HMM layers displayed in the interactive interface


meren:
## The need

Currently there is no alphabetical order to these additional layers, but it would have been great if there was:

![image](https://user-images.githubusercontent.com/197307/99424200-0a385900-28c7-11eb-83da-2e8119c3c9a1.png)

## The solution

Probably change something in the `interactive.py` where additional data are appended to the final views data.

## Beneficiaries

Anyone who may enjoy some additional sanity in their life.

meren:
Plus, perhaps we shouldn't show layers that have no data?

It really is becoming very distracting there.

![image](https://user-images.githubusercontent.com/197307/103262166-aeeb9300-4969-11eb-8fb4-994c924912c5.png)

meren:
#1612 addressed this.

PR #1612, opened 2020-12-30
Blank interactive layers refactor

meren:
This PR addresses the issue #1562 and more, including:

* If a given HMM and/or item additional data layer does not have any data points for a given set of splits to be displayed, remove them from anvi'o displays initialized by `anvi-interactive` and `anvi-refine`, but let the user overwrite that behavior via the flag `--show-all-layers`.
* Prune items additional data dict given the splits of interest prior to initializing the interface.

I imported the following items additional data to infant gut dataset prior to testing this:

|name|TEST_KEY_SHOW|TEST_KEY_NOSHOW|TEST_INT_SHOW|TEST_INT_NOSHOW|TEST_FLOAT_SHOW|TEST_FLOAT_NO_SHOW|
|:--|:--:|:--:|:--:|:--:|:--:|:--:|
|IGD_000000002399_split_00001|TRUE||1|0|0.5|0.0|
|IGD_000000001622_split_00001|FALSE||2|0|0.1|0.0|

This is how the IGD looks before the changes in this PR:

image

And this is how it looks now:

image

Mini test before:

image

Mini test now:

image

And this was shown in the terminal:

```
WARNING
Even though the following layer names were associated with your data, they will
not be shown in your interactive display since none of the splits you are in
terested at this point had any hits in those layers: "Ribosomal_RNA_12S,
Ribosomal_RNA_16S, Ribosomal_RNA_18S, Ribosomal_RNA_23S, Ribosomal_RNA_28S,
Ribosomal_RNA_5S". If you would like every layer to be shown in the interface
even when there are nohits, please include the flag `--show-all-layers` in your
command.
```

I think this is good to go.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -552,6 +552,14 @@ D = {
                      "searches that were performed against non-singlecopy gene HMM profiles into "
                      "their own layer. Please see the documentation for details."}
                 ),
+    'show-all-layers': (
+            ['--show-all-layers'],
+            {'default': False,
+             'action': 'store_true',
+             'help': "When declared, this flag tells the interface to show every additional layer even if "
+                     "there are no hits. By default, anvi'o doesn't show layers if there are no hits for "
+                     "any of your items."}
+                ),
     'taxonomic-level': (
             ['--taxonomic-level'],
             {'default': 't_genus',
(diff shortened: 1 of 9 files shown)


meren:
(I also updated the default state to make the default view a bit more pleasant)

PR #1615, opened 2021-01-01
tRNA-seq upgrade

semiller10:
This PR contains the changes made since tRNA-seq capabilities were first added to the master branch.

`anvi-convert-trnaseq-database` lies at the center of the new work. It takes a set of anvi'o trnaseq databases produced by `anvi-trnaseq` for each sample or split in an experiment and finds tRNA seed sequences, specific and nonspecific coverages, hierarchical clusters, and various other information, producing conventional contigs and profile databases that can be displayed and further manipulated by anvi'o. This program runs in a matter of minutes -- much faster than an average `anvi-trnaseq` run -- allowing the user to quickly refine the output to suit their needs. For instance, the number of tRNA seed sequences will not be known a priori by the user, except by a rough estimate of the diversity of tRNA genes in the samples, and so `--max-reported-trna-seeds` can be varied to understand how many significantly covered tRNAs are present.

Significant revisions were made to `trnaidentifier`, including abstraction of feature profiling through the creation of a `Profiler` class in addition to a `Profile` object and abstraction of feature parameterization through the creation of a `TRNAFeatureParameterizer` class. The handling of sequences with truncated feature profiles in the `trnaidentifier` and `trnaseq` modules was improved, as these sequences were discarded as potential chimeric artifacts, when in reality they also contained a large number of valid 3' tRNA fragments and sequences with deletions that can now be recovered. Related to this, tRNAs with deletions are now sought in different pools of sequences through a complicated procedure. This is one of the newest parts of the pipeline, as deletions are less frequent than substitutions at modification sites, and there is no easy way of identifying them. However, deletions provide complementary information to substitutions and coverage drop-offs in assigning modification sites.

Numerous new user options, workflow rules, and artifact records support these expanded capabilities.


meren:
@semiller10, when I type this, I get the error below:

```
anvi-trnaseq -h
Error: The default clustering configuration file for trnaseq runs, 'cov',
       is missing from data/clusterconfigs dir! I don't know how this happened,
       but I can't fix this! Anvi'o needs an adult :(
```

Perhaps you didn't commit the `.ini` file?

Plus, I don't see a `./run_mini_test.sh` equivalent in this PR. Is it coming, too?

anvio/trnaseq.py
```diff
@@ -2068,240 +2945,1932 @@ def write_trimmed_supplement(self):
...
+        'specific_nt_covs_dict',
+        'nonspecific_nt_covs_dict',
+        'specific_covs',
+        'nonspecific_covs',
+        'specific_del_covs',
+        'nonspecific_del_covs',
+        'norm_seq_summaries'
+    )
+
+    def __init__(self):
```
meren, line 3010: variables defined here are not introduced as members of the class with `self.`, so these lines do not contribute to anything when there is an instance as far as Python is concerned :)

anvio/trnaseq.py
```diff
@@ -2068,240 +2944,1932 @@ def write_trimmed_supplement(self):
...
+                ('sample_mean_specific_cov_dict', 'sample_mean_nonspecific_cov_dict', 'mean_coverage'),
+                ('sample_std_specific_cov_dict', 'sample_std_nonspecific_cov_dict', 'std_coverage'),
+                ('sample_specific_abundances_dict', 'sample_nonspecific_abundances_dict', 'abundance'),
+                ('sample_specific_relative_abundances_dict', 'sample_nonspecific_relative_abundances_dict', 'relative_abundance'),
+                ('sample_specific_detection_dict', 'sample_nonspecific_detection_dict', 'detection'),
+                ('sample_specific_max_normalized_ratio_dict', 'sample_nonspecific_max_normalized_ratio_dict', 'max_normalized_ratio'),
+                ('sample_mean_Q2Q3_specific_cov_dict', 'sample_mean_Q2Q3_nonspecific_cov_dict', 'mean_coverage_Q2Q3')]:
+                data_dict = self.get_combined_data_dict(specific_attr, nonspecific_attr)
+                self.create_combined_contigs_and_splits_tables(profile_db_path, table_basename, data_dict)
+            variability_data_dict = self.get_combined_data_dict('sample_variability_dict', 'sample_variability_dict')
```
meren, line 4738: you're not doing anything with this variable although it looks important :p is it a mistake or intentional?

meren:
This morning I've been going through the code, and this is definitely a tremendous amount of work, @semiller10.

Except some minor formatting issues (which is mostly our fault for not having a good coding style document), my superficial browsing of the code did not reveal any major problems. I am very thankful for your diligence and efforts. The documentation in the code is quite useful as well. Thank you very much for your attention to keep your code readable and understandable. But still, digesting these classes and functions for sure will take a long time for anyone who is interested in doing it.

I think one of the two biggest shortcoming at this stage is the lack of a small test that is  more or less equivalent to the mini test (if you decide to add tests, please don't forget **not to** commit flat FASTA or FASTQ files with many lines). The other shortcoming is the online documentation, but I think it is not that critical and we can recover from that over time while we are also working on publishing this grandiose work.

Please let me know if it is conceivable to use the mini test as a template to update this PR soon. If it is, I will wait for it. If not, I still suggest we merge this PR to master and include it in `v7` without the tests, which we can develop later. So in summary, here are the things I can think of:

* Please take a look at the changes I've made and make sure I didn't screw anything up
* Please add the `cov` file under `data/clusterconfigs/trnaseq/`.
* I know this code still needs tRNA taxonomy to be included in its contigs databases. I am not sure if we can pull it off before the release in a way that will not explode later.
* Let me know if I should wait for tests or should I plan to merge it after you are done with the first and second items in this list.

Best wishes,

meren:
ONE LAST THING!

Is `anvi-agglomerate` still releveant? It seems the class it uses is now used from within `trnaseq.py`.

PR #1616, opened 2021-01-01
Neat displays of output files during testing

meren:
This PR enhances `SHOW_FILE` function we use in our test routines to produce neat outputs like this one:

image

With the `--max-width` option that is set to `120` by default, longer outputs automatically cut like this to not cause a mess in the terminal:

image


meren:
Tested with everything, will test with Docker before release.

PR #1617, opened 2021-01-02
tRNA-seq upgrade

semiller10:
This PR contains the changes made since tRNA-seq capabilities were first added to the master branch.

`anvi-convert-trnaseq-database` lies at the center of the new work. It takes a set of anvi'o trnaseq databases produced by `anvi-trnaseq` for each sample or split in an experiment and finds tRNA seed sequences, specific and nonspecific coverages, hierarchical clusters, and various other information, producing conventional contigs and profile databases that can be displayed and further manipulated by anvi'o. This program runs in a matter of minutes -- much faster than an average `anvi-trnaseq` run -- allowing the user to quickly refine the output to suit their needs. For instance, the number of tRNA seed sequences will not be known a priori by the user, except by a rough estimate of the diversity of tRNA genes in the samples, and so `--max-reported-trna-seeds` can be varied to understand how many significantly covered tRNAs are present.

Significant revisions were made to `trnaidentifier`, including abstraction of feature profiling through the creation of a `Profiler` class in addition to a `Profile` object and abstraction of feature parameterization through the creation of a `TRNAFeatureParameterizer` class. The handling of sequences with truncated feature profiles in the `trnaidentifier` and `trnaseq` modules was improved, as these sequences were discarded as potential chimeric artifacts, when in reality they also contained a large number of valid 3' tRNA fragments and sequences with deletions that can now be recovered. Related to this, tRNAs with deletions are now sought in different pools of sequences through a complicated procedure. This is one of the newest parts of the pipeline, as deletions are less frequent than substitutions at modification sites, and there is no easy way of identifying them. However, deletions provide complementary information to substitutions and coverage drop-offs in assigning modification sites.

Numerous new user options, workflow rules, and artifact records support these expanded capabilities.


meren:
Thank you, @semiller10. This is now merged.

PR #1627, opened 2021-01-07
tRNA-seq variant of contigs database

semiller10:
This PR adds variant functionality to `anvi-interactive` in displaying tRNA-seq data. This is the first use of database variant information. Contigs databases containing tRNA seeds (generated by `anvi-convert-trnaseq-database`) now have a `db_variant` attribute set to `trnaseq`. `anvi-interactive` detects this and calls a function to generate items additional layers showing taxonomic assignments for each rank. These will only be displayed if `anvi-run-trna-taxonomy` has been run on the `contigs` database.

PR #1631, opened 2021-01-07
Merge tRNA-seq variant changes into master

semiller10:
This is another merge after PR [1627](https://github.com/merenlab/anvio/pull/1627) to get the tRNA-seq-specific changes into `master`.


semiller10:
@meren If I make a branch off another branch like this and the parent branch is merged into `master` before the child branch, is this the proper way to merge the child branch?

Issue #1644, opened 2021-01-18
[BUG] killing anvi-run-kegg-kofams prematurely leads to "this contigs database has already been annotated" error when re-running anvi-run-kegg-kofams


meren:
Anvi'o should update the `self` table as the very last step of running KOfams and only after a successful completion of the run.

PR #1645, opened 2021-01-18
Fix KEGG KOfams annotation bug #1644 

ivagljiva:
This PR addresses #1644 by splitting the one function that checked and set the modules db hash into two separate functions: the sanity check is called before `hmmsearch` starts running (as before), and now the modules db hash is set after it is done running.


meren:
Thank you, @ivagljiva! :)

PR #1649, opened 2021-01-25
anvi-display-functions

meren:
This PR introduces a new program that displays the distribution of unique functions (or accession ids) provided by any annotation source across genomes that are defined in anvi'o ecosystem through internal genomes, external genomes, or genomes storage databases (phew, it *does* a lot).

For most comprehensive information, please see the [program doc](https://github.com/merenlab/anvio/compare/anvi-display-functions?expand=1#diff-ea1a16c05611afa22d4643b21f585994c69d97ad318c90dbb4a8f1defa7a32bc).

What is more important is that `anvi-display-functions` relies on a new class that can aggregate functions from many sources, which I believe is the missing integration across different anvi'o databases we need for more streamlined implementation of functional enrichment.


anvio/docs/programs/anvi-display-functions.md
```diff
@@ -0,0 +1,95 @@
...
+
+In some cases a gene may be annotated with multiple functions. This is a decision often made at the function annotation tool level. For instance %(anvi-run-ncbi-cogs)s may yield two COG annotations for a single gene because the significance score for both hits may exceed the default cutoff. While this can be useful in %(anvi-summarize)s output where things should be most comprehensive, having some genes annotated with multiple functions and others with one function may over-split them (since in this scenario a gene with COGXXX and COGXXX;COGYYY would end up in different bins). Thus, %(anvi-display-functions)s will will use the best hit for any gene that has multiple hits. But this behavior can be turned off the following way:
+
+{{ codestart }}
+anvi-display-functions -e %(external-genomes)s \
+                       --annotation-source KOfam \
+                       --profile-db KOFAM-PROFILE.db \
+                       --aggregate-using-best-hit
+{{ codestop }}
+
```
ivagljiva, line 47: I am a bit confused here :) It sounds like `anvi-display-functions` aggregates using best hit by default, and that there is a flag to turn this off, but the flag `--aggregate-using-best-hit` sounds like it is turning this behavior on instead. What is the default?

Also I think the heading for this section is supposed to be something different :)

anvio/genomedescriptions.py
```diff
@@ -1026,3 +1042,285 @@ def sanity_check(self):
...
+    def check_layer_names(self, layer_names=[]):
+        if not isinstance(layer_names, list):
+            raise ConfigError("`layer_names` must be of type list :/")
+
+        already_in_the_dict = [g for g in layer_names if g in self.layer_names_considered]
+        if len(already_in_the_dict):
+            raise ConfigError(f"Anvi'o is not happy because there are some genome or metagenome names that are not unique "
+                              f"across all input databases :/ Here is an example: {already_in_the_dict[0]}.")
+        else:
+            # you good fam
```
ivagljiva, line 1265: 🤣

anvio/docs/programs/anvi-display-functions.md
meren: I agree and I fixed that later. Our flag is now `--aggregate-using-all-hits` (by default it uses the best)

meren: Ah, great, I forgot to update the md file, of course :)

anvio/genomedescriptions.py
```diff
@@ -1051,6 +1051,30 @@ class AggregateFunctions:
...
+    group stats output file for functioanl enrichment analysis:
+
+        >>> import argparse
+        >>> import anvio.genomedescriptions as g
+        >>> args = argparse.Namespace(external_genomes=external_genmes_path, annotation_source='KOfam')
+        >>> groups = {'adoles': ['B_adolescentis', 'B_adolescentis_1_11', 'B_adolescentis_22L', 'B_adolescentis_6', 'B_adolescentis_ATCC_15703'],
+                      'lactis': ['B_animalis', 'B_lactis_AD011', 'B_lactis_ATCC_27673', 'B_lactis_B420', 'B_lactis_BB_12', 'B_lactis_BF052'],
+                      'longum': ['B_longum', 'B_longum_AH1206', 'B_longum_BBMN68', 'B_longum_BORI', 'B_longum_CCUG30698', 'B_longum_GT15']}
+        >>> facc = g.AggregateFunctions(args, layer_groups=groups)
+        >>> facc.report_functions_per_group_stats(output_file_path)
```
meren, line 1066: Hey @ivagljiva :) There is now a function that reports a file that is somewhat similar to the file format that goes to the enrichment analysis as input, but there are missing columns and I believe the column names will need to be updated.

If you send me an example file and I can fix the function to match to that.

ivagljiva: I am on it :)

meren:
the new script `anvi-script-gen-functions-per-group-stats-output` is meant to test and see how the output file looks like (and demonstrate how programmatically this output file could be generated from elsewhere).

It is not yet fully compatible with `anvi-script-enrichment-stats`, but I think it should be minor changes from here.

PR #1657, opened 2021-01-28
1650 grids and shades

matthewlawrenceklein:
This PR seeks to add additional flexibility and functionality to selections in the interactive interface. Users are now able to independently enable/disable both grids and shade for their selections, as well as adjust the fill opacity for their shade. 
Screen Shot 2021-01-28 at 9 46 32 AM
The default selections setting is currently shade with an opacity of '0.1', consistent with the previous default.  
Screen Shot 2021-01-28 at 9 46 22 AM
These additional user-adjustable settings are stored in / retrieved from state. Fun!


meren:
Thanks for this, @matthewlawrenceklein! Looks great! :)

A suggestion, a bug, and an idea:

* Suggestion: I think it would be more user friendly if shade opacity input box appeared next to 'Show shade' the way the color picker and width appears next to Show grids once it is activated.
* Bug: When you switch to the rectengular view from circular view, shade no longer works :)
* Idea: Now shade 'highlights' the selection. Your example above gave me this idea: what if there was one more option to shade to 'invert' it? For instance, if you select shade, it covers what you have selected. When you invert, it covers everything else that is not selected. I'm sure it would be a great feature! Do you think this is doable?


Thank you!

meren:
> Idea: Now shade 'highlights' the selection. Your example above gave me this idea: what if there was one more option to shade to 'invert' it? For instance, if you select shade, it covers what you have selected. When you invert, it covers everything else that is not selected. I'm sure it would be a great feature! Do you think this is doable?

Of course in this case we would need to ask for a 'color' for the inverted shades since the current one follows the color of the bin, but the other option will require shade color to be the same for every portion.

matthewlawrenceklein:
@meren  - To clarify on the inverted-shade idea : When the user selects 'invert shade' (and sets an inverted color), anvio will shade all items not present in **any** bin? Should the 'invert shade' behavior also turn off non-inverted highlighting for any active bins?

meren:
> When the user selects 'invert shade' (and sets an inverted color), anvio will shade all items not present in any bin? 

Yes :) That's what I was thinking.

> Should the 'invert shade' behavior also turn off non-inverted highlighting for any active bins?

Yes, it should, in my opinion.

Thank you, @matthewlawrenceklein!

meren:
I presume it will not disrupt the smooth operation of databases with states that do not include these variables and they will continue working like the way they were working before?

If that is the case, please feel free to merge it whenever you wish, @matthewlawrenceklein!

meren:
goes into the main branch!

Thank you very much, @matthewlawrenceklein!

Issue #1658, opened 2021-01-29
[BUG] `anvi-estimate-metabolism --matrix-format` produces incorrect output for metagenome input


ivagljiva:
## Short description of the problem

Metagenome mode estimates metabolism within each contig separately, but the --matrix-format output is aggregating the estimation results for each contig and reporting them for the metagenome as a whole. Which is very wrong. So I am very sad.

## anvi'o version

`v7` and `v7-dev` 

## Detailed description of the issue

You can run `anvi-estimate-metabolism --matrix-format -M METAGENOMES.TXT` to get three matrix files - one for module completeness, one (binary) for module presence-absence, and one for # of KOfam hits. 

These matrices _should_ have one column for every contig in every metagenome, but instead they have one column for every metagenome. See the following example for module completeness:
```
module	meta_1	meta_2
M00001	0.90	11.01
M00002	1.00	11.77
M00003	0.75	10.98
M00004	1.00	9.36
M00005	1.00	17.00
M00006	1.00	11.8
```
In this example, meta_1 is not actually a metagenome, it is an isolate that I used for quick testing. But meta_2 is a real metagenome, and you can immediately see the problem in that column: completeness scores are larger than 1 (which should never happen), because the module's completeness in each contig is being added together. Not good. The other two matrix outputs suffer from a similar problem.

ivagljiva:
Solved in #1659

PR #1659, opened 2021-01-30
A lot of improvements to anvi-estimate-metabolism

ivagljiva:
This PR started due to a feature request for module-specific KO hit matrices, but expanded to a lot of needed improvements to the metabolism codebase, especially for `--matrix-format`. 

It includes the following changes:

## New feature: module specific KO matrices
Now multi-mode `--matrix-format` has an additional parameter: `--module-specific-matrices`. You can pass a comma-separated list of KEGG module numbers to this flag, and then your matrix output will include KO hit matrices for each module in the list. 

For example, 
```
anvi-estimate-metabolism -e input_txt_files/external_genomes.txt \
                                              --matrix-format \
                                              --module-specific-matrices M00001,M00009 \
                                              -O external_genomes
``` 

will produce the output files `external_genomes-M00001_ko_hits-MATRIX.txt` and `external_genomes-M00009_ko_hits-MATRIX.txt` (in addition to the typical output matrices). Here is what `external_genomes-M00001_ko_hits-MATRIX.txt` looks like:
image
So for every KO in the module definition, there will be a row in this matrix indicating how many hits to that KO were annotated in each input sample.

## A new set of functions for matrix generation
Making matrices for multi mode used to be a huge mess, involving like 4-5 functions and hard-to-read pandas dataframe manipulation. It is much simpler now. We have a generic function to write to a matrix, `KeggMetabolismEstimatorMulti.write_stat_to_matrix()`, and a driver function to call this one for each matrix output file, `KeggMetabolismEstimatorMulti.store_metabolism_superdict_multi_matrix_format()`. The code is much more readable and easier to control, as well (IMO), though maybe not as fancy as using pandas. Sorry not sorry. 🐼 

## Matrix output is just better in general
Here is the improvement list:
- now the full completeness score for each module is reported (before, we were rounding to 2 decimal places)
- you can now use `--include-zeros` for matrix output, in which case matrices will include rows of all 0s (this does not affect module-specific KO matrices, which always show all KOs in the module even if some have all 0s)
- The bug for metagenome mode matrix output, detailed in #1658, is now fixed. That means every contig in a metagenome gets its own column in the matrix (instead of one column for the metagenome overall with the wrong data)

## Metabolism estimation function doesn't return huge dictionaries anymore (by default)
`KeggMetabolismEstimator.estimate_metabolism()` used to return huge superdictionaries of metabolism estimation, even after PR #1543 when we changed the code to append this data to long-format files as we got it. Why? Because `--matrix-format` needed them. Well, no longer. 

Now, by default, this function doesn't return anything. You can ask it to return the entire superdictionaries (the previous behavior) if you need it to, using the `return_superdicts=True` parameter. For the `--matrix-format` case, we ask this function to return only a subset of the superdictionaries which we need for matrix generation (using the `return_subset_for_matrix_format=True` parameter). This relies on a new function to make those subsets, which is `KeggMetabolismEstimator.generate_subsets_for_matrix_format()`

I have not tested this on large datasets, but I am 99.35% certain that the memory consumption will go down, resulting in faster runtime and happier clusters.

## A new set of functions to standardize metadata access
For modules, you can now get a metadata dictionary using `KeggContext.get_module_metadata_dictionary()` and for KOs, you can use `KeggContext.get_ko_metadata_dictionary()`. These are used by functions for both long-format output generation and matrix output generation, so that we can access metadata on the fly rather than passing around large dictionaries including this information.

## REALLY IMPORTANT: `modules` mode output now has warnings for missing KOfam profiles
So, there is this problem where sometimes we don't have (or don't use) a KOfam profile for a KO. Usually this is because it doesn't have a bitscore in the ko_list.txt file from KEGG, which means we can't parse out the weak hits, so we don't even try to annotate it. The downstream problem is that any module with this KO can never have a 1.0 completeness score because we never annotate that KO. This was a problem that occurred silently before. 

But now, `modules` mode output includes a "warnings" column (by default) that tells you if the module contains a KO that we cannot annotate. So at least people know about this issue now. 

If the module has a '--' step (no associated KO), we also put a warning that these steps are assumed to be incomplete.

(If you are a Meren Lab member and want more info on this problem, I investigated it [here](https://github.com/merenlab/lab-fiesta/issues/955#issuecomment-743827152).)

## Bonus: re-organization
This doesn't matter to anyone except for me right now, but I've organized functions according to their larger purpose and added comments to section the code. Makes it easier to find the functions you are looking for.


meren:
A TON of improvements, indeed, and usual given the @ivagljiva style, complete with the updates in docs :) Thank you very much!

Looking at the `external_genomes-M00001_ko_hits-MATRIX.txt` output, I'm thinking...

Here we show all KOfams as rows, regardless of which alternative in module decomposition we have used to estimate completion. I can't see any other way of doing it, so I think it is fine. But the fact that there is no information regarding which KOfams have `OR` relationship among them will make it very difficult to interpret this output for biologists.

Here is a more concrete example using your example screenshot. Looking only this output, one can't see whether `K00845` and `K00886` should be considered 'equivalent' to one another (in which case all three genomes have that step covered for this module), or should be treated as KOfams for distinct steps (in which case one of the genomes is missing that step, and two others missing a step covered by the other).

image

Although it becomes clear when you go back to the module description that in fact they indeed are equivalent:

image

And here I am wondering if there is a way to retain that information in the output, somehow. The simplest way I can think of wold be to add comment lines like this:

```
KO      genome_1       genome_2     (...)
# step_01
K00845     ..            ..
K00886     ..            ..
K00844     ..            ..
(...)
# step_02
K01810     ..            ..
K13810     ..            ..
(...)
# step_03
(...)
```

I know we are both trying to simplify the information and keep its intricacies intact, which typically is a very challenging task, but anyone who will have to study these output files will have to do it manually otherwise.

meren:
Matrix mode is a game changer. Look how clear it is:

```
KO	W0	W1	W48
# (K00844,K12407,K00845,K00886,K08074,K00918)
K00844	0	0	0
K12407	0	0	0
K00845	2	2	2
K00886	0	0	0
K08074	0	0	0
K00918	0	0	0
# (K01810,K06859,K13810,K15916)
K01810	1	1	1
K06859	0	0	0
K13810	0	0	0
K15916	0	0	0
# (K00850,K16370,K21071,K00918)
K00850	0	0	0
K16370	0	0	0
K21071	2	2	2
K00918	0	0	0
# (K01623,K01624,K11645,K16305,K16306)
K01623	0	0	0
K01624	1	1	1
K11645	0	0	0
K16305	0	0	0
K16306	0	0	0
# K01803
K01803	1	1	1
# ((K00134,K00150) K00927,K11389)
K00134	1	1	1
K00150	0	0	0
K00927	1	1	1
K11389	0	0	0
# (K01834,K15633,K15634,K15635)
K01834	1	1	1
K15633	1	1	1
K15634	0	0	0
K15635	0	0	0
# K01689
K01689	2	2	2
# (K00873,K12406)
K00873	1	1	1
K12406	0	0	0
```

THANK YOU!

meren:
I need a refresher fro `-` and `+` to interpret these, but I'm loving them:

```
KO	W0	W1	W48
# (K01647,K05942)
K01647	1	1	1
K05942	0	0	0
# (K01681,K01682)
K01681	0	1	1
K01682	0	0	0
# (K00031,K00030)
K00031	1	1	1
K00030	0	0	0
# (K00164+K00658+K00382,K00174+K00175-K00177-K00176)
K00164	1	1	1
K00658	0	1	1
K00382	1	1	1
K00174	0	0	0
K00175	0	0	0
K00177	0	0	0
K00176	0	0	0
# (K01902+K01903,K01899+K01900,K18118)
K01902	1	1	1
K01903	1	1	1
K01899	0	0	0
K01900	1	1	1
K18118	0	0	0
# (K00234+K00235+K00236+K00237,K00239+K00240+K00241-(K00242,K18859,K18860),K00244+K00245+K00246-K00247)
K00234	0	0	0
K00235	0	0	0
K00236	0	0	0
K00237	0	0	0
K00239	0	1	1
K00240	1	1	1
K00241	1	1	1
K00242	0	0	0
K18859	0	0	0
K18860	0	0	0
K00244	0	0	0
K00245	0	0	0
K00246	0	0	0
K00247	0	0	0
# (K01676,K01679,K01677+K01678)
K01676	1	1	1
K01679	0	0	0
K01677	0	0	0
K01678	0	0	0
# (K00026,K00025,K00024,K00116)
K00026	0	0	0
K00025	0	0	0
K00024	1	1	1
K00116	0	0	0
```

ivagljiva:
Yes :) 

However, I just want to note that these comments aren't always perfectly placed. In particular, modules with repeated KOs can sometimes have issues. See the following example in which K07433 shows up twice and throws off the placement of one step comment:
```
KO	isolate	E_faecalis_6240	test_2
# K00513
K00513	0	0	0
# (K00497,K07433)
K00497	0	0	0
# K07433
K07433	0	0	0
K07433	0	0	0
```
(This is the only instance of this error that I have found, but there may be more.)

ivagljiva:
> I need a refresher fro - and + to interpret these

+ and - connect KOs which are components of a larger enzyme complex. + means that the following component is essential and - means it is not essential. We don't count non-essential components in our completeness scores.

ivagljiva:
I think this is ready to merge. Is that okay with you @meren?

meren:
Yes! Go for it! :)

Issue #1653, opened 2021-01-27
Why can't anvi'o be nice: A feature request


meren:
## The need

SO I AM FINALLY TOO TIRED OF TYPING `-p PROFILE.db -c CONTIGS.db` EVERY TIME I WANT ANVI'O TO DO SOMETHING STUPID. JUST FIGURE IT OUT. MAKE IT HAPPEN. LET ME GET AWAY BY TYPING `anvi-interactive` OR `anvi-display-pan`. WHAT IS THE BIG DEAL EH? WHY CAN'T YOU BE NICE YOU STUPID CALCULATOR :(

## The solution

Well. Our beloved anvi'o knows all of its databases, their types, and the network of association between them thanks to cryptographic hashes. It certainly has the technical capacity to make it happen if it wanted to (at least whenever it can unambiguously resolve database associations).

## Beneficiaries

ME.

And/or anyone who uses anvi'o for anything and/or older than 40 and/or have hands.

ivagljiva:
😂 

seriously, though, this will be next-level stuff. pretty soon anvi'o will be reading everyone's minds and taking over the world

PR #1662, opened 2021-02-02
Auto-fill anvi'o DBs

meren:
This PR addresses the feature request brought to our attention in #1653.

Just go to anywhere on your disk, and type `anvi-interactive` or `anvi-display-pan` without any parameters 😂


ivagljiva:
This is amazing. 😍 And very impressive!

ekiefl:
This is great. I am working on a more generalized framework for anvi'o being able to learn about its DBs. Please wait for my commits before merging.

meren:
I love the way this is getting over-engineered 😂 VERY ANVI'O LIKE :)

Are you also working on including `pan` and `genomestorage`, @ekiefl?

meren:
(I'm adding pan and genomestorage)

ekiefl:
Haha yes I am going to add those right now. This is actually going to be a lot broader than I had initially anticipated. Here is my planned roadmap but looking for guidance.

1. It would be nice for these child classes of DBInfo to know about current versions. Currently that's all housed in `tables/__init__.py`. Should we migrate that info over to `dbinfo.py`?
2. The class in `anvi-db-info` should make use of `DBInfo` either by inheriting or by moving all of its functionality into `DBInfo`.
3. Ideally, I think all utils fns like `get_db_type_and_variant` should be removed and replaced with static methods like `from anvio.dbinfo import DBInfo as dbi; dbi.get_db_type_and_variant(path)`. But given the breadth of code these functions inhabit, it may be more practical that these functions should remain in `utils` and instantiate `DBInfo` for their functionality.
4. I was hoping to create a framework to establish "bundles" of class instances. For example, a profile, contigs, and structure DB could all know about each other and could be checked for hash consistencies. This would help with loads of stuff, but especially when thinking about auto-filling DBs.

meren:
> "This is actually going to be a lot broader than I had initially anticipated"

I want this on a t-shirt with an anvi'o logo behind :)

> It would be nice for these child classes of DBInfo to know about current versions. Currently that's all housed in tables/__init__.py. Should we migrate that info over to dbinfo.py?

Why not keeping those in `tables/__init__.py` but being able to learn the anvi'o version of a given DB from there upon the initialization of the known DBs? The structure there already ready for that kind of access:

``` python
versions_for_db_types = {'contigs': contigs_db_version,
                         'profile': profile_db_version,
                         'genes': genes_db_version,
                         'structure': structure_db_version,
                         'pan': pan_db_version,
                         'genomestorage': genomes_storage_vesion,
                         'auxiliary data for coverages': auxiliary_data_version,
                         'trnaseq': trnaseq_db_version,
                         'config': workflow_config_version,
                         'modules': metabolic_modules_db_version}
```

> The class in anvi-db-info should make use of DBInfo either by inheriting or by moving all of its functionality into DBInfo.

Completely agreed. We can add DB-specific information that is currently done in an adhoc fashion in `anvi-db-info` into `DBInfo` classes, so there would be simpler programmatic access for downstream stuff.

> Ideally, I think all utils fns like get_db_type_and_variant should be removed and replaced with static methods like from anvio.dbinfo import DBInfo as dbi; dbi.get_db_type_and_variant(path). But given the breadth of code these functions inhabit, it may be more practical that these functions should remain in utils and instantiate DBInfo for their functionality.

I agree. Once you're done with your changes we can take another look and see if it makes sense to carry those functions into the new design.

> I was hoping to create a framework to establish "bundles" of class instances. For example, a profile, contigs, and structure DB could all know about each other and could be checked for hash consistencies. This would help with loads of stuff, but especially when thinking about auto-filling DBs.

Indeed. This would enable someone to search for "anvi'o projects" on their disk with accurate associations between each DB, and even the list of all the commands they could run on them and so on :)

ekiefl:
My edits are almost done. The current state of the branch appears stable and `run_all_tests.sh` and `run_pangenome_tests.sh` all run without error. But I have a couple of requests for @ivagljiva, @semiller10, and @meren.

@semiller10 I added a `trnaseq` DB to `dbinfo.py` and I got through `anvi-self-test --suite trnaseq`, but my testing is only as comprehensive as that test. If you can think of any potential snags please let me know.

@ivagljiva I added a `modules` DB to `dbinfo.py` and I got through `anvi-self-test --suite metabolism`, but my testing is only as comprehensive as that test. If you can think of any potential snags please let me know. Additionally, I stumbled upon something confusing. There is a `utils.is_kegg_modules_db` as well as a `utils.is_modules_db`. `utils.is_modules_db` is only ever used in the `anvio/migrations/modules/v1_to_v2.py` script. Since it has a `modules` db_type, just like the KEGG modules db, it is treated the same way, but I didn't test whether this is appropriate.

@meren You said you were excited to overload these classes with fun `anvi-db-info` information. If you have time, either in this PR or another, can you please take the first stab?

Finally, there remains two open topics that probably deserve their own PRs, but I will mention them here.

1. All of the pre-existing utils functions like `is_contigs_db` now create instances of `DBInfo` to satisfy their tasks. Are we okay with that or do we want to start moving things around to reduce redundancy? I am happy to keep things as they are but it is a bit redundant...
2. There is a lot of untapped potential with these new classes. My first instinct was to develop methods that link databases to each other and checking for hash consistencies etc. I was imagining some methods like `attach(other)` which would check the hash consistency and then add the database to a dictionary of the other database, and vice versa. Probably there are better ways, perhaps that make use of our artifact network.

As a final note, `db.DB` is a now a context manager. If you've been trying your best to manage your `disconnect` method calls, you don't have to anymore. You can now initialize `db.DB` as a context manager:

```python
with DB(path, None, ignore_version=True) as database:
    if 'self' not in database.get_table_names():
        return False
```

Upon exiting the `with` statement, the database will disconnect, even if errors are produced.

ivagljiva:
Hey @ekiefl, this is awesome!

`anvi-self-test --suite metabolism` only tests `anvi-estimate-metabolism`, so I guess that program is good. I also tested `anvi-setup-kegg-kofams` and `anvi-run-kegg-kofam` (the only other two programs that use the modules db). Everything looks good for those as well.

As for the double `is` functions, that must be an artifact from when I changed modules_db to be called kegg_modules_db everywhere in the code. I got rid of `is_modules_db` and changed that one instance to use `is_kegg_modules_db`.

meren:
I'm merging this to make sure it gets tested, there is no reason for it to be sitting here idle.

I'm hoping that there will future PRs to make good use of the `DBInfo` class to reduce redundancy in the codebase :)

Thank you all, especially @ekiefl, for your attention and help with this!

PR #1663, opened 2021-02-05
Proper pnps

ekiefl:
Starting a PR to track progress on this front

# The need

Essentially, I have been calculating pN/pS in a lazy and somewhat inaccurate manner. Rather than calculating a fractional pN and pS score per position, by comparing allele counts, I was instead just saying "is this SCV also a SAAV? Then it's non-synonymous! Otherwise its synonymous!"

# The new approach.

The new approach goes like this.

1. Find a SCV and set pN and pS to 0.
2. Determine what should be the reference that synonymous/non-synonymous is relative to. The user can pick either 'consensus' or 'reference'. If 'consensus' each allele is compared to the consensus allele at that position. If 'reference', each allele is compared to the reference allele at that position.
3. Loop through each coding codon and determine (a) its relative abundance and (b) whether its synonymous or non-synonymous with the reference. Add its relative abundance to pN or pS depending on where its synonymous or non-synonymous.
4. After looping through each coding codon, calculate the number of synonymous and non-synonymous sites of the reference codon. Then divide pN and pS by these respective numbers.
5. Repeat steps 1-4 until there are no SCVs left.

Besides being more canonically accurate, this per-SCV calculation of pN and pS in theory allows one to calculate pN/pS over arbitrary groupings of SCVs, not just (gene, sample) groupings, like is currently supported. For example, one could add functionality to group by relative solvent accessibility bins without much trouble at all.

# Comparison

Here I calculated pn/ps for 800ish genes over 70ish samples. Each datapoint is an environmentally averaged pn/ps value for a gene.

image

Here I ran the new method with `--comparison reference`. It is honestly very surprising how well the old method compares to the new method.

Those are environmentally averaged values. Here are a handful of values across metagenomes, where each plot is one gene.

image

image

image

image

Overall, you can see they correlate quite well, although much more deviation is observed. In all genes the trend line fill below `y=x`, indicating that the old method overestimated pN/pS. This makes sense because if a codon position was a SAAV, the old method precluded the possible of synonymous contribution, even though much of the variation may have been synonymous.

In the last example, you can see that the old method broke down because any gene without at least 1 SAAV would necessarily get a score of 0. In constrast, the new method allows for much more resolved values when SAAV count is low or even non-existant.

# Problems

This is currently very slow. I am going to try and speed it up in future commits of this PR with a numba implementation


ekiefl:
After prepping for a conversion to numba, which is seen in the latest commit, I can confirm that the functionality has not changed at all. I am getting the same values as before (ignore axes):

image

ekiefl:
Ok so things have gotten a lot faster. Here is a speed comparison for 3 different commit states, all of which produce exactly the same results:

### 7d3e304 = 28 mins
- I used a hash lookup for calculating synonymity, but I repetitively call the function `utils.get_synonymous_and_non_synonymous_potential` millions of times

### 12fac60 = 1 mins 54 secs
- avoid looping through a dataframe and instead loop through a numpy array
- I convert the hash lookup for calculating synonymity to a numpy array lookup
- I avoid calling `utils.get_synonymous_and_non_synonymous_potential` by creating a numpy array for fast lookup
- All of this is in anticipation of converting to numba, but already its much much faster

### 5e6abc5 = 0 mins 1.6 secs
- I numba-ize the function
- There's nothing else to say

So array-based lookups clearly helped massively (14x speed gain). But the biggest speed gain was numba, which was 72X faster than array-based lookups, and 1050X faster than the original implementation.

I forgot that speed gains get me HIGH> USDLK:FSDJ: FDS J:DSF J:

meren:
Crap. This must be the best application of mumba! :) Awesome 😎

PR #1672, opened 2021-02-13
Reducing stringency of anvi-run-kegg-kofams with a new annotation heuristic

ivagljiva:
Happy Friday, people. In this PR I'd like to present a heuristic for making `anvi-run-kegg-kofams` a bit less stringent.

**Motivation**
It turns out that KEGG's bitscore thresholds (which we use to get rid of weak HMM hits in `anvi-run-kegg-kofams`) are too stringent sometimes, leading us to miss valid KOfam annotations. This is problematic because downstream code, namely `anvi-estimate-metabolism`, relies on these annotations - if we miss valid annotations, then we will underestimate metabolic capability. 

Here's an example. @meren first noticed this issue when examining a pangenome (of 3 genomes), in which certain core gene clusters contained 2 genes with a KOfam annotation but the 3rd gene was not labeled with the KO, even though all 3 genes were annotated with the same COG. Upon examining one such gene cluster, we found that the  gene with the missing KO was extremely similar to the other genes (99% identity), but its HMM hit to the KO was just barely under the bitscore threshold (a score of 679.7, when the threshold was 680.23). Meren lab members can see more details on this example [here](https://github.com/merenlab/lab-fiesta/issues/1242).

**Heuristic**
We want to make sure we are including valid KO annotations like that, without being so permissible that we add garbage annotations to the database. The heuristic that has been implemented (credits to @meren for the idea :)  is as follows:

> For a given gene: if there is no KOfam hit above the bitscore threshold, we examine all the hits with an e-value <= X and a bitscore > Y% of the threshold. If those hits are all to a unique KOfam profile, then we annotate the gene call with that unique KO.

The current default for the e-value threshold (X) is 1e-05 and the current default bitscore fraction (Y%) is 0.5, or 50%. These parameters can be changed; I'll get to that in a second.

**Relevant parameters**
We apply this heuristic by default, but people who don't like this at all can turn it off with the `--skip-relaxation-heuristic` flag. 

Those who want this to happen, but want the heuristic to be more or less stringent can set the e-value threshold (X) with `-E` or the bitscore fraction (Y) with `-H`. 

**Default parameter values**
As I mentioned, the current defaults are an e-value of 1e-05 and a bitscore fraction of 0.5. I have tested these defaults on the same pangenome in which the issue was discovered, and they seem to perform reasonably well. Here is a summary of that investigation (more details can be found in this frighteningly-long comment [here](https://github.com/merenlab/lab-fiesta/issues/1242#issuecomment-778564871)):
- Of 32 core gene clusters that originally had only 1 out of 3 genomes with a KO-annotated gene, 15 gained 2 KO annotations to become a cluster with 3 out of 3 genomes having a KO-annotated gene. Only 1 gained a single KO annotation to become a cluster with 2/3 genomes having that KO annotation.
- Of 46 core gene clusters that originally had 2 out of 3 genomes with a KO-annotation, 24 gained 1 (or more) KO annotations to become a cluster with 3 out of 3 genomes having at least one KO-annotated gene.
- A gene cluster with multiple different KO annotations could be a marker of garbage. Applying this heuristic with default parameters added 3 such gene clusters to the pangenome. 1 of these was valid, but the other 2 contained garbage KO annotations.

In short, these default values do a decent job of adding valid KO annotations that are below the standard KEGG bitscore thresholds. They do not find everything - I saw a few genes that were still not annotated despite having very good alignments to other KOs in the same cluster (and the same COG annotation) - but it is decent. They also do not add too much garbage, as far as I could tell - the two bad annotations that I found while investigating clusters with multiple KOs were the only ones I saw, though there could be more.

However, my testing so far was not very rigorous, and I have no doubt that we could find better thresholds to serve as our defaults. I am currently running a script to make a bunch of pangenomes using annotations with different combinations of the heuristic parameters, and I will compare those results to see if there is a better set of defaults out there. If I find one, I'll update the defaults. (It's unlikely I'll find the most optimal set even after doing this, so I have added a cautionary note to the `anvi-run-kegg-kofams` docs for users to be aware of). 

I do not mind if we merge this PR before I finish testing different parameter combinations - I know at least one person who is waiting to re-run their analyses using this heuristic. In that case, I can make a different PR later to update the defaults if necessary. But if it is a better idea to set the defaults properly first before merging, that is fine with me, too :)

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2660,8 +2660,8 @@ D = {
                      "will only look at hits with bitscores > the KEGG threshold * this number. (This is Y.) "
                      "It should be a fraction between 0 and 1 (inclusive)."}
                 ),
-    'skip-relaxation-heuristic':(
-            ['--skip-relaxation-heuristic'],
+    'skip-bitscore-heuristic':(
+            ['--skip-bitscore-heuristic'],
             {'default': False,
              'action': 'store_true',
              'help': "If you just want annotations from KOfam hits that are above the KEGG bitscore "
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2660,6 +2660,13 @@ D = {
                      "will only look at hits with bitscores > the KEGG threshold * this number. (This is Y.) "
                      "It should be a fraction between 0 and 1 (inclusive)."}
                 ),
+    'skip-relaxation-heuristic':(
+            ['--skip-relaxation-heuristic'],
+            {'default': False,
+             'action': 'store_true',
+             'help': "If you just want annotations from KOfam hits that are above the KEGG bitscore "
+                     "threshold, use this flag to skip the mumbo-jumbo we do here to relax those thresholds. "}
+                ),
     'include-metadata': (
             ['--include-metadata'],
             {'default': False,
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2643,6 +2643,23 @@ D = {
              'help': "Use this flag to generate a tab-delimited text file containing the bit scores "
                      "of every KOfam hit that is put in the contigs database."}
                 ),
+    'heuristic-e-value': (
+            ['-E', '--heuristic-e-value'],
+            {'default': 1.0e-5,
+             'metavar': 'FLOAT',
+             'type': float,
+             'help': "When considering hits that didn't quite make the bitscore cut-off for a gene, we "
+                     "will only look at hits with e-values <= this number. (This is X.)"}
+                ),
+    'heuristic-bitscore-fraction': (
+            ['-H', '--heuristic-bitscore-fraction'],
+            {'default': 0.50,
+             'metavar': 'FLOAT',
+             'type': float,
+             'help': "When considering hits that didn't quite make the bitscore cut-off for a gene, we "
+                     "will only look at hits with bitscores > the KEGG threshold * this number. (This is Y.) "
+                     "It should be a fraction between 0 and 1 (inclusive)."}
+                ),
     'include-metadata': (
             ['--include-metadata'],
             {'default': False,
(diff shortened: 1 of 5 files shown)


meren:
Thank you very much for this PR, @ivagljiva, and your extensive documentation of your solution. 

I have only minor suggestions:

I think `self.heuristic_e_value` and `self.heuristic_bitscore_fraction` would be much more descriptive in the code if you replace those variable names with `self.bitscore_heuristic_e_value` and `self.bitscore_heuristic_bitscore_fraction`. And`--skip-relaxation-heuristic` with `--skip-bitscore-heuristic`. I think this is ready to merge even if you don't want to change those variable names for a reason that I miss.

I like your preliminary tests, and I'm wondering if those false positives would have behaved much better with 0.75 rather than 0.5 cutoff.

Either ways, this will be a game changer to cast a slightly wider net to the cloud of good hits that are just below the noise cutoffs that could only be defined given the database of known genes :)

meren:
I also think we should rename the class `KeggRunHMMs` to `RunKOfams` :)

meren:
🎊

ivagljiva:
Thanks for your suggestions, @meren! They are in.

> I like your preliminary tests, and I'm wondering if those false positives would have behaved much better with 0.75 rather than 0.5 cutoff.

I am testing this now. Do you want me to wait on merging this until after that is done, or should we just merge and update the defaults later as needed? :)

meren:
Thank you Iva. I think you should merge it :) We can update defaults later.

ivagljiva:
A small update for the community who is following this change:

After some more investigation, I determined that a bitscore fraction of 0.75 is better than 0.5. Therefore 0.75 is now the default, as of [this commit](https://github.com/merenlab/anvio/commit/a41801378be96bd23c810a9de8b099e793974f65) :)

PR #1676, opened 2021-02-16
Make sure all int/external genomes have KOfams

meren:
A small PR to make sure running `anvi-estimate-metabolism` does not explode after going through a few hundred genomes, but explodes much earlier.


ivagljiva (commented):
Oh. I am sorry this failed, @meren :( But your fix looks good to me!

meren:
No worries at all! :) We will iron these wrinkles over time :)

Issue #1675, opened 2021-02-15
[FEATURE] add flag to `anvi-run-hmms` to export raw output of `hmmscan`


mschecht:
## The need

`anvi-run-hmms` only provides the best hit based on e-value. It would be great if the user could have access to the raw `hmmscan` output files to look at other sequences that aligned to the hmm models provided by anvio.

## The solution

I think a solution to this problem would be to add a parameter to `anvi-run-hmms` which allows the user to provide a path to a directory to put the raw `hmmscan`/`hmmsearch` files for their own perusal. 

## Beneficiaries

This would benefit any anvio user who wants to take a deeper dive into the results of the `hmmscan` driver in the anvio code base without using the `--debug` flag.

ivagljiva:
Alright @mschecht, I've done some work on this and would like you to check it out to make sure it fills your needs :)

Please checkout the branch `fr_1675` to test it out. 
You can run 
```
anvi-run-hmms -c CONTIGS.db --hmmer-output-dir hmmer_output_test -I Bacteria_71
``` 
to get a directory that contains the `hmmscan` output (only the regular table file). If you also want the domain table file in that directory, you can run 
```
anvi-run-hmms -c CONTIGS.db --hmmer-output-dir hmmer_output_test -I Bacteria_71 --get-domtable-output
``` 
Please note that in the latter case, the domain table output will not be used to annotate the contigs database (it is still only the regular table file that is used for that).

I would appreciate hearing your feedback on whether this is what you need for PR #1678, and whether you find any bugs while testing. Thanks!

mschecht:
@ivagljiva I can't thank you enough for implementing this! 

> I would appreciate hearing your feedback on whether this is what you need for PR #1678, and whether you find any bugs while testing.

`fr_1675` works out of the box with #1678!

Here I start in your branch to get domtblout with `anvi-run-hmms` then move to my branch and seamlessly use `anvi-script-filter-hmm-hits-table`:
```
# start in Iva's branch where she added in the correct way to get domtblout (god bless Iva!)
# git checkout fr_1675
$ anvi-run-hmms -c CONTIGS.db --hmmer-output-dir hmmer_output_test_dom_hmmsearch -I Bacteria_71 --get-domtable-output --just-do-it --hmmer-program hmmsearch

WARNING
===============================================
Previous entries for "Bacteria_71" is being removed from "hmm_hits_info,
hmm_hits, hmm_hits_in_splits, genes_in_contigs, gene_functions"

[...]

PR #1678, opened 2021-02-17
Anvi-script-filter-hmm-hits-table

mschecht:
Hi everyone, here is my attempt at making the program `anvi-script-filter-hmm-hits-table`. @ivagljiva thank you so much for being available to bounce ideas around, I owe you big time!

The goal of this program is to replace the `hmm_hits` table in a contigsDB with a new, filtered set of `hmm_hits`. At the moment, anvi'o only filters `hmm_hits` based on e-value. I believe users may want more control over which HMM alignments are kept, whether it's making the alignments more sensitive or specific with various filters. 

At the moment the program can only filter based on target-coverage and query-coverage but hopefully we can add more filtering parameters in the future. Additionally, the program can only work if `hmmsearch` and hmm models built from protein sequences were used during `anvi-run-hmms`. This is because the columns from `hmmsearch` that are needed to calculate query and target coverage can only be found in `domtblout` which is only an output format with amino acid HMMs. 

To make a reproducible example (see below) I hacked my way through `anvi-run-hmms` to give me the `domtblout` from `hmmsearch`. I did not put that much thought into that aspect of this PR and hopefully, someone from the community can address it in this [feature request](https://github.com/merenlab/anvio/issues/1675) because when `anvi-run-hmms` is multi-threaded it creates multiple output files and I did not see a clear path to addressing that. But I did manage to get the `domtblout` and get started developing `anvi-script-filter-hmm-hits-table`.

Reproducible example from infant gut:
```
wget https://ndownloader.figshare.com/files/26218961 -O INFANT-GUT-TUTORIAL.tar.gz
tar -zxvf INFANT-GUT-TUTORIAL.tar.gz && cd INFANT-GUT-TUTORIAL
anvi-migrate --migrate-dbs-safely *.db
```

re-run  `anvi-run-hmms` to produce `domtblout` 
```
 $ anvi-run-hmms -c CONTIGS.db -I Bacteria_71 --just-do-it \
                 --debug --hmmer-program hmmsearch --hmm-domain-tblout-path test_hmmsearch

WARNING
Previous entries for "Bacteria_71" is being removed from "hmm_hits_info,
hmm_hits, hmm_hits_in_splits, genes_in_contigs, gene_functions"


[DEBUG] `run_command` is running .............: hmmpress /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmpxtup18ei/Bacteria_71.hmm

Contigs DB ...................................: CONTIGS.db
HMM sources ..................................: Bacteria_71
Alphabet/context target found ................: AA:GENE

HMM Profiling for Bacteria_71
Reference ....................................: Lee modified, https://doi.org/10.1093/bioinformatics/btz188
Kind .........................................: singlecopy
Alphabet .....................................: AA
Context ......................................: GENE
Domain .......................................: bacteria
HMM model path ...............................: /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmpxtup18ei/Bacteria_71.hmm
Number of genes in HMM model .................: 71
Noise cutoff term(s) .........................: --cut_ga
Number of CPUs will be used for search .......: 1
HMMer program used for search ................: hmmsearch
Temporary work dir ...........................: /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmp7zpqt6hn
Log file for thread 0 ........................: /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmp7zpqt6hn/AA_gene_sequences.fa.0_log

[DEBUG] `run_command` is running .............: hmmsearch -o /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmp7zpqt6hn/AA_gene_sequences.fa.0_output --cut_ga --cpu 1 --tblout
                                                /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmp7zpqt6hn/AA_gene_sequences.fa.0_table --domtblout test_hmmsearch_domtable
                                                /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmpxtup18ei/Bacteria_71.hmm /var/folders/l4/c5v07vmj4jj9phh9lqm_hbz00000gq/T/tmp7zpqt6hn/AA_gene_sequences.fa.0

* 1 out of 1 have finished
Number of raw hits ...........................: 674
Number of hits in annotation dict  ...........: 674

* Adding the chunk 0 with 674 entries of 674 total is being added to the db with
the SQL command 'INSERT INTO hmm_hits VALUES (?,?,?,?,?,?,?)'.

* Adding the chunk 0 with 674 entries of 674 total is being added to the db with
the SQL command 'INSERT INTO hmm_hits_in_splits VALUES (?,?,?,?)'.

✓ anvi-run-hmms took 0:00:10.224763
```

filter out hmm_hits with query_coverage (HMMs) x > 0.9
```
$ anvi-script-filter-hmm-hit-table -c CONTIGS.db --domtblout test_hmmsearch_domtable \
                                   --hmm-source Bacteria_71 --target-coverage 0.9

HMM profiles .................................: 9 sources have been loaded: Bacteria_71 (71 genes, domain: bacteria), Archaea_76 (76 genes, domain: archaea), Ribosomal_RNA_23S (2 genes, domain: None), Ribosomal_RNA_28S (1 genes,
                                                domain: None), Ribosomal_RNA_5S (5 genes, domain: None), Ribosomal_RNA_16S (3 genes, domain: None), Ribosomal_RNA_12S (1 genes, domain: None), Protista_83 (83 genes, domain:
                                                eukarya), Ribosomal_RNA_18S (1 genes, domain: None)
Database Path ................................: CONTIGS.db

WARNING
Previous entries for "Bacteria_71" is being removed from "hmm_hits_info,
hmm_hits, hmm_hits_in_splits, genes_in_contigs, gene_functions"


WARNING
This is the base parser class--a part of the code you should never hear from.
PLEASE READ THIS CAREFULLY. While anvi'o was trying to parse some files
assocaited with the annotation source `HMMScan`, it found that 16 of the lines
in this file were not able to made sense of. This part of the code does not know
anything more than that. It doesn't even know what file it is. But in general
this error occurs when the mapping function does not find what its looking for
in a line. For instance, a value that was supposed to be an integer ends up
being actually a piece of text or something. Well. Here are the line numbers if
you care and can make sense of this information: 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
1, 1, 1, 1, 1, 1

Number of hits in annotation dict  ...........: 176

```

`Number of hits in annotation dict` decreased from 674 to 176 with the strict filtering threshold of query_coverage > 0.9. 

I would also like to add that I wrote this program in ~1.5 days. This is a drastic improvement compared to when I first joined the Meren lab :)

Looking forward to your comments and suggestions!


anvio/__init__.py
```diff
@@ -2177,6 +2177,13 @@ def TABULATE(table, header, numalign="right", max_width=0):
                      "This way you can run HMM profiles that are not included in anvi'o. See the online "
                      "to find out about the specifics of this directory structure ."}
                 ),
+    'hmm-domain-tblout-path': (
+            ['-F', '--hmm-domain-tblout-path'],
+            {'metavar': 'HMM OUTPUT FORMAT',
+             'help': "HMMER has a few output formats for including --tblout and --domtblout"
+                     "depending on the alphabet of your alignment. By default anvi'o will provide"
+                     "--tblout but if you are doing a protein search you can change to --domtblout."}
```
ivagljiva, line 2185: I think this help message would be more clear if you instead described that users can provide a path to this argument, and the domain table output will be saved to that path. :) 
Right now it is not obvious that this will happen, and this param just sounds like a flag to turn `--domtblout` on.

sandbox/anvi-script-filter-hmm-hit-table
```diff
@@ -0,0 +1,223 @@
...
+
+        if not self.hmm_source:
+            raise ConfigError("Please provide a hmm-source :)")
+
+        info_table = hmmops.SequencesForHMMHits(self.contigs_db_path).hmm_hits_info
+
+        if self.hmm_source not in info_table:
+            raise ConfigError("Whoa there, the HMM source "'%s'" is not in your contigsDB: %s "
+                              "Please try re-running anvi-run-hmms and make sure to specify you HMM source of interest." % (self.hmm_source, self.contigs_db_path))
+
```
ivagljiva, line 80: You probably don't know about this convention yet, but a few months(?) ago we decided to start using python formatted strings (f-strings) instead of this %s replacement syntax. It is a bit more readable :) I will update some of these in a sec so you can see how it looks like, but no worries if you don't feel like changing the rest (we left all our old %s code alone when we decided to switch) - it's just one more thing to think about for next time!

sandbox/anvi-script-filter-hmm-hit-table
```diff
@@ -0,0 +1,223 @@
...
+        self.sanity_checks()
+
+        self.import_domtblout()
+
+        x = self.filter_domtblout()
+
+        self.get_hmm_hit_table(hmmsearch_tbl=x)
+
+        # remove the tmp file
+        os.remove(x)
```
ivagljiva, line 64: Usually when I remove files, I include the option to keep them around if the user runs with `--debug`. This way, you don't clutter up the tmp directory on their computer, but the user can explicitly request to see this file in case something goes wrong and they need to check it.

sandbox/anvi-script-filter-hmm-hit-table
```diff
@@ -0,0 +1,223 @@
...
+
+
+    def filter_domtblout(self):
+
+        # Filtering conditions
+        if self.target_coverage:
+            df_filtered = self.df[self.df['target_coverage'] > float(self.target_coverage)]
+
+        if self.query_coverage:
+            df_filtered = self.df[self.df['query_coverage'] > float(self.query_coverage)]
```
ivagljiva, line 144: If the user provides both target coverage and query coverage filtering values, these two conditionals will run in addition to the one below, which _I think_ makes the below if statement redundant? I didn't think about it too much so I might be wrong, but I just wanted to bring it up so you can take a look and make sure that is what you want to happen :)

anvio/docs/programs/anvi-script-filter-hmm-hit-table.md
```diff
@@ -0,0 +1,27 @@
+This program is for filtering a %(hmm-source)s from a %(hmm-hits)s in a %(contigs-db)s using HMM alignment parameters such as query-coverage and target-coverage. Briefly, the program will remove all records from an %(hmm-source)s in the %(hmm-hits)s then import a new %(hmm-hits)s table into the %(contigs-db)s that was filtered to your specifications. At the moment, this tool is only designed to work with `hmmsearch` with protein sequences. The `--domtblout` can be produced from running `anvi-run-hmms` with your %(hmm-source)s of interest and using the `--domtblout` parameter AND  `hmmsearch` as the program.
```
ekiefl, line 1: If you don't provide `hmmsearch` as the program is an error raised? Maybe the program should automatically be set to `hmmsearch` if `--domtblout` is provided.

anvio/drivers/hmmer.py
```diff
@@ -203,20 +203,29 @@ def run_hmmer(self, source, alphabet, context, kind, domain, num_genes_in_model,
...
             thread_num += 1
 
+            domain_out_fmt = "--domtblout"
+
             if noise_cutoff_terms:
                 cmd_line = ['nhmmscan' if alphabet in ['DNA', 'RNA'] else self.program_to_use,
                             '-o', output_file, *noise_cutoff_terms.split(),
                             '--cpu', cores_per_process,
                             out_fmt, table_file,
+                            domain_out_fmt, domain_table_file,
```
ekiefl, line 221: Does this not fail when `domtblout_path` is None? I would expect it to fail in all the other conditionals in which `cmd_line` is defined, too.

ekiefl (commented):
This is looking really good @mschecht. Congrats on writing this OO design in `anvi-script-filter-hmm-hits-table`.

sandbox/anvi-script-filter-hmm-hit-table
ivagljiva, line 64: One thing we usually like to do with classes is to keep the 'work' that the class is doing separate from the initialization code. That way, you can get an instance of the class without always having to go through all those other steps, if you need to. It makes the class (even) more modular and generalizable. 

In this case, I would take all these functions out of `__init__` and put them into a separate, "driver" function that can be called by the `main` code below after a FilterHmmHitsTable() instance is initialized. 

I hope that made sense :/ It would probably be better to see an example? Then you can decide if/how to change your code to follow this pattern. 

Here is an example, in `kegg.py` you can see that the `RunKOfams` class's `__init__` only grabs arguments and does sanity checks. 
image

Then later in the code there is a driver function `process_kofam_hmms()` that actually does all the work. So in `anvi-run-kegg-kofams`, the script 1) initializes and 2) calls that driver:
```
def main(args):
    p = kegg.RunKOfams(args)
    p.process_kofam_hmms()
```

ivagljiva (commented):
Very nice, @mschecht! And speedy, too :)

meren:
I haven't looked at the code carefully, but seeing comments from Iva and Evan, I have no doubt that its going to be OK. Thank you all very much.

Although, totally minor point, the name of the table is "hmm hits" so the script name should reflect that, IMO (i.e., `anvi-script-filter-hmm-hits-table` and not `anvi-script-filter-hmm-hit-table`, which will also require an update for the file name `anvi-script-filter-hmm-hit-table.md`).

Also it seems while the flag (and the corresponding variable name) in `anvi-hmm-hits` is `--hmm-domain-tblout-path`, it is called `--domtblout` in `anvi-script-filter-hmm-hit-table`. I think it would be better if these were the same.

anvio/docs/programs/anvi-script-filter-hmm-hit-table.md
mschecht: That's a great point @ekiefl! For this PR, I was mainly focusing on `anvi-script-filter-hmm-hits-table` so I did not think about checks and balances for the `-domtblout ` in `anvi-run-hmms`. In fact, I added the bare minimum so that I could start developing `anvi-script-filter-hmm-hits-table`. Do you think that functionality should be addressed in this PR. If so, I would be happy to start to improving this.

sandbox/anvi-script-filter-hmm-hit-table
mschecht: Thanks for pointing that out @ivagljiva, I'll change all of them to f-strings to practice the syntax :)

sandbox/anvi-script-filter-hmm-hit-table
ekiefl: I fixed this

sandbox/anvi-script-filter-hmm-hit-table
mschecht: Great idea, I added that funcitonality here: [9209511](https://github.com/merenlab/anvio/pull/1678/commits/9209511bd36c8279c131548ed9d36aa302d0f250)

mschecht: Addressed [here](https://github.com/merenlab/anvio/pull/1678/commits/504d0ca3d7cf99147523da27c70289e85705699c)

mschecht:
> Although, totally minor point, the name of the table is "hmm hits" so the script name should reflect that, IMO (i.e., `anvi-script-filter-hmm-hits-table` and not `anvi-script-filter-hmm-hit-table`, which will also require an update for the file name `anvi-script-filter-hmm-hit-table.md`).

Good idea @meren, addressed here: [44a0a20](https://github.com/merenlab/anvio/pull/1678/commits/44a0a20e63ae997e7afc9f1b46d20c9824795d6e)

anvio/docs/programs/anvi-script-filter-hmm-hit-table.md
mschecht: In this [commit](https://github.com/merenlab/anvio/pull/1678/commits/76a6b733e66e4363bf370c2c96956eeba1c20951) I added a config Error to stop the user from using hmmscan with `--hmm-domain-tblout-path`

anvio/__init__.py
mschecht: Good call. I hope this clears that up: [a9fc2f2](https://github.com/merenlab/anvio/pull/1678/commits/a9fc2f226f798588cef86ffb884ba0fff688515f)

anvio/drivers/hmmer.py
mschecht: I tried to address this [here](https://github.com/merenlab/anvio/pull/1678/commits/fb28845c4b873155423c677f9423ec2eadc44a8b)

mschecht:
Hi everyone,  when `anvi-run-hmms` is multi-threaded AND`domtblout` is provided, all the results from each thread are not concatenated to the final `domtblout`. I added a FIXME to this commit ([fb28845](https://github.com/merenlab/anvio/pull/1678/commits/fb28845c4b873155423c677f9423ec2eadc44a8b#diff-5414c9d8785789b62d8d5e1fcb68a664ad0813b2ec9bcd69000cfcdc6fde524bR245-R246)) where I think issue is arising. Any suggestions on how to fix this?

meren:
This looks great! Thanks @mschecht, and everyone else who contributed!

I shared some minor suggestions with @mschecht, and he will consider them in his RiboPhylo branch :)

PR #1683, opened 2021-02-25
Add coverage and detection values to metabolism output

ivagljiva:
This PR adds the `--add-coverage` flag to `anvi-estimate-metabolism`. When this flag is provided along with a profile database, long-format output files will contain additional columns containing gene coverage and detection information. This works for genome mode, collection mode, metagenome mode, and any of the multi modes (as long as `profile_db_path` is in the input file). 

```
anvi-estimate-metabolism -c CONTIGS.db -p PROFILE.db -O output_with_coverage --add-coverage --kegg-output-modes modules,kofam_hits_in_modules
```

In `modules` mode, the additional columns are of the following format:

| SAMPLE_1_gene_coverages | SAMPLE_1_avg_coverage | SAMPLE_1_gene_detection | SAMPLE_1_avg_detection |
|:--|:--|:--|:--|
| 3.0,5.0,10.0,2.0 | 5.0 | 1.0,1.0,1.0,1.0 | 1.0 |

These four columns are added for each sample in the provided profile database. The `gene_coverages` and `gene_detection` columns contain comma-separated mean coverage/detection values for each gene call in the current module (in the same order as the `gene_caller_ids_in_module` column). The `avg_coverage` and `avg_detection` columns contain the averages of these values, respectively. 

Since `modules` mode output can be customized, users have the ability to pick which of these additional columns are put into the output file (and in which order) by using `--kegg-output-modes modules_custom` and `--custom-output-headers`. For example:
```
anvi-estimate-metabolism -c CONTIGS.db -p PROFILE.db --add-coverage --kegg-output-modes modules_custom -O modules_custom_coverage \
    --custom-output-headers kegg_module,module_is_complete,DAY_15A_gene_coverages,DAY_15A_avg_coverage,DAY_15A_gene_detection,DAY_15A_avg_detection
```

In `kofam_hits_in_modules` mode, the added columns are of the following format:

| SAMPLE_1_coverage | SAMPLE_1_detection |
|:--|:--|
| 3.0 | 1.0 |

Again, these two columns are added for every sample in the profile database (no customization is available for this mode). Since every row in this output mode contains one KOfam hit, the `coverage` column contains the mean coverage of that particular gene and the `detection` column contains the detection of that gene. 

This PR passes `anvi-self-test --suite metabolism`, which now also includes some tests for `--add-coverage` cases.


anvio/__init__.py
```diff
@@ -2358,6 +2358,13 @@ def TABULATE(table, header, numalign="right", max_width=0):
              'help': "When declared, only reads that cover all positions will be reported. It is necessary to use this "
                      "flag if you want to perform oligotyping-like analyses on matching reads."}
                 ),
+    'add-coverage': (
```
meren, line 2361: I wonder if one could use the existing `--init-gene-coverages` flag for this :)

anvio/docs/artifacts/kegg-metabolism.md
```diff
@@ -19,6 +19,15 @@ The 'modules' output file will have the suffix `modules.txt`. Each line in the f
...
 
+**Adding coverage and detection values to output**
+If you use the flag `--add-coverage` and provide a profile database, additional columns containing coverage and detection data will be added for each sample in the profile database. Here is an mock example of the additional columns you will see (for a generic sample called 'SAMPLE_1'):
+
+| SAMPLE_1_gene_coverages | SAMPLE_1_avg_coverage | SAMPLE_1_gene_detection | SAMPLE_1_avg_detection |
+|:--|:--|:--|:--|
+| 3.0,5.0,10.0,2.0 | 5.0 | 1.0,1.0,1.0,1.0 | 1.0 |
+
+In this mock example, the module in this row has four gene calls in it. The `SAMPLE_1_gene_coverages` column lists the mean coverage of each of those genes in SAMPLE_1 (in the same order as the gene calls are listed in the `gene_caller_ids_in_module` column), and the `SAMPLE_1_avg_coverage` column holds the average of these values. As you probably expected, the `detection` columns are similarly defined, except that they contain detection values instead of coverage.
+
```
meren, line 30: As always, thank you very much for keeping up with the documentation, @ivagljiva :)

meren (approved):
looks great to me :) very exciting how this all come together.

anvio/__init__.py
ivagljiva: Technically, yes, we could. :) However, in order for the coverage/detection output to be customizable, it is necessary to initialize the profile database at the beginning of the run so that available headers to choose from are visible when using the flag `--list-available-output-headers`. If we were to use `--init-gene-coverages`, this initialization would load the coverage data, and that takes too much time if all you are doing is listing the headers.

(in fact, this initialization is what was causing the mysterious behavior that we couldn't figure out yesterday - I had two possible locations for profile db initialization - one at the beginning to handle the custom output headers case and one later in the code, and I placed `self.args.skip_create_gene_dbs = True` in the wrong one :)

Well, actually, now that I am thinking about it, I can simply turn `--init-gene-coverages` off if  `--list-available-output-headers` is on. 🤔 AND then this would eliminate the need for two functions that can initialize the profile database. Okay, never mind. :) I am convinced that this is the right way to go. I will change it.

ivagljiva: Well, it turns out that I will not be changing this, after all :) 
After some reflection, I think the design is better the way it is. The reason is that using `--init-gene-coverages` forces us to always initialize coverage data for all genes in the contigs database, even if not all of these genes have KOfam hits. (because we initialize the profile db with this parameter before we call `estimate_metabolism()` due to the need to know about available headers, as I alluded to in my previous comment). If we instead wait and call `profile_db.init_gene_level_coverage_stats_dicts()` after we know which genes we care about, it prevents us from processing this information for irrelevant gene calls, which should save us both time and memory. 

@meren, if you read this and still think it is better to go with `--init-gene-coverages` rather than `--add-coverage`, please let me know :)

meren: I am reading EVERYTHING and I am not thinking it is better to go with `--init-gene-coverages` :) apologies for the extra work and thanks for considering it. I agree that there is way too much power packed behind that community parameter in `dbops` and we can't just throw it around like that.

PR #1688, opened 2021-03-08
Functional enrichment refactor

meren:
This PR does two critical things:

* Unifies the running of `anvi-script-enrichement-stats` to a single utils function (that does all the footwork including casting int and float values to their proper types),
* Introduces a new feature for `anvi-display-functions` where this program can automatically compute functional enrichment stats according to a given `groups-txt` file and add as new layers to the interactive display. So the entire functional enrichment section the pangenomics tutorial is now reduced to a single line.

I did my best to test these features, but it would be an amazing help if @ivagljiva could test the parts that I will mark with individual comments in the PR.


anvio/kegg.py
```diff
@@ -5179,35 +5179,19 @@ def run_enrichment_stats(self):
...
-        if filesnpaths.is_file_empty(self.output_file_path):
-            raise ConfigError("It looks like something went wrong during the functional enrichment analysis. "
-                              "An output file was created, but it is empty "
-                              f"We don't know why this happened, but this log file could contain some clues: {log_file_path}")
 
+        # run the enrichment analysis
+        enrichment_stats = utils.run_functional_enrichment_stats(enrichment_input_path,
+                                                                 self.output_file_path,
+                                                                 run=self.run,
+                                                                 progress=self.progress)
```
meren, line 5195: I haven't tested this, @ivagljiva, would you mind making sure the output file is identical to the one you have tested previously?

anvio/docs/artifacts/groups-txt.md
```diff
@@ -1,13 +1,15 @@
-This is a two-column, TAB-delimited file that describes which samples belong to which groups. It is a required input for computing enrichment scores of metabolic modules in %(anvi-compute-functional-enrichment)s.
+This is a 2-column TAB-delimited text file to associate a given set of items with a set of groups. Depending on the context, items here may be individual samples or genomes. The first column can have a header name `item`, `sample`, `genome` or anything else that is appropriate, and list the items that are relevant to your input data. The second column should have the header `group`, and associate each item in your data with a group.
```
meren, line 1: @ivagljiva, I broadened the description of the groups-txt file. Is it OK?

meren:
The next step I see is to use `AggregateFunctions` class to report function occurrence stats, which will remove a ton of redundant code from the codebase.

We also need to update the pangenomics tutorial once all these features are in a stable release.

anvio/kegg.py
ivagljiva: Sorry for the delay, @meren. I just tested it, and the output file is identical to my previous copy (when they are sorted, which is as expected). :)

anvio/docs/artifacts/groups-txt.md
ivagljiva: Yes, I think it is okay, but in that case we should probably change which headers are accepted for this file. 
Right now, for running `anvi-compute-functional-enrichment` on metabolic modules, the first column is required to have the header `sample` - if not, you get an error that looks like this:
```
Config Error: The file 'test_diff_groups_txt_header.txt' does not contain the right type of
              header. It was expected to have these: 'sample, group', however it had these:
              'group'
```
(In this case, my file `test_diff_groups_txt_header` had `item` as the header of the first column)

I think it is likely we will use this groups-txt file for broader purposes in the future and it need not be restricted to that header, which is why I like that you changed this description. But now the code should also be flexible enough to accept any header. I can change it, if you agree.

anvio/__init__.py
```diff
@@ -2779,10 +2779,14 @@ def TABULATE(table, header, numalign="right", max_width=0):
             ['-G', '--groups-txt'],
             {'default': None,
             'metavar': 'TEXT_FILE',
-            'help': "A 2-column tab-delimited text file specifying which group each sample belongs to. "
-                    "The first column should have the header 'sample' and contain sample names matching "
-                    "to those are in the other input data. The second column should have the header 'group' "
-                    "and contain the group name for each sample (each sample should be in 1 group only)"}
+            'help': "A 2-column tab-delimited text file specifying which group each item belongs to. "
+                    "Depending on the context, items here may be individual samples or genomes. "
+                    "The first column should have the header 'name' and contain item names matching "
```
ivagljiva, line 2784: As I just mentioned in my previous comment, right now we require the first column to have the header `sample`, so this help text will have to change depending on whether we decide to keep the 'sample' requirement or allow any header for the first column.

anvio/docs/artifacts/groups-txt.md
meren: thank you very much for noticing this, @ivagljiva. I think it would be best if we changed the description. I think the best way to do is to have a function in `utils.py`, parallel to `get_samples_txt_file_as_dict`.

That function could do the sanity check for 'groups-txt', and look for a `group` column regardless of the number of columns in a given tab delimited file to only return the first and the groups column information (or it could also return a tuple of two dictionaries where in one of them keys are items and values are groups, and in the other keys are groups and values are a list of items in each group. With a function like that in utils, this entire code in `genomedescriptions.py`,

``` python
   (...)
            if self.layer_groups_input_file_path:
                filesnpaths.is_file_tab_delimited(self.layer_groups_input_file_path,)
                self.layer_groups = {}

                if 'group' not in utils.get_columns_of_TAB_delim_file(self.layer_groups_input_file_path):
                    raise ConfigError("A groups-txt file should have a single column that is called `group`.")

                layer_groups_d = utils.get_TAB_delimited_file_as_dictionary(self.layer_groups_input_file_path)

                for layer_name in layer_groups_d:
                    group_name = layer_groups_d[layer_name]['group']

                    if not group_name in self.layer_groups:
                        self.layer_groups[group_name] = []

                    self.layer_groups[group_name].append(layer_name)
            elif layer_groups:
                self.layer_groups = layer_groups

            # Finally, last AND least, a small helper dictionary we will use if there are gorups defined
            # by the user:
            if self.layer_groups:
                self.layer_groups_defined = True
                for group_name in self.layer_groups:
                    for layer_name in self.layer_groups[group_name]:
                        self.layer_name_to_group_name[layer_name] = group_name

   (...)
```

to this one to be used from everywhere in similar ways:

``` python
    (..)
    self.layer_groups, self.layer_name_to_group_name = utils.get_samples_txt_file_as_dicts(file_path)
    (...)
```

If you have the energy and throughput, please feel free to do it. If not, let me know and I can do it myself :)

PR #1690, opened 2021-03-10
A script to generate external and internal genomes files

ivagljiva:
Here is a very simple script that will automatically generate an external genomes file for all contigs dbs in a provided input directory, or an internal genomes file for a collection. 

## External genomes file 
You must provide an input directory and a name for the output external genomes file:
```
anvi-script-gen-genomes-file -I path/to/dir -e external_genomes.txt 
```

Here is an example from the Infant Gut dataset:
image

## Internal genomes file 
To get an internal genomes file, you must provide a profile database, its corresponding contigs database, and a collection.
```
anvi-script-gen-genomes-file -i internal_genomes.txt -c CONTIGS.db -p PROFILE.db -C default
```

Again, an example from the IGD:
image

I welcome suggestions for improvement :)


meren:
image

sandbox/anvi-script-gen-genomes-file
```diff
@@ -0,0 +1,136 @@
...
+                          "just so you don't forget, we also need an output file path in either case.")
+
+if __name__ == '__main__':
+    from anvio.argparse import ArgumentParser
+
+    parser = ArgumentParser(description=__description__)
+
+    groupE = parser.add_argument_group("EXTERNAL GENOMES FILE", "Provide a directory, and anvi'o will provide an external genomes "
+                                                                "file containing all contigs dbs in that directory.")
+    groupE.add_argument(*['-I', '--input-dir'], **anvio.K('input-dir'))
```
meren, line 118: `-I` is used for internal HMM models :) I would remove it and go with `*anvio.A('input-dir')` here.

ivagljiva: I started out with `*anvio.A('input-dir')`, but unfortunately that one takes an `-i` flag which conflicts with the `-i` for `*anvio.A('internal-genomes')` :) 
But what I can do is change the `-I` flag to something else, maybe `-D`? Or I can get rid of the short flag altogether and force people to use `--input-dir` in full.

meren: Let's remove `-i` for `--input-dir` from init since it shouldn't have been there at the first place.

ivagljiva:
> Let's remove -i for --input-dir from init since it shouldn't have been there at the first place.

It is gone forever. :)  And now the appropriate way to request an external genomes file is like this:
```
anvi-script-gen-genomes-file --input-dir path/to/dir -e external_genomes.txt
```
I have updated the help page, but wanted to record it here as well :)

PR #1706, opened 2021-03-30
works with output of contigs workflow

mschecht:
Hi Iva!

First off, thanks for making this! It's saved me a lot of time!

I've found that I use it a lot in the context of the contigs workflow which outputs `*-contigs.db` names for contigsDBs. To adjust for this I added a conditional to the script.

Please let me know if can make this more general or any other improvements.

### DIFF
--- a/sandbox/anvi-script-gen-genomes-file
+++ b/sandbox/anvi-script-gen-genomes-file
@@ -91,7 +91,10 @@ def main(args):
         ext_genomes_dict = {}
         for file in glob.glob(os.path.join(args.input_dir, '*.db')):
             if utils.is_contigs_db(file, dont_raise=True):
-                db_name = os.path.basename(file).split(".db")[0]
+                if "-contigs.db" in file:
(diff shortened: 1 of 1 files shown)


ivagljiva (approved):
It looks good to me, @mschecht, and I agree that it is useful to handle this case since we use this naming convention in a lot of places :) 

Thanks for the update!

meren:
We shouldn't learn the contents of the name column from contigs db file names. I think best would be `self.meta['project_name_str']` for this.

ivagljiva:
> I think best would be self.meta['project_name_str'] for this.

Done!

meren:
I think it is good to go :) Thank you very much, @mschecht and @ivagljiva!

ivagljiva:
> I think it would make more sense if is_contigs_db() returned False if dbinfo returns None. what do you think? would that break things elsewhere?

I am leaving this fix for another time :)

Issue #1670, opened 2021-02-11
A complete redesign of HMMs is needed


ekiefl:
As the number of ways and manners in which anvi'o interacts with HMMs has grown and grown, the quality of our HMM code has begun to deteriorate. Here, I have compiled a list of problems that I have identified.

## **Parsing is shared between both the driver and the parser.** 

`HMMer` from `anvio.drivers.hmmer.py` is a multi-threaded driver for `HMMer` which concatenates output files from individual hmmer runs. But it does more than concatenate raw output files. It _also_ does a light parsing and reformatting of the data into a file format that is compatible with the assumed file format of `Parser` from `anvio.parsers` and all its child classes.

In my opinion, `HMMer` should not do any parsing whatsoever during the concatenation of raw output files. At most, it should remove headers so that the output file looks as though it was created in a single hmmer run (i.e. no headers appearing in the middle of the file).

The responsibility of HMMER output parsing needs to be consolidated in one spot, aka the dedicated `HMMerTableOutput` parser. This is problematic because the raw information that you would expect a driver to provide you is actually not raw, or even complete.

## **`hmm_hits` table in the contigs database stores a subset of fields that it could/should**

For example, knowing the lengths of alignments, targets, queries, as well as the regions of coverage for targets and queries is useful information that anvi'o should store. To address this, [this branch](https://github.com/merenlab/anvio/tree/expand-hmm-hits-filtering) has been made to add these new columns, however the parser module suffers from certain shortcomings that make this very difficult to do cleanly (see next)

## **The HMMERTableOutput infers how to parse the output, rather receiving explicit parsing specifications**

The format of files output by hmmer will depend upon 3 factors:

1. the program used (`hmmsearch`, `hmmscan`, `nhmmscan` etc)
2. the specified table format (`--domtblout`, `--tblout`, `--pfamtblout`)
3. the alphabet used (`AA`, `RNA`, `DNA`)

But the current parser only takes the program and alphabet used, and does not explicitly know which table format is used.

To properly parse the output file, the parser instead receives a `context` string, which is used throughout the class with many `if context == ...` conditionals which rely on upstream assumptions of what `context` means. As the number of use cases of the parser increases, this is not an extendable design, and inevitably leads to the spaghetti code we are beginning to see within the parser.

[...]

Issue #1564, opened 2020-11-22
[BUG] Multithreading issue with anvi-run-hmms


meren:
## Short description of the problem

anvi-run-hmms works well on a contigs database as a single thread, but fails when running in multiple threads.

## anvi'o version

This happens with the main branch.

## Detailed description of the issue

Running it with a **single thread**, everything is fine:

```
anvi-run-hmms -c CONTIGS.db -I Ribosomal_RNA_23S

HMM sources ..................................: Ribosomal_RNA_23S
Alphabet/context target found ................: RNA:CONTIG

HMM Profiling for Ribosomal_RNA_23S
===============================================
(...)
Number of raw hits ...........................: 8
Number of weak hits removed ..................: 0
Number of hits in annotation dict  ...........: 8
Pruned .......................................: 3 out of 8 hits were removed due to redundancy
Gene calls added to db .......................: 5 (from source "Ribosomal_RNA_23S")

✓ anvi-run-hmms took 0:01:47.840152
```

Running it with **multiple threads**:

```
anvi-run-hmms -c CONTIGS.db -T 4 -I Ribosomal_RNA_23S --just-do-it

HMM sources ..................................: Ribosomal_RNA_23S
Alphabet/context target found ................: RNA:CONTIG

HMM Profiling for Ribosomal_RNA_23S
===============================================
(...)
Number of raw hits ...........................: 8

✖ anvi-run-hmms encountered an error after 0:00:45.813513


Config Error: Number of column names declared (15) differs from the number of columns found
              (16) in the matrix
              ('/var/folders/x5/gt4031w53fs63csv1fp0r_3w0000gn/T/tmpe_h6eje3/hmm.table') :/

```

Indeed, there is something wrong with the `hmm.table` file that is produced. This is how it looks with **single thread**:

```
cat /var/folders/x5/gt4031w53fs63csv1fp0r_3w0000gn/T/tmp2sl4ixmm/hmm.table
23S_rRNA_bac	-	215_10M_human_filteredcontig37	-	3	1185	1619	2816	1617	2818	2906	+	2.7e-278	913.6	27.9
23S_rRNA_bac	-	215_10M_human_filteredcontig414	-	3	1265	1270	4	1272	1	2906	-	1.1e-258	848.4	40.
23S_rRNA_bac	-	215_10M_human_filteredcontig422	-	4	2111	2097	3	2100	1	2906	-	0	1567.7	63.
[...]

PR #1709, opened 2021-04-08
Fix hmmer parsing

ivagljiva:
This PR addresses the bug reported in #1564, in which HMMER parsing was failing in certain cases where the alignment of space-delimited HMMER output files was off in certain lines.  

**Brief description of the bug**
In rows where some of the columns contained shorter text or numbers than the previous rows, the whitespace alignment in raw HMMER output is shitty. Here is one example:
image

Previously, the `anvio.drivers.HMMer` class was pre-processing the HMMER output files to get rid of the 'description' field, because that field internally contains spaces and therefore cannot be directly converted into a tab-delimited format where each row has a consistent number of columns. In edge cases like the one shown above, the attempted removal of the 'description` field itself was producing inconsistent numbers of columns. This was causing `anvi-run-kegg-kofams` as well as `anvi-run-hmms` to fail at the parsing step.

**Brief description of the fix**
The first thing I did was to remove pre-processing from the `anvio.drivers.HMMer.append_to_main_table_file()` function which concatenates raw HMMER output from multiple threads, with the exception of removing header lines (as suggested by @ekiefl in #1670). So now that class does not attempt to remove the description field or convert spaces to tabs, and the parser class receives essentially raw, space-delimited HMMER output, just without the comment lines.

Because we must wait until [HMMER4 is released to have tab-delimited output](https://github.com/EddyRivasLab/hmmer/issues/235) that works nicely with our current parsing implementation, my solution was to add a function to the `anvio.parsers.HMMERTableOutput` class which uses pandas to read the file by splitting on spaces. In the resulting dataframe, the space-containing 'description' field may be split up into multiple columns; we combine up to six of these columns into one 'description' column and then print the dataframe to a new tab-delimited file that can then be parsed as usual. 

**Some considerations**
This solution does produce yet another intermediate, temporary file. Perhaps a more elegant way would be to write a parser from scratch instead of generating a tab-delimited file and then sending that to `get_TAB_delimited_file_as_dictionary()`. But in the hopes that we will eventually get tab-delimited output directly from HMMER, I did it this way to save time at the expense of tmp disk space. I'm happy to change it if anyone has strong opinions against this strategy :)

Special thanks to @michaelkyu and @ekiefl for discussing the speed and memory implications of using pandas with me. I have their assurances that it will be quite fast and efficient. @michaelkyu even wrote a much faster implementation than what I had originally (thanks, Mike!).

I have not been able to test this with `anvi-run-kegg-kofams` on very large metagenomes which yield hundreds of thousands of KOfam hits. I do not anticipate that it will be a problem, but I wanted to explicitly mention it in case I am wrong, so we all have someone to blame if things blow up later. :)

**Testing**
I tested this implementation on genomes that were failing prior to the fix, (as well as a few genomes that were doing fine), and everything worked perfectly. I tested with different numbers of threads, too.

To make sure that my changes don't produce different results from what we were getting before, I took a test genome and compared HMM hits that we get with `master` and with this branch.
```
## git checkout master
anvi-run-hmms -c GCF_000725495.1-contigs.db --tmp-dir HMMS_GCF_000725495.1_MASTER/ --just-do-it -T 1
sqlite3 GCF_000725495.1-contigs.db "select * from hmm_hits" | tr '|' '\t' | cut -f 2- >> GCF_000725495.1_master_HMM_1T
anvi-run-hmms -c GCF_000725495.1-contigs.db --tmp-dir HMMS_GCF_000725495.1_MASTER/ --just-do-it -T 4
sqlite3 GCF_000725495.1-contigs.db "select * from hmm_hits" | tr '|' '\t' | cut -f 2- >> GCF_000725495.1_master_HMM_4T
## git checkout fix_hmmer_parsing
anvi-run-hmms -c GCF_000725495.1-contigs.db --tmp-dir HMMS_GCF_000725495.1_FIXED_2 --just-do-it -T 4
sqlite3 GCF_000725495.1-contigs.db "select * from hmm_hits" | tr '|' '\t' | cut -f 2- >> GCF_000725495.1_fixed_HMM_4T
anvi-run-hmms -c GCF_000725495.1-contigs.db --tmp-dir HMMS_GCF_000725495.1_FIXED_2 --just-do-it -T 1
sqlite3 GCF_000725495.1-contigs.db "select * from hmm_hits" | tr '|' '\t' | cut -f 2- >> GCF_000725495.1_fixed_HMM_1T
```
There were no differences between `master` and this branch, and the only differences between single- and multi-threaded runs on the same branch were two gene calls that swapped their 23S rRNA annotation:
image

I did a similar test for KOfams. 
```
## git checkout master
anvi-run-kegg-kofams -c GCA_000007185.1-KOFAM-WORKS-contigs.db -T 4 --just-do-it
sqlite3 GCA_000007185.1-KOFAM-WORKS-contigs.db "select * from gene_functions where source='KOfam'" | tr '|' '\t' | cut -f 2- > GCA_000007185.1_master_KOfam
## git checkout fix_hmmer_parsing
anvi-run-kegg-kofams -c GCA_000007185.1-KOFAM-WORKS-contigs.db -T 4 --just-do-it
sqlite3 GCA_000007185.1-KOFAM-WORKS-contigs.db "select * from gene_functions where source='KOfam'" | tr '|' '\t' | cut -f 2- > GCA_000007185.1_fixed_KOfam
```

And this is the result of the `diff`:
image

As you can see, the differences are due to slight rounding error in the e-values. I don't think it is a problem. :)


meren:
Thank you very much for looking into this, @ivagljiva!

All looks good to me. These changes will serve us well for a long time :)

Issue #1648, opened 2021-01-22
[BUG] Anvi'o states are not flexible enough


meren:
## The need

A user defined minimal state should work well when imported, even if it does not describe the entire visual interface exhaustively.

## The solution

The interactive interface should take in a state file with a grain of salt. If a piece of information exists in it, it should use it. But if something is missing, it shouldn't give an error and use the defaults instead.

# Example

The archive attached contains the following key files:

* `default-orig.json`: A working JSON file exported using `anvi-export-state`.
* `default-mini.json`: A minimal version of the `default-orig.json`.
* `PROFILE.db`: A profile database that contains `default-mini.json` as its default state.

Running this command does not work:

```
anvi-interactive -p PROFILE.db -c CONTIGS.db
```

Because this is how the minimal state looks like, and many of the state variables are not defined in it (to see the difference, please compare it to the `default-orig.json`):

``` json
{
    "version": "3",
    "angle-max": "180",
    "layers": {
        "gc_content": {
            "type": "intensity",
            "color": "#ff0000",
            "color-start": "#e8e8e8"
        }
    },
    "samples-layers": {
        "default": {
            "percent_mapped_reads": {
                "color": "#00ff00"
            }
        }
    }
}
```

Running this command first overwrites the default state with the original one, and works:

```
anvi-import-state -s default-orig.json -p PROFILE.db -n default
anvi-interactive -p PROFILE.db -c CONTIGS.db
```

One can re-import the minimal state this way to test it:

```
anvi-import-state -s default-mini.json -p PROFILE.db -n default
anvi-interactive -p PROFILE.db -c CONTIGS.db
```

## Beneficiaries

Those who wish to generate minimal state files programmatically to mark things in the interactive interface :)

## Archive for tests

[MINITEST.tar.gz](https://github.com/merenlab/anvio/files/5858191/MINITEST.tar.gz)

matthewlawrenceklein:
@meren 

after doing some manual debugging in the `processState()` function in `main.js`, I believe the root issue here is that `processState()` currently calls `buildSamplesTable()`, passing it the 'samples layer order' and 'samples layers' from state, whether or not they exist. Making this call conditional upon the existence of the above state items now allows the interactive interface to continue processing, and it no longer hangs on 'loading'.

[...]

PR #1715, opened 2021-04-14
1648 state flexibility

matthewlawrenceklein:
This PR aims to resolve the issue mentioned in bug report #1648, wherein user-supplied incomplete state data caused the interactive interface to hang. 

## The background
- upon launching the interactive interface, anvio will check to see if state data has been supplied. If no state data is provided, the interface will not render a phylogram until the user initiates the `draw()` method, at which point the `serializeSettings` method is called to create the default values needed for a phylogram to be successful rendered. 
- if user-supplied state data is found, the `processState` method is called to conditionally check each expected element of state, and set those values to the user-supplied values if present. 
## The bug 
- the `processState` method makes a few function calls under the assumption that it has been provided a complete set of state data. In the case of the original issue #1648, `processState` tries to pass `samples-layers` and `samples-layers-order` to `buildSamplesTable` even if these values do not exist, causing a crash. While not all possible state values are required for anvio to function successfully (for example, `max-font-size`) There are a few that need both exist and be complete for the interface to render meaningful data. 
## The fix 
- At present the solution is to combine the functionality of `serializeSettings` and `processState`. 
- If user supplied state data is found, `processState` will first create a `serializedStateObj`, set as the return value of `serializeSettings`. 
- processState then checks to see if the user-supplied state data contains all the required values. If a value is missing, it is set to the `serializedStateObj` counterpart. If the value exists but is different from the serialized counterpart, then the `traverseNestedData` method is called to repeat the above process for each level of nested state data. 
- The user is warned via toastr message that their state data has been altered. They can optionally download the updated state as json to compare. 

## The outcome
- Users can now provide incomplete and deeply nested state data. Anvio will supplement the user's data with all the pieces necessary for a successful render, while maintaining the user-supplied values.


meren:
This looks very good, @matthewlawrenceklein! I can't wait to test play with it in the context of `anvi-display-functions`. I will merge it now so we can see if it explodes for same rare cases :)

I think it would have been even more amazing if the state doc included some tips for the programmers regarding the fact that anvi'o can deal with incomplete states. For instance, what would they need to do if all they want to do is to set a specific color to a handful of genes? what would they need to do to start the interactive interface with a phylogram display rather than a circular plot? When we don't leave behind some traces in the documentation, people never realize the existence of such amazing features.

PR #1716, opened 2021-04-15
tRNA-seq indels

semiller10:
This pull request includes a number of changes to the tRNA-seq workflow.

As the name of the branch indicates, the biggest additions involve the detection of modification-induced deletions. The mutational signature of modifications is most frequently substitutions. Deletions are much less common, but are much more common than insertions, which are ignored. Certain modifications, such as queuosine, preferentially leave deletions. After identifying tRNA sequences with potential modification-induced substitutions, the workflow introduces in silico deletions around substitution sites, forming query sequences that are searched against other target sequences to efficiently find deletions supported by the data. As in many other aspects of the workflow, the user can parameterize most aspects of this process, including the size and frequency of in silico deletions.

This pull request fixes the de novo tRNA profiler's handling of the D arm in Type II tRNA (long V loop). There is also a fix to the detection of the extra 5'-G in tRNA-His.

A glossary of terms was added to the `trnaseq` module that orients the user or developer to the workflow's model of the tRNA-seq experiment, e.g., unique, trimmed, normalized, and modified sequences. The aforementioned sequence types were further developed through subclassing, as different methods of finding these sequences, such as tRNA feature profiling and mapping fragments to profiled tRNAs, produce objects with different attributes. Sequence objects are now referenced by dictionaries keyed by representative name rather than lists.

Certain abstruse sources of error in tRNA prediction were targeted for removal. When a cluster of 3'-dereplicated trimmed sequences are condensed into a normalized sequence, but multiple of the trimmed sequences were found to have complete profiles, then the shortest complete profile is chosen as correct. The longer sequences therefore have extra 5' nucleotides, and so are collapsed into the shorter completely profiled sequence.

To speed up fragment mapping, the relatively large pool of 3' "fragments" that lack a 3' acceptor variant is 3'-dereplicated from profiled tRNA sequences in a new method.

This pull request corrects an error that slipped through in the formation of modified sequences. The same modified sequence was sometimes forming multiple times due to the aggregation of the same normalized sequence with different longer normalized sequences.

Finally, some changes were made in the Snakemake workflow to accommodate compressed files, which were not handled adeptly.


meren:
A lot of work. Thanks again, @semiller10. Feel free to merge it to the main branch whenever you feel like it.

PR #1721, opened 2021-04-19
patch for state processing, skipping layers and removing scary toastr warns

matthewlawrenceklein:
This is a quick patch to remedy the immediate issues caused by new state processing methods. The main issues addressed here are 
- Anvio's default generated state doesn't provide `layers` names, creating an inherent mismatch against user's `layers` data and causing user's `layers` data to be ignored. This `layers` portion of `processState()` has been removed for now. 

- Current Toastr warning messages relating to state are overly dramatic and potentially wrong. Currently anvio warns users when missing data has been added to state, or when certain elements of user state are compared against anvio's generated state and found to be different. My working assumption had been that if these state elements were different, it was because the user's state data was incomplete. In retrospect it is often the case that user's state elements may contain **more** data than the generated default, and users should not be warned of anything if that's the case. Toastr warnings have been disabled in this patch.

### DIFF
--- a/anvio/data/interactive/js/main.js
+++ b/anvio/data/interactive/js/main.js
@@ -2464,7 +2464,8 @@ function processState(state_name, state) {
         modifiedItems.push('views')
     }
 
-    if (state.hasOwnProperty('layers') && state['layers'] === serializedState['layers']) { //check if user provides incomplete data against serialized data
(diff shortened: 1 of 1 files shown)


meren:
No commented out lines in committed code blocks please :)

meren:
let's merge this so @FlorianTrigodet can test it. clearly we need more tests to ensure that this very fancy functionality does not interfere with existing states to perform as expected :)

Issue #1717, opened 2021-04-16
Missing submodules in cloned repository prevents interactive interface to initiate


microDM:
## Short description of the problem

anvio's interactive stuck at still loading

## anvi'o version

Anvi'o .......................................: hope (v7-dev)

Profile database .............................: 35
Contigs database .............................: 20
Pan database .................................: 14
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 2
tRNA-seq database ............................: 1

```
anvi-self-test --version
```

## System info

I am using Pop OS 21.10. I have install anvio using conda

## Detailed description of the issue

I have used following command after following anvio protocol of pangenome

`anvi-display-pan -p Burkh_Pan-PAN.db -g Burk-GENOMES.db`

 ## The output is

`Interactive mode .............................: pan                                                                                                    
Functions found .............................................: COG20_FUNCTION, COG20_CATEGORY, COG20_PATHWAY                                           
Genomes storage .............................................: Initialized (storage hash: hash1a269245)                                                
Num genomes in storage ......................................: 143
Num genomes will be used ....................................: 143
Pan DB ......................................................: Initialized: Burkh_Pan-PAN.db (v. 14)
Gene cluster homogeneity estimates ..........................: Functional: [YES]; Geometric: [YES]; Combined: [YES]
                                                                                                                                                       
* Gene clusters are initialized for all 28145 gene clusters in the database.                                                                                                                                        
WARNING
==============================================================
This pangenome (which you gracefully named as 'Burkh_Pan') does not seem to have
any hierarchical clustering of gene clusters it contains. Maybe you skipped the
clustering step, maybe anvi'o skipped it on your behalf because you had too many
gene clusters or something. Anvi'o will do its best to recover from this
situation. But you will very likely not be able to see a hierarchical dendrogram
in the resulting display even if everything goes alright.. In some cases using a
parameter like `--min-occurrence 2`, which would reduce the number of gene
clusters by removing singletons that appear in only one genome can help solve
this issue and make your pangenome great again.                                                                                                                            
The server is up and running 🎉
WARNING
===============================================
If you are using OSX and if the server terminates prematurely before you can see
anything in your browser, try running the same command by putting 'sudo ' at the
beginning of it (you will be prompted to enter your password if sudo requires
super user credentials on your system). If your browser does not show up, try
manually entering the URL shown below into the address bar of your favorite
[...]

PR #1724, opened 2021-04-26
1719 submodule check

matthewlawrenceklein:
## What 
A basic error messaging framework for dependency load errors for the interactive interface

## Why 
As noted in #1717 , the interactive interface does not currently provide any error messaging prior to all dependencies being loaded, resulting in the user being greeted with an endless 'loading' screen with no context. 

## How 
Anvio attempts to load submodule dependencies in `index.html` before loading any javascript files. A new js file `error.js` has been created (and loaded prior to the submodule calls), which contains the `alertDependencyError` method that each submodule call points to for its `onerror` attribute. `alertDependencyError` keeps track of the total number + names of the failed dependencies, and when the final synchronous submodule call is made, `alertDependencyError` is passed a special argument `isFinalDependency` for either the `onerror` or `onload` attribute, triggering an alert notification for the user. 

## Test
In `index.html`, changing any of the submodule `src` paths and refreshing the interactive interface page wil trigger the alert messaging.

### DIFF
--- a/anvio/data/interactive/index.html
+++ b/anvio/data/interactive/index.html
@@ -20,7 +20,7 @@
     <script type="text/javascript" src="lib/bootstrap/dist/js/bootstrap.min.js"></script>
     <link rel="stylesheet" href="css/main.css" type="text/css" />
 
-    <script type="text/javascript" src="lib/jquery-svgpan/jquery-svgpan.js"></script>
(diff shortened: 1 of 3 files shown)


meren:
Thank you! :)

meren:
this branch has been merged, but additional commits that are not merged. what is the situation with this, @matthewlawrenceklein?

PR #1728, opened 2021-04-28
state flexibility bug fixes 

matthewlawrenceklein:
## What
This PR addresses a few major + minor bugs first introduced by PR #1715 , seeking to make anvio's state processing more adaptable. 

## Why 
It turned out no one liked the bugs, so we decided to get rid of them. 

## How 
- A major bug in PR #1715 was an inability to correctly process user's `layers` data from their state file. This was because of a naming mismatch generated by anvio's `serializeSettings` method. Luckily, `serializeSettings` accepts a boolean argument for a `use_layer_names` parameter (thanks Ozcan!). Now included in the function call, we're able to directly compare serialized and user `layers` data as expected. 
- Previously, user-provided `views` data was ignored by `processState`, even though that data was correctly persisted in exported state files. This was due to the exclusion of a helper process that iterates through `views` and calls `getLayerId()` for each view. In #1715 that helper process was only called when user's `views` data exactly matched the equivalent serialized data. The solution was to move this process to its own method `setViews`, which is now called when user's `views` data is missing entirely, or simply different from the serialized data. 
- Here's a fun one, thanks @FlorianTrigodet . Previously, any elements of the user's state data that were set to zero ('0') were overridden by `processState`. This was due to a bug in the `traverseNestedData` method, which was set to rep[ace any falsey (initially intended to mean 'missing') elements of user state with the serialized parallel. Confusingly (to me!), Javascript casts the `string '0'`to `int 0`, which returns falsey. The `traverseNestedData` logic has been adjusted to only replace `null` or `undefined` elements. The moral of the story - in javascript, an empty string is falsey and a non-empty string is truthy, unless the non-empty string is '000000', in which case it is then cast as falsey. Neat! 

## Testing 
Using the infant gut dataset (or any other), boot up an interactive interface session. Make any changes you'd like, and save them to a new state. Kill the server, run `anvi-export-state` to export your new state to json. Open up the json, remove any elements you'd like, from whole objects like `layers` to individual data points like min/max values of `views`. Save those changes, import your new state via `anvi-import-state`, and start a new interactive session. Load your edited state, and anvi'o should fill in the cracks : )


meren:
Thanks, @matthewlawrenceklein. Let's merge it to the main branch and see if people find other things.

Issue #1726, opened 2021-04-26
[BUG] Uniform bin name input + display values 


matthewlawrenceklein:
## The need

anvi'o currently allows users to set bin names with spaces (ie 'my special bin')
anvi'o currently does an iffy job stripping underscores from bin names when displayed
![image](https://user-images.githubusercontent.com/62408216/116136621-24593d00-a698-11eb-9be2-739426e500bd.png)

## The solution

strict pattern matching to replace all spaces in user-set bin names with underscores (while also showing the user these changes so they can see the correct naming convention)
strict pattern matching to replace all underscores with spaces when displaying bin names

## Beneficiaries
anyone with an appreciation for standardized naming conventions and prettified display values

matthewlawrenceklein:
addressed in PR #1729

PR #1729, opened 2021-04-29
uniform bin naming + display

matthewlawrenceklein:
This PR addresses issue #1726 in two ways: 

- displayed bin labels now replace **all** underscores with spaces, instead of just the first underscore. 
- user-supplied bin names are now coerced into a naming convention that replaces all spaces with underscores. 

Raising the question, 'if underscores are added silently on input, and removed silently upon display, do they still make a sound?". Only time will tell.


meren:
Thank you for this, @matthewlawrenceklein :)

PR #1741, opened 2021-05-17
Inverse bug fix

matthewlawrenceklein:
This PR addresses a bug when utilizing inverse-shading in rectophylograms, as caught by @watsonar (thanks!). The bug caused inverse-shaded segments to vary in length based on branch height.

The solution is to calculate the largest x value across all nodes, and apply that value for each shaded area when drawn.


meren:
Thank you both! :)

PR #1752, opened 2021-06-09
`anvi-run-hmms --get-domtable-output` input

mschecht:
Hi Iva, 

I think this flag for `anvi-run-hmms` does not require any input string. I ran `run_all_tests_.sh` and everything checks out but wanted your opinion first before committing to `main`

Thanks!

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2253,13 +2253,6 @@ D = {
                      "This way you can run HMM profiles that are not included in anvi'o. See the online "
                      "to find out about the specifics of this directory structure ."}
                 ),
-    'domtblout': (
-            ['-F', '--domtblout'],
-            {'metavar': 'HMM OUTPUT FORMAT',
-             'type': str,
-             'help': "This flag will provide the domain hits table (protein search only) from hmmsearch to a specified path."
-                     "The output file will contain the suffix '_domtable.txt'."}
-                ),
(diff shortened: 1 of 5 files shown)


meren:
While we are on this topic, can we also do something about the fact that there are two very different parameters for `anvi-run-hmms` and `anvi-script-filter-hmm-hits-table`:

https://merenlab.org/software/anvio/help/main/programs/anvi-script-filter-hmm-hits-table/

And this current flag does not even match to either mentioned in the documentation 😂 

Perhaps the flag for both could be `--domain-hits-table`, which is what it really is, and in `anvi-run-hmms` it would be used to report the table, in `anvi-script-filter-hmm-hits-table` it would be expected as an input to filter hits.

The docs and snakemake workflows will likely be updated with such a change, too.

mschecht:
> Perhaps the flag for both could be --domain-hits-table, which is what it really is, and in anvi-run-hmms it would be used to report the table, in anvi-script-filter-hmm-hits-table it would be expected as an input to filter hits.

That's a great point @meren , I addressed this in the most recent commits. I also ran `run_all_tests.sh` and think this is ready for merging.

ivagljiva:
Sorry for the delay @mschecht! These changes look good to me :)

mschecht:
Thanks @ivagljiva!

meren:
Thank you both!

PR #1753, opened 2021-06-15
Refactoring anvi'o view tables

meren:
This PR proposes a complete refactor for one of the most annoying early design decisions made in anvi'o codebase in 2014 (the passive voice is passive).

This design shortcoming prevented anvi'o to merge more than 2,000 samples, and affected at least @efogarty11, @watsonar, @ekiefl, @tdelmont, and probably many others.

The PR carries out a significant change in the structure of the 'view' tables of anvi'o. It not only removes this limitation, but also results in significant speed and memory gains during `anvi-merge`. These changes require pan and profile databases to advance their version numbers, and the PR includes migration scripts for both database types. I have tested these migration scripts with anvi'o pan and profile databases from 2014, as well as recent projects. Everything seems to be working.

This PR also removes the following two tables from anvi'o profile databases:

* `max_normalized_ratio`
* `relative_abundance`

The reason to this is two folds: one, preparing data for these tables required us to read all view data from all single profiles into memory during merge. This perhaps could have been fixed with a better design, but I decided to not invest time into that since I don't think these tables have been useful to anyone. If there are protests in the future, we can find a smarter way to prepare those data from scratch.

This took 2.8 days of work, and I think it was worth it. But it will require us to test it rigorously, by first merging a project that has like 5,000 profiles :)

Best,


anvio/migrations/profile/v35_to_v36.py
```diff
@@ -0,0 +1,177 @@
...
+        # さようなら
+        profile_db.disconnect()
+
+    else:
+        ###########################
+        #     SURPRISE PROFILE    #
+        ###########################
+
+        raise ConfigError("Anvi'o is confuse. Your profile database does not fit into anything we have "
+                          "anticipated to run into here. For full disclosure, ")
```
ivagljiva, line 147: I think this sentence needs an end :p

meren: lol

ivagljiva:
It looks so much cleaner than before, and I have no doubt the new performance is impressive. The anvi'o world thanks you @meren :)

meren:
Thanks for taking a look, Iva!

anvio/migrations/profile/v35_to_v36.py
meren: 5a775d1724654ec8ff623e309718b79d73782082 addresses this. I will merge this repo soon to make sure we start testing it.

ivagljiva: yay :)

btw, this migration script is refreshingly full of easter eggs and it makes me very happy :)
> баяртай
>さようなら
> if splits, I sits

things like this make code so much more enjoyable to read ❤️

meren: hehe :) I'm glad that someone is reading the code :)

ekiefl:
I am late to the party but just looked through the code and it looks great. 👍 

One thing I want to mention was the placement of `get_view_data` in our `db.DB` class. Ideally, `db.DB` should only contain methods that pertain to all database types, so as not to become a dumping grounds for stray methods. In this sense `get_view_data`'s placement kind of reminds me of `init_gene_calls_dict`'s placement in our `tableops.Table` base class, which should really be placed in some inheriting class rather than at the bottom of the inheritance tree. I don't this deserves remediation but its just something that caught my eye.

meren:
Thanks for the input, @ekiefl!

> One thing I want to mention was the placement of get_view_data in our db.DB class. Ideally, db.DB should only contain methods that pertain to all database types, so as not to become a dumping grounds for stray methods.

I thought a lot about this also. Although I wasn't sure how to handle this, I did it this way since I couldn't find a much better solution. I looked into `tableops.Table`: the current implementation of `get_view_data` enables quick access to it from any `DB` instance while putting it in `tableops.Table` was going to require that class to be inherited from places where there is no inheritance. But I generally agree that we should keep anything that makes `db.DB` less modular across very different projects beyond anvi'o.

I will try to come up with a better solution. Fortunately it will not require any migration and can be done seamlessly.

PR #1759, opened 2021-06-28
Vmatch integration

semiller10:
This pull request includes multiple major changes to the tRNA-seq workflow.

## Vmatch
A [Vmatch](vmatch.de) driver was added to anvi'o. Vmatch performs fast, large-scale, *comprehensive* sequence alignments, working best with similar sequences. Unlike Bowtie, Vmatch can report every possible alignment given the parameterization. This is important in the tRNA-seq workflow, which considers both specific and non-specific tRNA coverage due to the abundance of short fragments.

The addition of Vmatch significantly reduced multiple bottlenecks in the workflow:

1. Mapping unprofiled sequences to the interior of profiled tRNA to identify non-3' fragments (exact substring matches)
2. Aligning normalized tRNA sequences to each other as part of agglomerative clustering in the search for modification-induced substitutions (end-to-end alignment with mismatches but not indels)
3. Aligning sequences to find modification-induced indels (end-to-end alignment with indels but not substitutions)

These bottlenecks were crippling with large tRNA-seq libraries of millions of high-quality reads. These bottlenecks are reduced by 2-3 orders of magnitude for large libraries (saving hours if not days of time) and less for smaller libraries.

The indel search was completely overhauled. Now insertions as well as deletions are identified. The highly parameterized deletion model used before was replaced by the still complicated but more agnostic and straightforward alignment approach.

## Sequence objects
Sequence objects, such as `ModifiedSequence` and `NormalizedSequence`, had attributes referencing other sequence objects. For example, `NormalizedSequence` referenced its component `TrimmedSequence` objects and any `ModifiedSequence` objects that contain it. This produced a subtle but critical flaw in the checkpoint intermediate pickle files that store objects at different stages of the workflow. Dicts of seq objects, such as all `ModifiedSequence` objects and all `NormalizedFullProfileSequence` objects, are stored separately -- if they were bundled into a single object, say a list of dicts, it would often exceed the size of objects that can be passed through a pipe in pickling. When these dicts are loaded back into memory, however, they contain seq objects that do not reference each other! If Dict 1 contains Object A, Dict 2 contains Object B, and Object A contains a reference to Object B, then after reloading the dicts, Object A in Dict 1 would reference an Object B at memory location X and Dict 2 would reference Object B' at memory location Y. Any changes to Object B would not have no bearing on Object B'. This defeats the purpose of checkpoints that restore the previous state of the workflow.

This problem was solved by removing cross-references between sequence objects and instead referencing the names of these objects. The only pointers to the sequence objects exist in the dicts themselves: Dict 1 has one reference to Object A, Dict 2 has one reference to Object B, and there are no other references to these objects outside these dicts.

The number of other mutable objects in sequence objects was also minimized to save memory during the workflow and disk space in the intermediate files.

Lastly, the names of different categories of sequences were simplified and systematized in the code. This is explained in the Abbreviations section under the Glossary in the `trnaseq` module. For example, a trimmed sequence with a full tRNA feature profile (a `TrimmedFullProfileSequence` object) is now called `seq_Tf`.

## Terminal messages
Very little information was previously written to the terminal during runtime. Now, a battery of statistics is calculated after each stage of analysis and reported to the terminal.

Progress during different stages was made more informative and excessive workflow parameters allowing the user to specify progress intervals were removed.


meren:
After the commits above, the branch runs the tests well. But please double check if everything is working as you expect, @semiller10.

As far as I am concerned, you're good to go to merge this awesome work to the development branch. My only comment is this:
 
There seems to be a lot of zombie variables that were perhaps used at some point, but they no longer seem to be relevant. Cleaning up could be a good idea for the longevity of the module which is now pushing 8,000 lines of code o_O.

semiller10:
I hunted down and killed the zombies. Half of them were trivial. Others were messier.

Issue #1682, opened 2021-02-24
[BUG] Replace this text with a short but descriptive title


TonyMane:
## Short description of the problem

Getting the following: ModuleNotFoundError: No module named 'sklearn.ensemble._forest'. 
I tried installing this with pip, which says it can find it. This error occurs when doing 'anvi-self-test', however, it also occurs while doing 'anvi-gen-genomes-storage'. 

## anvi'o version

(anvio-master) (anvio-master) -bash-4.2$ anvi-self-test --version
Anvi'o .......................................: hope (v7-dev)

Profile database .............................: 35
Contigs database .............................: 20
Pan database .................................: 14
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 2
tRNA-seq database ............................: 1



```
anvi-self-test --version
```

## System info

Please tell us which operating system you are using, and how did you install anvi'o.

## Detailed description of the issue
:: Firing up the interactive interface ...

Traceback (most recent call last):                                                                                                                                                                                                                                                                                                        
  File "/home/abertagnolli3/github/anvio/bin/anvi-interactive", line 122, in <module>
    d = interactive.Interactive(args)
  File "/home/abertagnolli3/github/anvio/anvio/interactive.py", line 198, in __init__
    self.completeness = Completeness(self.contigs_db_path)
  File "/home/abertagnolli3/github/anvio/anvio/completeness.py", line 45, in __init__
    self.SCG_domain_predictor = scgdomainclassifier.Predict(argparse.Namespace(), run=terminal.Run(verbose=False), progress=self.progress)
  File "/home/abertagnolli3/github/anvio/anvio/scgdomainclassifier.py", line 234, in __init__
    SCGDomainClassifier.__init__(self, args, run, progress)
  File "/home/abertagnolli3/github/anvio/anvio/scgdomainclassifier.py", line 73, in __init__
    self.rf.initialize_classifier()
  File "/home/abertagnolli3/github/anvio/anvio/learning.py", line 103, in initialize_classifier
    classifier_obj = pickle.load(open(self.classifier_object_path, 'rb'))
ModuleNotFoundError: No module named 'sklearn.ensemble._forest'


Config Error: According to the exit code ('1'), anvi'o suspects that something may have gone
              wrong while running your tests :/ We hope that the reason is clear to you from
              the lines above. But if you don't see anything obvious, and especially if the 
[...]

PR #1761, opened 2021-06-29
small requirement fix in meta.yaml of anvio-minimal

moritzbuck:
sklearn requirement is wrong, or at least was generating error (see https://github.com/merenlab/anvio/issues/1682).
Proposed fix should solve this...


meren:
Thank you very much for this, @moritzbuck!

moritzbuck:
You very welcome @meren. Very small stuff, and now I have to worry that I broke yours ;)
But at least I am learning how to use github for collab' stuff!

PR #1767, opened 2021-07-12
added anvi-script-compute-bayesian-pan-core to the sandbox

moritzbuck:
After only three attempts made the conda package of mOTUlizer friendly enough with the current anvio conda package, so I feel finally comfortable to push here my anvio script (curated and edited by @meren). 

It computes a likelihood for each gene cluster of being either in the core or in the accessory and the ratio of these can be used to partition all gene-clusters into the core-genome and the accessory genome accordingly.

### DIFF
--- a/sandbox/anvi-script-compute-bayesian-pan-core
+++ b/sandbox/anvi-script-compute-bayesian-pan-core
@@ -1,0 +1,185 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+
+import sys
+
+import anvio
+import anvio.dbops as dbops
+import anvio.terminal as terminal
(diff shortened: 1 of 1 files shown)


meren:
Thank you very much for this awesome contribution, @moritzbuck! :)

Going forward there are a few things that need to happen, and I will list them here to remember:

- [ ] Add Moritz to the AUTHORS.txt
- [ ] Update anvio/docs with an `anvi-script-compute-bayesian-pan-core.md`.
- [ ] Include `anvi-script-compute-bayesian-pan-core` in pan tests? TBD.
- [ ] Include `anvi-script-compute-bayesian-pan-core` in the pangenomics tutorial.

Best,

PR #1771, opened 2021-07-19
Interactive interface error messaging framework + landing page

matthewlawrenceklein:
## What
A zero-dependency landing page and notification framework that alerts the user when a relevant error is encountered, and provides helpful troubleshooting hints and links to additional resources. 

## Why
This PR seeks to provide clarity and context in cases where the interactive interface may fail silently, or fail with limited logging in the Chrome console. 

## How 
The `displayAlert()` method in `error.js` can be hooked into the codebase anywhere with a `try... catch` block. A conditional block parses the error log into something a little more contextual, and then the `error-landing.html` page maps over the `issueCatagories` object to render out error/troubleshooting sets.

### DIFF
--- a/anvio/data/interactive/css/error.css
+++ b/anvio/data/interactive/css/error.css
@@ -1,3 +1,7 @@
+* {
+    font-family: 'PT Serif',serif;
+}
+
 #body {
     display: flex;
     justify-content: center;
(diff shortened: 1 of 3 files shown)


meren:
Thank you, @matthewlawrenceklein!

Issue #1770, opened 2021-07-19
[FEATURE] Improvements + fixes for support value rendering


matthewlawrenceklein:
## The following fixes should be implemented
- Branch support symbols are shifted right in Phylogram view, the should be centered as in circlephylogram view
![shifted support](https://user-images.githubusercontent.com/62408216/126182018-cea13f36-d657-4026-90bb-5958ceb3da5f.png)

- In phylogram view, support values should be rotated 90deg for legibility
![rotate support values](https://user-images.githubusercontent.com/62408216/126182186-de9cb9d4-922f-4c05-a085-5c3db632eab2.png)

- Display range UI inputs should have max and min vals of 0 and 1.0 respectively
![range](https://user-images.githubusercontent.com/62408216/126182372-cdc8cb15-c86c-4283-8e8f-7349c9aa18df.png)

- A new UI input for users to select 'floating precision', and an associated method to appropriately truncate support values. 
![floating precision](https://user-images.githubusercontent.com/62408216/126182728-cb82fe21-88f4-40d0-98c4-f424e21a00a9.png)

PR #1772, opened 2021-07-21
1770 support value fixes

matthewlawrenceklein:
this PR addresses fixes/feature noted in issue #1770 

## fixes
- support value symbols now render in the correct place when tree == phylogram (previously, they had shifted to the next node to the right)
- for legibility, support value text is shifted 90deg when tree = phylogram
- Display value range is now set with min/max val of 0.0 and 1.0, respectively

## features
- a new UI element, 'floating point precision' is available for users to set, with the corresponding value being implemented upon render
- this value is hooked into the interactive interface state

### DIFF
--- a/anvio/data/interactive/index.html
+++ b/anvio/data/interactive/index.html
@@ -298,6 +298,12 @@
                                             <input type="number" id="support_font_size" style="width: 60px;" value='15' max='100' min='1'>
                                         </div>
                                     </div>
+                                    <div class="form-group form-margin" id='support_text_rotation'>
+                                        <label class="col-md-4 settings-label" for="support_text_rotation" data-help="support-text-rotation">Text Rotation:</label>
+                                        <div class="col-md-8">
+                                            <input type="number" id="support_text_rotation" style="width: 60px;" value='0' max='359' min='-359'>
+                                        </div>
+                                    </div>
                                 </div>
 
                                 <span class="settings-header">Selections</span>
--- a/anvio/data/interactive/index.html
+++ b/anvio/data/interactive/index.html
@@ -292,7 +292,7 @@
                                             <input type="number" id="support_font_size" style="width: 60px;" value='15' max='100' min='1'>
                                         </div>
                                     </div>
-                                    <div class="form-group form-margin" id='support_text_rotation'>
+                                    <div class="form-group form-margin" id='show_text_rotation'>
                                         <label class="col-md-4 settings-label" for="support_text_rotation" data-help="support-text-rotation">Text Rotation:</label>
                                         <div class="col-md-8">
                                             <input type="number" id="support_text_rotation" style="width: 60px;" value='0' max='359' min='-359'>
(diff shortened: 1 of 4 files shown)


meren:
Just git pulled, even though now symbols appear at the right place, I don't see any text :/

Display value range shows 0 to 100.

image

But when I change it to 80 to 100, all symbols disappear. They come back when I change the values to 0.8 and 1.0, but when I read the mouse panel, they are indeed 80, 90, 100, etc.

Perhaps "Floating Precision" should be under Text?

These are my settings and I'm visualizing `example-newick-tree.txt` under tests/sandbox:

image

When I click invert, this is what I see:

image

In either of the displays symbols do not seem to be properly scaled to max symbol size. I think we need a min symbol size parameter, too, to make this right. And when we have it, if the user says visualize branch support values between 0.8 and 1.0, 0.8 should be min symbol size, 1.0 should be max symbol size, and the symbol sizes of anything in between is scaled. Inverting it should only change the way they are shown (0.8 max, 1.0 min, and anything between is scaled).

We also need an input box for "Rotate" under text, which takes a value between 0 and 360 (deg) and the default is 90 for phylogram view.


Thank you :)

matthewlawrenceklein:
@meren - I think I may be misunderstanding bootstrap values - currently on render we're iterating each node `p` and displaying its associated `p.branch_support` . This is a number (in string form) between 0-100, which is why the initial `Display Value Range` was set to 0-100. I can see in the newick file itself that support values have 10 decimal places - are these values getting converted/rounded/truncated somewhere before getting passed to the frontend?

meren:
> I think I may be misunderstanding bootstrap values - currently on render we're iterating each node p and displaying its associated p.branch_support

This could be the case for all of us :) somehow the interface must make sure it uses the right ranges / values for each tree individually. Whether those values are 0.0 to 1.0 or 1 to 100.

>  I can see in the newick file itself that support values have 10 decimal places - are these values getting converted/rounded/truncated somewhere before getting passed to the frontend?

It is likely the doing of the library we use to read newick files (ETE). Another library may cut them at the second digit, or leave them as is. Being flexible _and_ wise (by anticipating future alternatives) is the best option, I think.

I am not sure if this helps, but I hope it does :)

matthewlawrenceklein:
@meren I think we're good to go here :) in summary - 

- by default, support values are displayed as whole numbers between the range of 0-100
- support symbol relative sizing is based on percentile within any given selected display range, broken into thirds (low, med, high)
    - `invert symbol` flips the above logic
- by default, support value text for phylograms is rotated -90 for legibility 
- A new parameter, text rotation, takes an input in range from -359 to 359, and will render text rotated accordingly
    - this parameter has been added to state 


## anyone can test! 
from the root anvio directory, run 
`anvi-interactive -t anvio/tests/sandbox/example-newick-tree.txt --manual`

meren:
Many things improved, but there is a selection issue due to symbols. Probably they need to be behind the tree?

selection

This is a minor thing, but it prevents from being able to read the support values of branches behind symbols.

But when you make a selection this is how it looks:

image

The selection pane should be behind the tree and symbols. I guess the best order from top to bottom is this: tree, symbols, selection pane.

PR #1782, opened 2021-08-12
Metabolism docs update :)

ivagljiva:
This PR clarifies and adds information to the help pages for the metabolism-related programs and artifacts. That is all :)

@meren, I would really appreciate it if you take a quick read through these docs and let me know if anything else should be added or changed. Thank you very much!


meren:
wow. this is incredible amount of useful information, @ivagljiva! On behalf of all the future users of this framework, thank you very much :)

I am merging this, and will update the web accordingly.

PR #1783, opened 2021-08-13
A few updates to anvi-estimate-metabolism

ivagljiva:
This PR originally started as an effort to fix some of the bottlenecks (*cough* memory *cough*) in the metabolism code, but evolved into fixing different things :) 

First, I re-vamped the gene-call loading strategy in `init_hits_and_splits()` (see [this commit](https://github.com/merenlab/anvio/commit/bb7c4fab71f981ce5312dcf0a13876f791f2a135)) so that it loads only the gene calls from the splits we need, rather than loading everything and then parsing out the ones we don't want later.

I tested an additional idea - which was to load genes for each set of splits (ie the splits for each bin, or each metagenomic contig), but this strategy turned out to be less fast than loading everything from the database at the beginning of the run, so I reverted those changes. You can see the outcome of my testing [here](https://github.com/merenlab/lab-fiesta/issues/1221#issuecomment-876779417).

Next, I updated the code to have much fewer database lookups. We now use the function `init_data_from_modules_db()` to load the `kegg_modules` table into memory at the start of the program, and rely on that for most of the information we need throughout the run (rather than repeatedly querying the `MODULES.db`). 

Finally, I fixed a couple of silly bugs in the program output. 

1) It used to be that the KO list under the `kofam_hits_in_module` column in modules mode output contained only unique KO numbers, even if some of the KOs were present in multiple copies. In these cases, the `gene_caller_ids_in_module` list was longer than the KO list because contained every single gene call for the KOs in the module, including for the KO duplicates. Now this is fixed - there is a 1:1 match between the KO list and the gene caller ID list, so people can easily figure out which gene caller ID belongs to which KO.

2) The `warnings` column in modules mode output used to include strings like `No KOfam profile for M00173` for modules that are defined by other modules, even though you cannot have a KOfam profile for a module. This doesn't happen anymore.

I have checked that these changes do not affect the metabolism results (except for the expected differences that come from the output-related bug fixes). 

☺️


meren:
I'm sorry to hear you couldn't put your finger on the memory leak, but it is great that it lead to finding other ways to improve stuff in general :) thank you very much!

PR #1794, opened 2021-09-10
Mode for adding invariant entries to variability tables (--kiefl-mode)

ekiefl:


ekiefl:
As of 0ec2897, invariant sites are now properly calculated but not yet added to self.data.

The implementation is quite fast. Initially I was worried, but the method runs in 4.75 seconds for 10 genes, whereas recovering base allele counts with `--quince-mode` takes 93 seconds for the same 10.

ekiefl:
## Tests

Ran variability on 10 genes:

```
anvi-gen-variability-profile -c HIMB083/CONTIGS.db -p HIMB083/PROFILE.db --genes-of-interest goi_test --kiefl-mode --engine CDN --samples-of-interest soi
```

Loaded up the data and ran the following sanity checks:

```python
df = pd.read_csv("variability.txt", sep='\t')
```

Gene lengths are properly added:

```python
In [5]: df.groupby('corresponding_gene_call').gene_length.unique()
Out[5]:
corresponding_gene_call
1248    [1191]
1249     [270]
1250     [531]
1251     [426]
1252     [690]
1253     [522]
1255    [4089]
1256    [4173]
1257     [372]
1258     [471]
Name: gene_length, dtype: object
```

Each `entry_id` is unique:

```python
In [6]: (df.entry_id.value_counts() == 1).all()
Out[6]: True
```

Every codon has 74 entries:

```python
In [8]: (df.groupby(['corresponding_gene_call', 'codon_order_in_gene']).unique_pos_identifier.value_counts() == 74).all()
Out[8]: True
```

Importantly, the number of `unique_pos_identifiers` exactly equals the number codons in the 10 genes:

```python
[18]: df['unique_pos_identifier'].nunique() == df['unique_pos_identifier'].nunique()
Out[18]: True
```

And finally, the reference codon for each entry is equal to the protein sequence's codon for each and every entry:

```python
In [10]: import anvio.dbops as dbops
          ...: import argparse
          ...: import anvio.utils as utils
          ...: db = dbops.ContigsSuperclass(argparse.Namespace(contigs_db='HIMB083/CONTIGS.db'))
          ...: db.init_contig_sequences()
          ...: codon_lookup = {}
          ...: for gene in df['corresponding_gene_call'].unique():
          ...:     codon_lookup[gene] = utils.get_list_of_codons_for_gene_call(db.genes_in_contigs_dict[gene], db.contig_sequences)
          ...: for i, row in df.iterrows():
          ...:     if (i % 10000) == 0:
          ...:         print(i)
          ...:     assert row['reference'] == codon_lookup[row['corresponding_gene_call']][row['codon_order_in_gene']], "Oh no!"
          ...:
Contigs DB ...................................: Initialized: HIMB083/CONTIGS.db (v. 20)
0
10000
20000
30000
40000
50000
60000
70000
80000
90000
100000
110000
120000
130000
140000
150000
160000
170000
180000
190000
200000
210000
220000
230000
240000
250000
260000
270000
280000
290000
300000
310000
```

ekiefl:
I will subject this to further testing by running it on a large dataset, but merging now because it doesn't interfere with any existing functionality. If I don't update this post, everything went well.

ekiefl:
A small fix was needed (9de47a4c342726ceb3e17eba4076e8387f9678f4). Unfortunately, we cannot add _every_ invariant site, because some sites have ambiguous nucleotides :\

meren:
Hey @ekiefl, this is great. Thank you very much.

While you're on this, the PR also needs some edits in this file (and other relevant files if there are any):

```
anvio/docs/programs/anvi-gen-variability-profile.md
``` 

We need to mention and describe all these changes in our docs to make sure they are accessible to others. And there is no one who can do it better than the person who realized the need and implemented it.

For instance, even your tests here could be a great addition to that document, so future generations could learn how to play with anvi'o databases from a python environment like a pro. For instance I added some things in that spirit here:

https://merenlab.org/software/anvio/help/main/artifacts/pan-db/

Not to create too much additional work, but since you already put the effort, I feel like it is sad when these details get buried in PRs.

ekiefl:
> While you're on this, the PR also needs some edits in this file (and other relevant files if there are any):

Good call. I added some info about it and linked the PR.

> For instance, even your tests here could be a great addition to that document, so future generations could learn how to play with anvi'o databases from a python environment like a pro.

I agree that showcasing how people could use pandas to explore the variability table effectively could be useful. I think a much more useful subject matter could be used to showcase this, however, so I opted to not include those tests directly in the document.

PR #1797, opened 2021-09-12
Aa support

ekiefl:
As @meren and I discussed, there is no good reason not to add `--engine AA` to the list of suitable engines for which `--kiefl-mode` may be applied. In fact, I soon realized that I need it for my own personal workflow. And thus it became.


meren:
🎊

PR #1807, opened 2021-09-22
A new program for fast summaries of read recruitment statistics

ivagljiva:
This PR contains a new program for rapid summaries of many single profile dbs, `anvi-quick-summary` (and its associated help pages). The base script was written by @meren (incredibly quickly, as I watched in amazement), but I have modified it slightly, cleaned it up and tested it, so here I am writing this PR. :)

You can check the doc pages for a description of how to run this program and the output it produces, but briefly, this program takes a list of single profile dbs, a contigs db, and a collection name. For each bin in the collection, it computes the weighted average of each statistic for the bin in every sample. The average is taken across every split in the bin, and the weights are the split lengths. The statistics to be summarized are by default detection and mean_coverage_Q2Q3, but generally can include any of the view tables in profiles (described in the constant `anvio.constants.essential_data_fields_for_anvio_profiles`). 

@meren, here are the things that have changed about this script since we last discussed it:
- there is no more `--matrix-format` option
- the output file is appended to as each profile is processed rather than all at once at the end, as you suggested for better memory usage/speed gains
- it now works with collections that do contain all splits in the database (stats are loaded only for the splits belonging to the collection, not for all splits)
- users can select which statistics they want to summarize with the `-S` flag
- more sanity checks :) 

You can test this program by running a command like the following:
```
anvi-quick-summary -c CONTIGS.db -C my_collection PROFILE_1.db PROFILE_2.db PROFILE_3.db [...]
```
on any set of single profiles (I have my own test set that I can share if you need it!). Note that only the first profile in the list needs to have the collection defined.

I would appreciate any input or advice you may have on improving it further :)


meren:
Hey @ivagljiva, thank you very much for creating this PR! You have done quite a bit of coding over my preliminary attempt 😂 

I was thinking, maybe we can rename the program `anvi-quick-summary` to `anvi-summary-blitz`? Since we now have a program called [anvi-profile-blitz](https://github.com/merenlab/anvio/pull/1804) :p

ivagljiva:
LOL. Absolutely, that name is so much more fun!

meren:
Thank you, @ivagljiva! :)

PR #1815, opened 2021-10-12
parseFloat instead of parseInt

matthewlawrenceklein:
support value symbol rendering now correctly handles values in the range of 0.000 - 1.000, and correctly scales symbol size based on percentile.


meren:
If you tested it, please feel free to merge it.

PR #1835, opened 2021-11-11
new lambda catches; close #1828

### LINKED ISSUE #1828: [BUG] Error with p-values in functional enrichment

I get errors about invalid p-values when running `anvi-compute-functional-enrichment-in-pan` with the `--include-gc-identity-as-function` option.

## anvi'o version

Anvi'o .......................................: hope (v7.1)

Profile database .............................: 38
Contigs database .............................: 20
Pan database .................................: 15
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 2
tRNA-seq database ............................: 2

## System info

OS is Red Hat Enterprise Linux release 8.3 (Ootpa).
Anvi'o is installed as a conda environment.

I ran the command `anvi-compute-functional-enrichment-in-pan -g ANME-3-EVO-GENOMES.db -p pangenomics/ANME3EVO-PAN.db --category-variable group --include-gc-identity-as-function -o function-enrichment/enrichment-IDENTITY.tsv -F function-enrichment/occurrence-IDENTITY.tsv --annotation-source IDENTITY` and successfully got an occurrence table file but got an error during the enrichment analysis. The log file for the error is copied below:

```
# DATE: 28 Oct 21 10:49:20
# CMD LINE: anvi-script-enrichment-stats --input /export/data1/tmp/tmpi6r4xx2z --output function-enrichment/enrichment-IDENTITY.tsv
Note: Using an external vector in selections is ambiguous.
ℹ Use `all_of(n_columns_before_data)` instead of `n_columns_before_data` to silence this message.
ℹ See <https://tidyselect.r-lib.org/reference/faq-external-vector.html>.
This message is displayed once per session.
Error: Problem with `mutate()` column `adjusted_q_value`.
ℹ `adjusted_q_value = ...$NULL`.
✖ ERROR: The estimated pi0 <= 0. Check that you have valid p-values or use a different range of lambda.
Backtrace:
     █
  1. ├─`%>%`(...)
  2. ├─dplyr::select(...)
  3. ├─dplyr::left_join(., df_in, by = "accession")
  4. ├─dplyr::mutate(...)
  5. ├─dplyr:::mutate.data.frame(...)
  6. │ └─dplyr:::mutate_cols(.data, ..., caller_env = caller_env())
  7. │   ├─base::withCallingHandlers(...)
  8. │   └─mask$eval_all_mutate(quo)
  9. ├─qvalue::qvalue(...)
 10. │ └─qvalue::pi0est(p, ...)
 11. │   └─base::stop("ERROR: The estimated pi0 <= 0. Check that you have valid p-values or use a different range of lambda.")
 12. └─base::.handleSimpleError(...)
 13.   └─dplyr:::h(simpleError(msg, call))
Execution halted
```

## Files to reproduce

I copied the temp file that gets passed to `anvi-script-enrichment-stats` and attached it below. I get the same message as above when I run the script directly and use this as input.
[IDENTITY-input.txt](https://github.com/merenlab/anvio/files/7436773/IDENTITY-input.txt)

adw96:
In #1828 , it was brought to our attention that running `anvi-compute-functional-enrichment-in-pan` could return the error "ERROR: The estimated pi0 <= 0. Check that you have valid p-values or use a different range of lambda." 

In this particular case, we had a lot of p-values between 0.45 and 0.55, and relatively few above 0.55. Determining q-values relies on an estimate of the proportion of features (here, functions) that are truly null, and most ways to do this assume a uniform distribution of p-values (at least above a threshold). With the particular distribution of p-values in this dataset, the smooth-and-extrapolate approach fails due to the small number of p-values above 0.55. 

There were a couple of ways to deal with this, but having read more about the smoothing approaches, I don't think they are necessary. I'm happy to declare that p-values above 0.2 are likely null, so in the language of Storey & Tibshirani (2003, PNAS), this PR sets set lambda=0.2 and uses this to estimate pi0 in the event of a failure of smooth-and-extrapolate. I think this approach induces minimal bias (for estimating pi0) but robust-ifys the program against situations where we have relatively few genomes in a particular group, and a discrete distribution for p-values with many near 0.5 and few near 1 (as happened in #1828 ). 

As an added bonus, I've significantly reduced the run time of the script by no longer duplicating calculation of the test statistics and p-values when the input data is identical. Therefore, if two different functions have the same proportions and numbers of observations in each group, then we only need to calculate the statistic once. For example, in the dataset that prompted the original issue, we ran ~24k tests, only 906 of which were unique. I think this was the largest bottleneck in the code, so we may speed up the script by 20x. Note that since the distribution of p-values matters for q-values (i.e., multiplicities are important), we do need to "unroll" the p-values to get the q-values -- no skipping that! 

I also used this issue as an excuse to replace the old school `gather` and `spread` with `pivot_wider` and `pivot_longer`, which are awesome. We may have some teething issues with tidyverse versions, but I think it's worth dealing with now as  it will be easier to maintain going forward. 

Happy to take questions/critical feedback! And huge thanks to @meren for helping with the git end of this.

### DIFF
--- a/sandbox/anvi-script-enrichment-stats
+++ b/sandbox/anvi-script-enrichment-stats
@@ -5,6 +5,7 @@
 
 ## Original Author: Amy Willis, March 2020
 ## Generalized by: Iva Veseli, Sept 2020
+## Last updated: Amy Willis, November 2021
 
 ## Usage:
 #         anvi-script-enrichment-stats --input=input.txt --output=output.txt
@@ -115,7 +116,7 @@ if (file.access(opts$input) == -1) {
 df_in <- readr::read_tsv(file = opts$input, col_types = cols())
 
 if (nrow(df_in) < 2) {
-    stop(sprintf("Your input file contains less than two lines :/ Not OK."))
+  stop(sprintf("Your input file contains less than two lines :/ Not OK."))
 }
 
 # Find number of groups
@@ -130,13 +131,13 @@ if (num_groups %% 1 != 0) {
 
 # make sure the input file contains expected columns
 for (column_name in c("accession", "associated_groups")) {
-    if (not (column_name %in% colnames(df_in)))
-        stop(sprintf("The input file '%s' does not contain an essential column ('%s').", opts$input, column_name))
+  if (not (column_name %in% colnames(df_in)))
+    stop(sprintf("The input file '%s' does not contain an essential column ('%s').", opts$input, column_name))
 }
 
 # make sure associated_groups column includes group names.
 if (nrow(drop_na(distinct(df_in, associated_groups))) == 0) {
-    stop(sprintf("You have no group associations in your data. Which means, nothing is differentially occurring across your groups :/"))
+  stop(sprintf("You have no group associations in your data. Which means, nothing is differentially occurring across your groups :/"))
 }
 
 # Set up functions to run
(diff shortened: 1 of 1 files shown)


meren:
I just tested it using the _Prochlorococcus_ pangenome. Most of the gene clusters had identical unadjusted p values and adjusted q values (1050 of 1450), and those that had different values (400) had extremely minimal differences for the vast majority of cases, such as this one here:

|COG14_FUNCTION|enrichment_score|unadjusted_p_value|adjusted_q_value|associated_groups|accession|gene_clusters_ids|p_LL|p_HL|N_LL|N_HL|
|:--|:--|:--|:--:|:--:|:--:|:--:|:--:|:--:|:--:|:--:|
|(2Fe-2S) ferredoxin|0.025619837268266114|0.87283225096348|1|LL|COG3411|GC_00001048|0.7273|0.7|11|20|
|(2Fe-2S) ferredoxin|0.0256716082324945|0.8727049241101241|1|LL|COG3411|GC_00001048|0.7273|0.7|11|20|

and never in cases where there was a significant difference between groups (I guess as expected if understand the problem correctly).

I believe this change will not change anyone's previous findings with the approach, but I confirm that the increase in speed is impressive.

Thank you very much, Amy!

Issue #1841, opened 2021-11-20
[BUG] Contig boundary errors in anvi-export-locus


philipwoods:
## Short description of the problem

`anvi-export-locus` sometimes fails to identify when a requested locus crosses a contig boundary.

## anvi'o version

```
Anvi'o .......................................: hope (v7.1)

Profile database .............................: 38
Contigs database .............................: 20
Pan database .................................: 15
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 2
tRNA-seq database ............................: 2
```

## System info

OS is Red Hat Enterprise Linux release 8.3 (Ootpa).
Anvi'o is installed as a conda environment.

## Detailed description of the issue

I ran `anvi-export-locus` on several genomes asking for 5 genes on each side of the target, and in several of them the target gene is near the end of a contig. In most of these, Anvio properly trimmed down the gene calls to just those on the correct contig, but in three of them it did not. This manifested in a few different ways down the line.

In two of the problem loci, a contig boundary was only within 5 genes on one side of the target. One of them had the target in the forward direction, and the other had the target in the reverse direction. In these, the eventual failure was creating a gene calls file with negative start and/or stop positions for the genes in the locus, which then caused an error.

In the third problem locus, the target was within 5 genes of a contig boundary on both sides. Anvio properly trimmed the gene calls on one side of the target, but not on the other side. This resulted in a gene calls file with genes from two contigs, and eventually there was an error in utils.get_most_likely_translation_frame because one of the sequences was too short (presumably one of the genes from the wrong contig, as the others all seem fine).

## Files to reproduce

Here is a Dropbox link with the databases, the console output I saw for each one, and the incorrect gene calls files they created. I also included instructions to reproduce.

https://www.dropbox.com/s/q0ht6r0pleo4yni/anvi-export-locus.zip?dl=0

mschecht:
@philipwoods thanks for documenting this and providing files to reproduce the error, this looks like a bug. I will look into it on Monday and get back to you.

mschecht:
Hi @philipwoods, apologies for the delay. I just proposed a solution in PR #1846 on the main branch and hopefully this will be solved soon. 

[...]

PR #1846, opened 2021-11-25
sort locus by gene start position

mschecht:
Hey @meren and @ekiefl, please check out this important change I am making to `anvi-export-locus` to address #1841.

Sorting by gene start position will allow for all gene-callers to be considered at once.

Shoutout to @ivagljiva for helping me come up with a solution!


meren:
Looks good :)

Issue #1848, opened 2021-11-30
[FEATURE] Anvi'o pangenomics self test is BORING!


matthewlawrenceklein:
## The need

The slumbering behemoth known as `Genome View` currently lies dormant, its genome clustering abilities waiting silently in the wings. Meanwhile, rumblings from the current Anvio user base speak towards a common need - our pangenomics test files are _boring_, and we need to spice them up!

## The solution

More exciting test files for Anvio's pangenomics self test

## Beneficiaries

self-testers at large, but especially users of the yet-to-be-finalized `genome-view` self test and tutorial, who will be able to utilize `Genome View`'s clustering features in a test environment :)

meren:
There is no call we do not answer. There is no faith that we betray.

If the pangenomics self test is boring, we make it unboring (it is now brewing in #1849).

Atreides. Atreides. Atreides. Atreides. Atreides. Atreides.

matthewlawrenceklein:
> There is no call we do not answer. There is no faith that we betray.
> 
> If the pangenomics self test is boring, we make it unboring (it is now brewing in #1849).
> 
> Atreides. Atreides. Atreides. Atreides. Atreides. Atreides.

The self tests must flow!

meren:
This is no longer the case.

Here's the screenshot of the pangenome new test yields:

<img width="1372" alt="image" src="https://user-images.githubusercontent.com/197307/162973150-ab12f2d0-3469-4b85-b95b-3c3d2b86d16e.png">

PR #1849, opened 2021-12-06
Update mock files for the component testing of pangenomics

meren:
We heard in #1848 that the current set of files for pangenomics tests were very boring (probably since they are three near-identical copies of the same DNA), so the purpose of this PR is to introduce a new set of files that are more biologically realistic.


meren:
@matthewlawrenceklein, you can switch to the `pan-test-files-update` branch, run

```
anvi-self-test --suite pangenomics -o PAN_TEST_FILES
```

and switch to the genome view branch to see if they look good (please send a screenshot for posterity).

Before we merge this branch there are a few more things to do even if everything is fine, including,

- [ ] Updating component tests for the anvi'o pangenomics workflow, which relies on the old files.
- [ ] Identifying and adjusting the online documentation to make sure they work with these new files.

I will take care of them once I have the OK from you that these are less boring compared to the previous files.

matthewlawrenceklein:
thanks for addressing this @meren ! When I switch over to the `pan-test-files-update` branch and run 

> `anvi-self-test --suite pangenomics -o PAN_TEST_FILES`

The following error kicks back 

```
Config Error: Something went wrong during the functional enrichment analysis :( We don't know
              what happened, but this log file could contain some clues: /tmp/tmp5h51_pg8




Config Error: According to the exit code ('255'), anvi'o suspects that something may have gone
              wrong while running your tests :/ We hope that the reason is clear to you from
              the lines above. But if you don't see anything obvious, and especially if the
              test ended up running until the end with reasonable looking final results, you
              shouldn't worry too much about this error. Life is short and we all can worry
              just a bit less.
```

reading from the log file, I see 

```
# DATE: 13 Dec 21 15:20:35
# CMD LINE: anvi-script-enrichment-stats --input /tmp/tmph6e3mkxw --output functions-enrichment.txt
Error in pivot_longer(., cols = -1, names_to = "type", values_to = "value") :
  could not find function "pivot_longer"
Calls: %>% ... withVisible -> eval -> eval -> _fseq -> freduce -> <Anonymous>
Execution halted
```

`pivot_longer` is called from the tidyverse (tidyr?) dependency, so I'm guessing I need to update/resolve something from within my conda environment, but I'm not sure where to start :)

meren:
Your system is broken :) The easiest solution for you is to remove the following lines from `anvio/tests/run_component_tests_for_pangenomics.sh` (and never commit it):

``` bash
INFO "Estimating enriched functions per pan group"
anvi-compute-functional-enrichment-in-pan -p TEST/TEST-PAN.db \
                                          -g TEST-GENOMES.db \
                                          --category group \
                                          --annotation-source COG_FUNCTION \
                                          -o functions-enrichment.txt \
                                          -F functional-occurence.txt
SHOW_FILE functions-enrichment.txt
SHOW_FILE functional-occurence.txt


```

PR #1852, opened 2021-12-08
Changing anvi-export-locus to use gene index rather than gene-caller-id

mschecht:
Hi Everyone,

I realized my original PR or #1841 was not a robust solution. However, I think I got it. 

This update of `anvi-export-locus` relies on ordering gene-calls of a contig by start position, then any locus exporting functionality is done with the index of the gene positions rather than the gene-caller-id which as we all know is not always in numerical order. 

Looking forward to your thoughts!

Cheers,
Matt


meren:
Every time I am surprised that the amount of detail one can pay attention to a problem has no limit for even most straightforward needs. Thanks for increasing the coverage of edge cases. I'm sure you tested thoroughly, as far as I'm concerned you should go ahead and merge it whenever you're ready.

Issue #1748, opened 2021-06-04
[BUG] anvi-run-hmms silently fails when multithreaded


dspeth:
## Short description of the problem

anvi-run-hmms silently fails when multithreaded, and run on a single (or low number of) contigs

## anvi'o version

```
Anvi'o .......................................: hope (v7)

Profile database .............................: 35
Contigs database .............................: 20
Pan database .................................: 14
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 2
tRNA-seq database ............................: 1
```

## System info

Linux version 4.9.0-13-amd64 (debian-kernel@lists.debian.org) (gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1) ) #1 SMP Debian 4.9.228-1 (2020-07-05)

## Detailed description of the issue

Hello anvio team, i hope this bug is something you have already encountered and fixed in the master branch, since it concerns HMMs, and I did notice you have spent quite a bit of energy on handling them recently. However, I could not find an issue specifically referencing this problem, so I'll open a new one.

I was building a phylogenomic tree using the steps described in the anvio phylogenomic workflow. My target was a recently published gammaproteobacterial endosymbiont ( https://www.ncbi.nlm.nih.gov/nuccore/LR794158), and so I was curious what the alignment for this specific organism looked like. However, a quick `grep` of the concatenated protein fastafile did not yield anything, which was somewhat surprising. Some other contigs databases, which were in multiple contigs, seemed to be fine. 

(Note: a potential side issue is that despite it's many warnings `anvi-get-sequences-for-hmm-hits` does not warn you when one of your contigs dbs does not have any hits to the HMM collection of choice, it juts omits them from the final alignment)

some sleuthing ensued, and I realized I had used `anvi-run-hmms` with the `-T 50` flag. (Did I really use 50 cores to run HMMs on a 300kb genome you ask? Yes I did, and so I maybe deserved the ensuing mess for being wasteful). Taking the contigs db and running `anvi-run-hmms` yielded a decreasing number of hits to the `Bacteria_71` HMM collection, with increasing number of cores, hitting 0 somewhere between 40-50, as detailed by the pasted output of the number of hits to the collection below. 
Since this seemed to behave fine for genomes in more pieces (no systematic testing done), I think it might have something to do with the partitioning of a single sequence during the multithreading. 

Hope this helps!
Daan

```
[...]

PR #1853, opened 2021-12-09
fix multithreading error. related to #1748

dspeth:
Pull request to fix the hmmer driver multithreading error described in issue #1748


meren:
Thank you very much for this CLEAN fix, @dspeth! :)

PR #1860, opened 2021-12-14
News modal

matthewlawrenceklein:
you __will__ be made aware of anvi'o news, and you'll __like__ it!

## the change
The news tab is gone! 'What news tab?', you say? Exactly! Anvi'o news now lives in a modal popup. An unavoidable, attention-grabbing modal. Don't worry, we still use a cookie to track whether you've already read the news. We seek not to annoy, simply to _inform_!

## how it works
Anvi'o stores a cookie in the browser that tracks the last-read news item. If there's a newer item, or if the cookie is trashed, the user will see all _new_ news items, as denoted by a lovely blue orb. The user can select 'mark all as read', and the cycle will begin anew. Should the user wish to review already-seen news, they can access _all_ news items from the interactive interface's hamburger dropdown menu, affectionately known as __Anvi'o UI Jail__ . 

## to test 
from this branch, start a new interactive session in the browser. Nothing pops up, right? Try deleting the cookie (I use the `EditThisCookie` chrome extension). Refresh your window, boom! You should see all the news. Mark all as read and then head into the codebase to `anvio/data/interactive/js/news.js` and uncomment lines `15 - 20`. Save and refresh your browser. This will register a new mock news item, which should appear all by itself. 'Mark all as read' again, and you're all set.  

## note
Once this is approved for merging, let me know and I'll remove the test code from the branch :)


meren:
This looks very good. I've made some minor changes, but I think it is good to go :)

Although I realized when I update the NEWS.md it doesn't update at the client side (the releases link was broken in the NEWS.md, I fixed it, it is still broken). Is there some sort of hashing of the file I wonder.

matthewlawrenceklein:
@meren thanks for taking a look! Do you recall which release link was broken? I've clicked through all of the new links and everything appears to be in order, at least from my end :)

As far as I'm aware, the only information the browser cookie stores is the title of the last seen news item (element `0` in the returned news items array), so it should be content agnostic. If the issue persists, I'll do some digging!

meren:
Great to hear. That must be me, then. As far as I'm concerned this is good to merge. Perhaps we can even make a little release (`v7.2`) to get this onto people's computers so we can SPAM them with all the exciting stuff :p

PR #1877, opened 2022-01-15
Update evan's info

ekiefl:
How does this look?


meren:
This looks great to me, Evan. I think the development of the INDELs framework deserves to be a separate entry. One may say it was a by product of a better designed anvi-profile in a way, but I think it still is a significant intellectual contribution that I hope the anvi'o community will make good use of in the future.

I would also suggest this for the bio part to maintain a common style with everyone else:

> Computational biologist and software developer with a research focus of developing analysis and visualization strategies for complex multi-'omics data, particularly at the intersection of structural biology and metagenomics

The language is dry because this way we don't have to pretend first-person speech when we are filling the information or updating for previous developers.

My 2 cents.

ekiefl:
Yeah good point about first person.

Ok I just split the contribution into 2. Merging now. Thanks for creating this.

PR #1890, opened 2022-02-17
Enzymes-txt input to anvi-estimate-metabolism

ivagljiva:
This PR adds a new input option to `anvi-estimate-metabolism`: the `enzymes-txt` artifact. This is a text file that contains a list of genes and their enzyme annotations (and optionally their coverage/detection data), which allows users to estimate metabolism on arbitrary sets of enzymes without requiring sequence data in a contigs database.

Big shout-out to @genomewalker, who implemented most of the basic code here in [this fork](https://github.com/genomewalker/anvio/tree/xfilter), which allows `anvi-estimate-metabolism` to work on output from his software [x-filter](https://github.com/genomewalker/x-filter)! All I did was generalize his work, make sure it works with user-defined modules, etc :)

Just a few more details (which can also be found in the updated documentation files, soon to be pushed to anvio.org)...

This is an example enzymes-txt file (only first 3 columns are required):

|**gene_id**|**enzyme_accession**|**source**|**coverage**|**detection**|
|:--|:--|:--|:--|:--|
|aad:TC41_3038|K02886|KOfam|4.44|0.7862318840579711|
|aca:ACP_1744|K02626|KOfam|4.522875816993464|0.7790055248618785|
|aco:Amico_1604|K00606|KOfam|2.63953488372093|0.8063380281690141|
|ade:Adeh_0623|K02669|KOfam|2.011764705882353|0.6639344262295082|

`anvi-estimate-metabolism` will take all entries from the `enzyme_accession` column and consider them to be coming from the same 'genome' for the purposes of estimating pathway completeness scores. The `gene_id` serves as a unique identifier for each enzyme instance, similar to the gene caller ID (and it can even be a gene caller ID, if the user chooses, but it doesn't have to be). 

The `source` column doesn't currently matter for the functioning of `anvi-estimate-metabolism`, so I struggled with the decision to leave it as a mandatory column. I realized that in my upcoming refactor of the metabolism code, this column may imminently become necessary, so I decided to keep it mandatory. If people really do not want to bother with finding the annotation sources of their enzymes, they can (for now) put an arbitrary string in this column and `anvi-estimate-metabolism` would not care, because it doesn't do anything with it.

Which brings me to the question: what happens if someone puts in a random enzyme from some random source that does not belong to any metabolic pathways (either KEGG Modules or user-defined)? The answer is, absolutely nothing. There will be no error, just a little warning about how the enzyme is not present in any metabolic pathways. I think that's fine, but feel free to express opinions to the contrary.

Anyway, here is how you run `anvi-estimate-metabolism` using this input file:
```
anvi-estimate-metabolism --enzymes-txt my_enzyme_set.txt
```

And if your input file contains both `coverage` and `detection`, you can elect to use these values (ie to compute pathway coverage/detection) with the `--add-coverage` flag. This only works if both columns are present. And if they are not both present, the program will not fail but will warn the user that this data was not available.

I think that is everything. Oh! Actually there are a couple more random fixies/behavior changes that are also included in this PR:

- `anvi-estimate-metabolism` used to fail when we couldn't find the function definition for a given enzyme. But it was looking for the function definition only in this silly file from KEGG called `ko_list.txt` - which worked fine for KEGG stuff but obviously not for user-defined modules. Now this is fixed:
  - first, the search for function definition is smarter, because if we can't find the enzyme in the `ko_list.txt` file, we look for it in the modules database (which contains this information based on the ORTHOLOGY lines in the module files)
  - second, if we can't find the function, we don't crash the whole program. We just list the function as "UNKNOWN" and move on with our lives.
- there was a bug in the function that extracts the modules database `modules` table into a dictionary, where it was not saving data_definition fields for data_names that occurred only once in a given module, even if the data_definition field contained a real piece of data (and not null). This is now fixed.

Okay. Thoughts, opinions, suggestions, etc - send them my way. :)


meren:
I was just able to go through this. Great work, Iva. As far as I can see it is good to go 🎉

PR #1895, opened 2022-03-04
Clusterize metagenomic workflow

mschecht:
This PR addresses EcoPhylo logging issues during the clusterizing of the metagenomic workflow .


meren:
I thought the directory name `00_LOGS` was not hard coded but parameterized?

mschecht:
I think it's hardcoded here but I could be mistaken. When I test it, the EcoPhylo `LOGS_DIR` will be overwritten in the default config file: https://github.com/merenlab/anvio/blob/master/anvio/workflows/__init__.py#L118

None of the other workflow `__init__` have a `LOGS_DIR` so it makes sense to have this in the workflow superclass init but unfortunately, EcoPhylo has a different directory structure. I would like to avoid messing with the workflow superclass but it might be necessary, what do you think?

meren:
> I would like to avoid messing with the workflow superclass but it might be necessary, what do you think?

Never mind! We can get back to it later if we have to :)

meren:
Merge as you wish.

PR #1910, opened 2022-04-06
KEGG BRITE

semiller10:
This pull request adds KEGG BRITE hierarchy annotations to genes in the contigs database upon running `anvi-run-kegg-kofams`. Categorizations of all known KOfams in BRITE gene/protein hierarchies (currently 57 hierarchies) are added to a new table in the modules database.

The purpose of adding BRITE categories is their utility in more quickly and accurately interpreting gene functions. Most gene orthologs, such as ribosomal proteins and RNAs, are not part of KEGG modules, which for the most part are biochemical pathways with defined reactants and products. Almost all orthologs, on the other hand, are classified in BRITE. The hierarchical nature of BRITE categories allows the user to group functions at finer or coarser resolution. The presence of orthologs in multiple categories can also reveal the range of different processes in which the gene participates.

The contigs database is updated to `v21`, adding a small table of information on the maximum depths of BRITE hierarchies that will be useful for further functionality in adjusting the specificity of categorizations. The modules database is updated to `v4`, adding the table of all KOfam BRITE categorizations and some additional metadata.

At the moment, BRITE annotations can only be used when downloading the most recent KEGG release to set up the modules database: `anvi-setup-kegg-kofams -D`. It is not advisable to add BRITE data to the archival KEGG data corresponding to anvi'o releases. @iva plans to allow users to supplement these archives with new downloaded data (i.e., the BRITE json files), allowing the user to remake the modules database with the old KOfam/module data but with BRITE as well.

Regarding the API, there are two significant methods added to the `ModulesDatabase` class in `kegg.py`: `get_brite_table_as_ortholog_dict` and `get_brite_table_as_hierarchy_dict`. The former returns a dictionary of ortholog categorizations, and the latter returns a dictionary of hierarchy structures. Both methods have a number of important parameters and provide a framework for dealing with this hierarchical data in other contexts.


semiller10 (commented):
Double checked changes

anvio/kegg.py
```diff
@@ -5232,20 +5527,106 @@ def create(self):
                               "'ANNOTATION_SOURCE' fields to those module files, and then re-do this setup.")
         annotation_source_list = ",".join(list(self.annotation_sources))
 
+        if self.brite_dict:
```
meren, line 5530: instead of such a long nested if, this part could go into its own function, e.g., `populate_brite_table`, and it would return with a warning early on `if not self.brite_dict`. It would also enforce a healthy level of modularity within the class.

meren:
This all looks very good to me, @semiller10. Beautiful work. Thank you for the migration scripts, as well.

I had a single suggestion on an `if` block in the code, and here are a few minor things that could improve everything a bit more IMO:

* `anvi-setup-kegg-kofams` can benefit from a `--skip-brite-hierarchies`. The self table of modules-db should know if it has BRITE hierarcies setup.
* `anvi-run-kegg-kofams` can benefit from a `--skip-brite-hierarchies`.
* A warning in `anvi-run-kegg-kofams` to let them know that there will be none of BRITE hierarchies if the user is using a modules-db without BRITE hierarchies.
* Finally, an update to the docs at `anvio/docs/programs/anvi-setup-kegg-kofams.md` and `anvio/docs/programs/anvi-run-kegg-kofams.md` -- since there is no mentioning of the BRITE in either of them at the moment.

My 2 cents.

anvio/kegg.py
```diff
@@ -1494,6 +1718,40 @@ def update_dict_for_genes_with_missing_annotations(self, gcids_list, hits_dict,
                                 'e_value': None,
                             }
 
+                            # add associated KEGG BRITE hierarchy information to database
+                            ortholog_brite_dict = self.kegg_modules_db.get_ortholog_brite_categorizations(knum)
```
ivagljiva, line 1722: This is definitely my bad, because I set up the code this way - the `update_dict_for_genes_with_missing_annotations` function has a lot of the same code as `parse_kofam_hits`, and I see that it now forced you to put the same code in both places as well. Sorry :(  

I think we should make this part of the code (for adding the annotations to the dict) its own function, if possible, and simply call that function in both places. It may not be straightforward, since I'm pretty sure I already thought about doing this (and for some reason, I didn't). But I'm happy to take another look and try to fix it (unless you've already considered it and know why it is this way)

anvio/migrations/modules/v3_to_v4.py
```diff
@@ -0,0 +1,81 @@
...
+    modules_db.remove_meta_key_value_pair('version')
+    modules_db.set_version(next_version)
+
+    modules_db.disconnect()
+    progress.end()
+
+    run.info_single(f"The modules database is now {next_version}. An empty table of KEGG BRITE hierarchy "
+                    "categorizations of all orthologs was created, and related self table attributes were added. "
+                    "We suggest generating a new modules database from `anvi-setup-kegg-kofams` (using the `-D` flag "
+                    "in anvi'o v7 or earlier) and re-running `anvi-run-kegg-kofams` on your contigs databases "
```
ivagljiva, line 67: Shouldn't "anvi'o v7 or earlier" actually be "anvi'o v7 or later"? I don't think the -D flag exists prior to v7 :)

anvio/kegg.py
```diff
@@ -1518,12 +1776,32 @@ def store_annotations_in_db(self):
...
             self.run.warning("There are no KOfam hits to add to the database. Returning empty handed, "
                              "but still adding KOfam as a functional source.")
             gene_function_calls_table.add_empty_sources_to_functional_sources({'KOfam'})
 
 
+    def store_brite_info_in_db(self):
+        """Fills in the table of BRITE hierarchy information in the contigs database.
+
+        The only data stored in this table regards hierarchy depths, gleaned from the full set of
+        information in the modules database. Depth data is needed for pruning categorizations to the
```
ivagljiva, line 1791: If every contigs database would store the same hierarchy depth information, wouldn't it be better to just make this another table in the modules database instead? That way we would only increase one database version and different contigs DBs wouldn't be storing redundant information (and would be smaller). 

But I might be misunderstanding something here, so please tell me if there is a reason to have this info in the contigs db :)

ivagljiva (commented):
This looks great @semiller10 :) I left just a few comments and suggestions for you to consider.

I like that you added the accession decorator in the constants file; I should do that for the KEGG MODULE and KOfam annotations as well!

anvio/kegg.py
meren: Very good point :) I think this is worth thinking about!

anvio/kegg.py
semiller10: Don't worry about it, but thanks for reminding me of this, since it was straightforward to make a new method to remove the duplication on my part.

anvio/kegg.py
semiller10: It's funny when @ivagljiva reads my mind and grasps nuances of implementation that I mulled over myself. The reason I didn't store hierarchy depth information in the modules database is that operations on the contigs database shouldn't depend on the existence of the modules database. I envision the depth data coming into use when displaying categories in `anvi-display-functions`. If the user wants to display more general categories, lumping together the N categories at the bottom of the hierarchies, then depths need to be stored somewhere, and ideally this should not require a call to the modules database that originally categorized functions in the contigs database. The thought of storing depth data as an anvi'o constant also flitted through my mind, but hierarchy depths can surely change in the future.

I now see this might have implications for module completeness. Should module completeness calculations require referring from the contigs to the modules database? Or should module completeness be stored directly in the contigs database at the time of module annotation.

meren: > Should module completeness calculations require referring from the contigs to the modules database?

AFAIK, it already does. You need to have the modules-db properly setup on your computer to run completion stats based on KOfam hits in a given contigs-db. But not having it does not prevent us from distributing a contigs-db alone.

It would make a lot of sense to keep BRITE categories or hierarchy depth information only in modules-db, rather than the contigs-db. But this is going to add a lot of technical headaches (but will be much more elegant).

First of all, every time we run "get gene functions", we will have to add BRITE data on the fly to the resulting functions dict based on KOfam hits we find (i.e., get all functions will return more sources than what is literally in the gene functions table). I am not sure if this is a serious computational bottleneck, but I wouldn't think so.

Then there is the issue of reproducibility / interoperability. For instance, imagine this scenario:

A user downloads a contigs-db from a paper. It has KOfam annotations, but they don't have a modules-db setup on their computer. Well, in that. case they will not be able to get the BRITE hierarchy information for the genes in that contigs-db. There will be a warning or an error saying that they should run `anvi-setup-kegg-kofams` first. I'm not sure if this is a problem either because the same user will not be able to generate metabolism completion estimates without a modules-db on their system, either.

An alternative is to have this 'on-the-fly' / modules-db requiring behavior the default, and also have a flag for the program `anvi-setup-kegg-kofams`, `--store-brite-hierarchy-data-in-contigs-db` where these hierarchies would be literally stored in contigs-db. In this case the same user above would be able to work with BRITE hierarchies without having to setup a modules-db on their computer if the contigs-db was annotated with that flag. But I'm not sure if this level of complexity is justified since those who will use BRITE hierarchies will likely also have modules-db setup on their computers. It is just an issue of helpful warning / error messages so users can find their ways by themselves.

What do you guys think of all this?

semiller10: I'm in favor of keeping BRITE categories in the gene functions table of the contigs db since KOfam module assignments are stored there. To guarantee that the BRITE and module categorizations are from the same version of KEGG, they should either both be in the contigs db or both be in the modules db, since the modules db can be changed.

@meren's proposed flag to store BRITE categorizations in the contigs db provides a solution to this problem by "locking" the version in place. Anvi'o could also emit a warning if the modules db hash is different than the hash recorded in the contigs db.

On a related note, it is probably worth recording a timestamp for KEGG in the modules db when the db is set up with a downloaded or archived version of KEGG so that the user can reference the version used in the analysis.

The issue of database dependencies could be addressed by a tabular log file listing the paths and hashes of dbs associated with the contigs db that anvi'o generates and modifies as different commands are run. The file would appear in the same directory as the contigs db upon creation. If `anvi-run-kegg-kofams` is then run, a line would be added showing that the contigs db is associated with the modules db. If `anvi-profile` is then run, another line would be added with this historical record.

anvio/migrations/modules/v3_to_v4.py
semiller10: I think I have clarified what I meant.

anvio/kegg.py
semiller10: After talking to @meren, I better understand the rationale for confining BRITE hierarchies to the modules db and categorizing KOfams from the contigs db on the fly. We warn the user to keep KEGG annotations in the contigs db up-to-date with the modules db. The user shouldn't have a modules db from a version of KEGG ahead of the contigs db annotations.

So how should BRITE categorizations be found on the fly? `anvi-interactive` pulls functions from the contigs db using `dbops.ContigsSuperclass.init_functions`, populating a dictionary of function calls from different sources. If KOfams are present, the modules db can be loaded and BRITE categorizations added to this dictionary. There should be a method of `kegg.ModulesDatabase` that takes a list of KOfams and returns BRITE categorizations in a dictionary compatible with the gene functions format.

semiller10:
> here are a few minor things that could improve everything a bit more IMO:
> 
> * `anvi-setup-kegg-kofams` can benefit from a `--skip-brite-hierarchies`. The self table of modules-db should know if it has BRITE hierarcies setup.
> * `anvi-run-kegg-kofams` can benefit from a `--skip-brite-hierarchies`.
> * A warning in `anvi-run-kegg-kofams` to let them know that there will be none of BRITE hierarchies if the user is using a modules-db without BRITE hierarchies.
> * Finally, an update to the docs at `anvio/docs/programs/anvi-setup-kegg-kofams.md` and `anvio/docs/programs/anvi-run-kegg-kofams.md` -- since there is no mentioning of the BRITE in either of them at the moment.

Thank you, the code has been changed accordingly.

meren:
> Thank you, the code has been changed accordingly.

You're the best, Sam. Thank you.

anvio/kegg.py
semiller10: I decided to completely get rid of the separate table for BRITE hierarchy depths, since we decided against including this as "static" information in the contigs db, and depths can be computed on the fly from the modules db for `anvi-display-functions` or another application.

semiller10:
Everything passed final tests with the commands:
```
anvi-setup-kegg-kofams --reset
anvi-setup-kegg-kofams --reset -D
anvi-run-kegg-kofams -c CONTIGS.db --just-do-it
anvi-run-kegg-kofams -c CONTIGS.db --skip-brite-hierarchies
```

PR #1913, opened 2022-04-08
Added CI github workflow 

nick-youngblut:
Added a simple github workflow for continuous integration. More tests can easily be added besides `anvi-self-test --suite mini --no-interactive`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,12 +28,12 @@ jobs:
         conda install -y mamba
         mamba install -y python=${{ matrix.python-version }} "sqlite >=3.31.1" prodigal mcl muscle=3.8.1551 hmmer diamond blast megahit spades bowtie2 tbb=2019.8 bwa samtools=1.9 centrifuge trimal iqtree trnascan-se r-base r-stringi r-tidyverse r-magrittr r-optparse bioconductor-qvalue fasttree fastani
     - uses: actions/checkout@v2	
-    - name: package install
+    - name: pip install of anvio
       shell: bash -l {0}
       run: |
         curl -L https://github.com/merenlab/anvio/releases/download/v7.1/anvio-7.1.tar.gz --output anvio-7.1.tar.gz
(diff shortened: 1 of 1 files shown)


meren:
Hey @nick-youngblut, thank you very much for this! I have a quick question:

The following lines will always install the latest stable version to run things,

```
        curl -L https://github.com/merenlab/anvio/releases/download/v7.1/anvio-7.1.tar.gz --output anvio-7.1.tar.gz
        pip install anvio-7.1.tar.gz
```

But for instance when I merge #1914, even if I update the list of conda packages to match to the changes in the installation instructions (https://github.com/merenlab/anvio.org/pull/11), the `requirements.txt` in `anvio-7.1.tar.gz` will be out-dated, and thus the test will fail until there is a new stable version, i.e., `anvio-7.2.tar.gz` or `anvio-8.tar.gz`. Do you think it would be better to pip install from the GitHub repository directly?

nick-youngblut:
>Hey @nick-youngblut, thank you very much for this!

No problem! It really only took a few minutes, given the simplicity of Github workflows. 

>Do you think it would be better to pip install from the GitHub repository directly?

Yeah, that would be especially helpful for the CI, so you can quickly see if your last commit broke the install or tests. I just followed your install instructions, but a `pip install .` would be better.

meren:
> No problem! It really only took a few minutes, given the simplicity of Github workflows.

It is still infinite times more minutes than the amount of time almost everyone else in the world had or anvi'o :) So thank you.

> 
Hey @nick-youngblut, thank you very much for this!

No problem! It really only took a few minutes, given the simplicity of Github workflows.

Do you think it would be better to pip install from the GitHub repository directly?

Yeah, that would be especially helpful for the CI, so you can quickly see if your last commit broke the install or tests. I just followed your install instructions, but a pip install . would be better.

Saw the update. I'm merging it so we can see how things go.

While you're on this, would you mind sending another PR to add yourself to our `DEVELOPERS.yaml` file? Some info and instructions are here:

Best wishes,

PR #1916, opened 2022-04-10
Added exclude threshold for when using enzymes_txt

genomewalker:
Hey @ivagljiva and @meren 
I added a new exclude_threshold so when generating the dict it includes all the ones without thresholds. I set [here](https://github.com/genomewalker/anvio/blob/exclude-threshold/anvio/kegg.py#L2095) the annotation_source_set to `KOfam`, I don't know if in the future you want to use the source coming from the TSV file. At the moment I just followed a KISS approach. Also fixed a small keyerror that can happen as a result that we are hitchhiking the KOfam system you set up initially.

I attach [here](https://github.com/merenlab/anvio/files/8459412/302cec25c7.profile.kegg_group-abundances-anvio.tsv.gz) one of the offending files


ivagljiva:
Hey Antonio, sorry for my delay in reviewing this. 

I like the addition of the `exclude_threshold` parameter, but I don't think we should always enforce it for enzymes-txt input (or JSON input). It basically becomes an invisible parameter that users will have no control over or understanding of. The behavior of this program on the --enzymes-txt input should be exactly the same as its behavior on other input types, and this exclude_threshold flag should be tunable for all input types.

Therefore, we need a flag, which I am calling `--include-kos-not-from-current-kegg` for now (if you can think of a more concise but still clear flag name, please let me know and we can change it). Anyone who has obtained KOfam annotations from outside anvi'o (which may include those profiles that do not have a bitscore threshold or profiles from a newer version of KEGG) can use this flag, regardless of whether they have a contigs database or an enzymes-txt. And by explicitly setting this flag, they will know and understand that they are including these annotations into the metabolism estimates. 

I implemented the changes necessary to turn this into a flag, and I also created a nicer warning for the KeyError situation which tells users that they can use this flag. I used the file you shared to test this.

The following 
```
anvi-estimate-metabolism --enzymes-txt 302cec25c7.profile.kegg_group-abundances-anvio.tsv
```
results in this error:
```
Config Error: We cannot find the KEGG enzyme K25518 in the dictionary of enzyme hits, even
              though this enzyme is annotated in your data. This usually happens when you are
              using `KOfam` annotations that are different from the set of profiles we use to
              annotate in `anvi-run-kegg-kofams` (most typically, this will happen with
              --enzymes-txt input, but it can also happen if you imported external KOfam
              annotations with the source name `KOfam`). If you want to include these enzymes
              in this analysis, you will have to re-run this program with the flag --include-
              kos-not-from-current-kegg. If you just now realized that it is a bad idea to
              include these enzymes, then you'll have to re-do your annotations or remove them
              from your input --enzymes-txt file.
```

And the following code just works.
```
anvi-estimate-metabolism --enzymes-txt 302cec25c7.profile.kegg_group-abundances-anvio.tsv --include-kos-not-from-current-kegg
```

I don't have permission to write to this branch so if you want to review my code updates, I will make a PR to this branch, which you could then add to this PR. Otherwise, I can just try to merge it to the anvi'o dev branch - please let me know what you would prefer.

genomewalker:
Hi @ivagljiva 
thanks! The flag is a better solution. Maybe we can call it `--include-kos-not-in-kofam`. These KOs are in the KEGG release, but they don't have enough sequences to build an HMM and be included in KOfam.

I gave you access, but feel free to merge it into anvi'o dev branch.

Many thanks!

ivagljiva:
Thanks @genomewalker :) This is much easier.
I've added my changes here, and I updated the flag name to what you suggested, it's much better.

Feel free to look over the changes, and let me know if you think anything else needs changing. I'll add documentation to the flag and plan to merge to master in about an hour if I don't hear from you.

PR #1917, opened 2022-04-11
Python upgrade trnaseq patch

semiller10:
This PR patches bugs that arose in testing the tRNA-seq workflow with python `3.7`, upgraded from `3.6`.

Following anvi'o installation instructions, the default version of `pandas` installed without version specification was `0.25.1`. I changed `requirements.txt` to use `pandas>=1.2.1` in order to fix an error involving DataFrame creation. See:
https://stackoverflow.com/questions/65998646/pandas-error-for-creating-an-emptydataframe

The answer to the stack overflow question indicates that the error is due to incompatibility between versions of pandas and numpy. The default version of numpy being installed is `1.21.5`: it is possible the error can also be fixed by downgrading numpy to `1.19.5` if pandas cannot be upgraded for other reasons.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ anvio/data/misc/SCG_TAXONOMY/GTDB/SCG_SEARCH_DATABASES/*.dmnd
 anvio/tests/sandbox/update-contigs-and-bams-for-mini-test/output
 anvio/tests/sandbox/test_visualize_split_coverages/TEST_OUTDIR
 anvio/data/misc/KEGG/
+anvio/data/misc/TRNA_TAXONOMY/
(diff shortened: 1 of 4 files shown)


meren:
Interesting. When I follow anvi'o instructions on the web I get this:

```
$ python
Python 3.7.13 (default, Mar 28 2022, 07:24:34)
[Clang 12.0.0 ] :: Anaconda, Inc. on darwin
>>> import pandas
>>> pandas.__version__
'1.3.5'
>>>
```

meren:
But it can't hurt to be a bit more explicit in the `requirements.txt`, I guess.

meren:
Thanks for these changes, Sam! :)

Issue #1921, opened 2022-04-13
anvi-gen-contigs-database: force-overwrite


nick-youngblut:
## The need

If the user cannot force-overwrite output, then the user must delete any existing output, which requires more code and a potential to delete the wrong file(s), versus just using `--force-overwrite`. Specifically:

```
Config Error: Anvi'o will not overwrite an existing contigs database. Please choose a
              different name or remove the existing database ('/path/to/contigs.db') first.
```

## The solution

Include `--force-overwrite` for `anvi-gen-contigs-database` (and similar commands) 

## Beneficiaries

Everyone who doesn't want to write their code as:

```
rm -f $CONTIG_DB_FILE && anvi-gen-contigs-database -f $INPUT_FASTA -o $CONTIG_DB_FILE 
```

..but rather:

```
anvi-gen-contigs-database --force-overwrite -f $INPUT_FASTA -o $CONTIG_DB_FILE 
```

## Notes

I'm using anvio v7.1

meren:
Now in main.

PR #1923, opened 2022-04-13
A new community flag: --force-overwrite

meren:
This PR introduces a new community flag, `--force-overwrite`, which can be used by any anvi'o program, and would result in the removal of the output files and directories that are not needed.

This was requested by @nick-youngblut in #1921.


ivagljiva (approved):
Looks good to me @meren! :)

meren:
🎉

PR #1925, opened 2022-04-13
Legends Tab Performance and UI upgrades

matthewlawrenceklein:
Apologies in advance y'all, I'm gonna goof up some terminology below. Please feel free to correct :) 

## the problem
Historically, the interactive interface scans through each layer of a given dataset and generates a color dictionary value for each distinct categorical stat in that layer. These color values are then used to populate the legends tab of the settings pane, with lovely functionality to sort, count, and color your items there. Unfortunately, this process becomes a major bottleneck with datasets containing hundreds (or thousands!) of unique categories per layer. And even if it were not a performance bottleneck, a legend tab with 1000+ entries becomes pretty unruly to work with. 

### part one
first, we've established a hard cutoff for processing categorical values per layer when building our legends tab
- if the layer has < 20 distinct categorical stats, we proceed as normal
- if the layer has > 20 distinct categorical stats, we implement a new Query UI for that layer. The UI provides:
    - a text input box and colorpicker to query the layer against a specific category and update that color
    - a colorpicker and button to color all items in that layer the same color
    - a button to set new random color values for all categories in the layer

Users are greeted with a friendly and ephemeral success message upon the completion of any of the above actions

### part two
Because we know that copying and pasting categories from the mouse context menu or, even worse, writing the by hand, can be tedious and impractical, we've added some additional quality of life functionality
- right clicking on an item in your visualization now provides an 'open in legends tab' button. Clicking this will
    - toggle open the settings pane if not already open
    - toggle open the legends tab if not already open
    - toggle open the specific legend of the right-clicked item
    - pre-fill the text input box with the clicked item's category value
all that's left to do is select your action and go!

## the payoff
!!anecdotal data alert!!
Testing against a dataset with ~ 10 layers each containing ~ 10,000 categories
- with the existing legends tab functionality, the interactive interfaces time to first draw was around 80 minutes
- with the new legends tab implementation, the time to first draw dropped to about __10 seconds__

obviously your mileage may vary, but this seems like a win overall :)


meren:
Hi @matthewlawrenceklein, 

Here is something that that does not seem to be working (and reproducible with the infant gut tutorial data):

Untitled3

matthewlawrenceklein:
@meren 
latest commits resolve:
- 'open in legends tab' menu option is only available for layers with categorical data
- using the jquery accordion 'set active' hook instead of a simulated click, so individual legends should show/hide intelligently now

tested against vanilla IG dataset, IG dataset with 1 cat layer, IG dataset with many cat layers, and matt's giant dataset

thanks for bug hunting! :)

meren:
I tested and tweaked the cosmetics a bit with the last commit which turned this,

image

into this:

image

Looking more at the legends tab, I realized I don't see any reason why we couldn't make available the Batch coloring options we have for the layers with smaller number of items also in layers with many legend items:

image

But I have to move on to other things now. If you feel like it please take a look at it. Apart from that, I think this is ready to go into the main branch. Thank you for solving this bottleneck.

PR #1937, opened 2022-06-03
Fix metabat2 threads

dutter:
Ran into a case on our server where threads via -T wasn't being passed to metabat2 through anvi-cluster-contigs, so I manually added it :)


meren:
Hey @dutter,

Thank you very much for sending this PR!

When I look at the [changes](https://github.com/merenlab/anvio/pull/1937/files), I realize there are more changes than the `metabat2` related one. I am very happy that you decided to fix this for everyone and I don't want to put more work on your plate, but would you be so kind to send another PR only with the relevant change? :)

Thank you again,
Best wishes.

PR #1938, opened 2022-06-06
metabat2 threads v2

dutter:
Sorry about the messy PR before! This should be cleaner :)


meren:
Thank you very much, @dutter! :) It is now in the main repository.

Issue #1383, opened 2020-03-19
get-enriched-functions-per-pan-group, R issue?


TonyMane:
```
Anvi'o version ...............................: esther (v6.1-master)
Profile DB version ...........................: 31
Contigs DB version ...........................: 14
Pan DB version ...............................: 13
Genome data storage version ..................: 6
Auxiliary data storage version ...............: 2
Structure DB version .........................: 1
```

Having an issue with 'get-enriched-functions-per-pan-group'. I have ran this function successfully on the Prochlorococcus tutorial. But can't seem to get it work on data-set i have been working on. Below is the specific command I have run:

```
(anvio-master) (anvio-master) -bash-4.2$ anvi-get-enriched-functions-per-pan-group -p marine-PAN.db -g marine-GENOMES.db --category clade --annotation COG_FUNCTION -o OMZ.txt
Genomes storage .............................................: Initialized (storage hash: hash46f728cc)                                                                                                                                                       
Num genomes in storage ......................................: 20
Num genomes will be used ....................................: 20
Pan DB ......................................................: Initialized: marine-PAN.db (v. 13)
Gene cluster homogeneity estimates ..........................: Functional: [YES]; Geometric: [YES]; Combined: [YES]
                                                                                                                                                                                                                                                              
* Gene clusters are initialized for all 1733 gene clusters in the database.

Category ....................................................: clade                                                                                                                                                                                          
Functional annotation source ................................: COG_FUNCTION
Exclude ungrouped ...........................................: False
Functional occurrence summary ...............................: /tmp/tmpkphxhqfo                                                                                                                                                                               
                                                                                                                                                                                                                                                              

Config Error: It looks like something went wrong during the functional enrichment analysis. We
              don't know what happened, but this log file could contain some clues:           
              /tmp/tmpi9qoa3k5
```

[github_0319202.tar.gz](https://github.com/merenlab/anvio/files/4357205/github_0319202.tar.gz)

I have attached the marine-PAN.db , marine-GENOMES.db, the text file i used to populate the layers (marine_layer.txt, with anvi-import-misc-data) and tmp/tmpi9q0a3k5 (all in github_0319202.tar.gz).

Any commments would be useful!!!

meren:
I think this is an @adw96 or @mooreryan question (whoever sees / or has time for this first, as I am not qualified to debug in R).

But to make testing easier, I created the following two files:

[...]

PR #1953, opened 2022-07-06
Add lambda parameter for q-value estimation in enrichment script

ivagljiva:
This PR addresses an issue discovered [here](https://github.com/merenlab/anvio/issues/1383#issuecomment-1169142656), whereby very small p-values would lead to an ugly `seq` bug like this:
```
Error in seq.default(0.05, max_lambda, 0.05) :
  wrong sign in 'by' argument
```
(which can happen because the input file is too short, due to the fact that we remove all core functions/annotations when we generate the input file).

There are two new aspects of the code as of these changes:
1. When the `max_lambda` is too small, we stop the script and print a nicer error for the user:
```
Error: Unfortunately, the maximum lambda for q-value estimation is < 0.05. This
         value is coming from the p-values in the enrichment test, so something in
         your data is making those p-values low. This sometimes happens when you
         have a very short input file (your input file has 4 rows). Regardless,
         a possible solution is for you to pick your own max lambda that is >= 0.05,
         and provide it to the program calling this script by using the
         --qlambda parameter.
```
2. The enrichment script now accepts a `--qlambda` parameter allowing the user to set `max_lambda` themselves. This optional parameter has been added to the program `anvi-compute-functional-enrichment-across-genomes`.

On my test data (Enterococcus genomes from the Infant Gut Tutorial, using the annotation source `COG20_PATHWAY` which only has 3 annotations differentially present across the genomes), using the `--qlambda` parameter allows the script to proceed through the q-value estimation, but it still ends in this later error:
```
Error: Doh! We still can't estimate the proportion of features that are truly null. It's Amy's fault, so please let her know ASAP so she can fix it!
```
Which I suspect is happening because the input file is just too darn small. So even the addition of this parameter doesn't really fix this issue. But at least now users have a bit more control.

@adw96, could you please take a look at these changes and let me know what you think? If you like them, I'll update the documentation and merge the code :)

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -626,6 +626,14 @@ D = {
                      "A file name must be provided. To learn more about how the functional occurrence is computed, please "
                      "refer to the tutorial."}
                 ),
+    'qlambda': (
+            ['--qlambda'],
+            {'default': None,
+             'metavar': 'NUM',
+             'type': float,
+             'help': "If you want to set the maximum lambda value for q-value estimation, provide a fraction between "
+                     "0.05 and 1."}
+                ),
     'table': (
             ['--table'],
             {'metavar': 'TABLE_NAME',
--- a/anvio/genomedescriptions.py
+++ b/anvio/genomedescriptions.py
@@ -922,6 +922,7 @@ class AggregateFunctions:
         self.print_genome_names_and_quit = A('print_genome_names_and_quit') or False
         self.functional_occurrence_table_output_path = A('functional_occurrence_table_output')
         self.functional_enrichment_output_path = A('output_file')
+        self.qlambda = A('qlambda')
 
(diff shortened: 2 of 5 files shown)


adw96:
@ivagljiva Thanks for your work on this!

As an alternative to erroring when we can't estimate lambda is to return the rest of the results (with p-values) but not the q-values. I think a better approach is to provide the users with a **warning** that we couldn't do FDR correction (maybe with q-value = NA for all entries so they don't miss the warning) but not an error. 

Essentially I think we skip over 

```
qvalues_df <- pvalues_df %>%
  mutate("adjusted_q_value" = qvalue::qvalue(p=unadjusted_p_value, lambda=lambdas, pi0.method="smoother")$qvalues)
```

(or replace it with `qvalues_df <- pvalues_df %>% mutate("adjusted_q_value" = NA)` so we have the same number of columns always)

if the `"try-error" %in% class(pi0_est)` is `TRUE`. 

What do you think?

PR #1959, opened 2022-07-12
Hmm hit combine cov

mschecht:
This PR addresses merging profile HMM alignments that hit a target sequence in multiple parts of the protein in `anvi-script-filter-hmm-hits-table` and helping `anvi-report-inversions` report inversions more comprehensively.


mschecht:
> Instead of --target-coverage and --query-coverage, we should use --gene-coverage (how much of the gene is covered by the model) and --model-coverage (how much of the model is covered by the gene) so it is more intuitive.

You got it!

PR #1968, opened 2022-08-12
anvi-script-process-genbank += --include-locus-tags-as-functions

meren:
During the conversion of GenBank entries, anvi'o assigns a new gene call id to each entry, breaking the link between locus tags defined in the GenBank file and the gene entries that will later appear in the anvi'o generated from a GenBank file.

This PR adds a new flag to the program, `--include-locus-tags-as-functions`, which instructs anvi'o to add a new 'function' source for each gene in the output file for functional annotations so that the user can trace back a given gene call to the original locus tag.

We thank Luke McKay for [bringing this up](https://anvio.slack.com/archives/C8SFMGYF3/p1660244245176539) on anvi'o Slack :)

### DIFF
--- a/anvio/contigops.py
+++ b/anvio/contigops.py
@@ -728,7 +728,7 @@ class GenbankToAnvio:
 
         output_fasta = {}
         output_gene_calls = {}
-        output_functions = {}
+        output_functions = []
         num_genbank_records_processed = 0
         num_genes_found = 0
         num_genes_reported = 0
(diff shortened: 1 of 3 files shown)


luuuuuuuke:
This is awesome! But it looks like the external-functions files that were exported during anvi-script-process-genbank lost their headers. When I try to import them with anvi-import-functions, it gets mad at me because the headers aren't there. Eg:

Config Error: The file 'Genome1-external-functions.txt' does not contain the
              right type of header. It was expected to have these: 'gene_callers_id, source,
              accession, function, e_value', however it had these: 'NCBI_PGAP,
              extdb:pgaptmp_000001, TOBE domain-containing protein, 0'

meren:
Ouch. Give me 2 mins and I'll fix it. Sorry!

meren:
OK. https://github.com/merenlab/anvio/commit/b3506ad1c084260a315c971f9b1b018293da8e0c should fix it. Apologies for this.

luuuuuuuke:
no worries and thank you 1 billion times over for the fix!

Issue #1970, opened 2022-08-21
[BUG] anvi-report-inversions error when contig has coverage of 0 in sample


watsonar:
## Short description of the problem

Hello! I am running anvi-report-inversions using a contigs db consisting of multiple genomes. I am getting a type error (TypeError: 'NoneType' object is not iterable) that I think is happening because one of my contigs has zero coverage in one of my samples.

## anvi'o version

```
Anvi'o .......................................: hope (v7.1-dev)

Profile database .............................: 38
Contigs database .............................: 20
Pan database .................................: 16
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 4
tRNA-seq database ............................: 2
```

## System info

I using an AWS EC2 instance running Ubuntu 22.04 LTS, with 32 CPUS and 128 GB of memory. I installed anvio-dev following these instructions: https://anvio.org/install/#5-follow-the-active-development-youre-a-wizard-arry

## Detailed description of the issue

I am running the following command:
```
anvi-report-inversions -P bams_and_profiles.txt.short \
                       -T 32 \
                       --force-overwrite \
                       --debug
```
And receiving this output:
```
[General] Input BAMs and profiles file ......................: bams_and_profiles.txt.short
[General] Number of samples to process ......................: 1
[General] R1/R2 for raw reads present? ......................: False
[General] Be talkative (--verbose)? .........................: False

[Defining stretches] Min FF/RR coverage to qualify ..........: 10
[Defining stretches] Min length .............................: 50
[Defining stretches] Min dist between independent stretches .: 2,000
[Defining stretches] Num nts to pad a stretch ...............: 100

[Finding palindromes] Algorithm .............................: [will be dynamically determined based on sequence length]
[Finding palindromes] Min palindrome length .................: 10
[Finding palindromes] Max num mismatches ....................: 0
[Finding palindromes] Min mismatch distance to first base ...: 1
[Finding palindromes] Min distance betwen each sequence .....: 50

[Confirming inversions] Check all palindromes in a stretch? .: False
[Confirming inversions] Process only inverted reads? ........: False

[Genomic context] Recover and report genomic context? .......: True
[Genomic context] Gene caller to use ........................: prodigal
[Genomic context] Number of genes to consider ...............: 3

[Inversion activity] Compute inversion activity? ............: False
[Inversion activity] Not computing because ..................: `bams-and-profiles-txt` does not contain raw reads for r1/r2

[...]

PR #1971, opened 2022-08-22
more #1970 fix. if contig has too low cov, cont

watsonar:
The fix to skip a contig if it has a max coverage of 0 worked! But then we still ran into trouble if the max coverage was greater than zero but still not high enough to find inversions.


meren:
Excellent solution :) Goes right into the main branch! Thank you very much, Andrea.

PR #1976, opened 2022-08-31
check names in lazy init

mschecht:
`external-genomes.txt` genome names are not sanity checked during a lazy init. This PR ensures that contigs-db names are checked regardless of lazy or full init.


mschecht:
I ran this with `anvi-self-test --suite metagenomics-full --no-interactive` and `run_component_tests_for_pangenomics.sh`, both check out :)

ivagljiva:
The code looks good to me @mschecht! I also tested it with EcoPhylo using a small set of GTDB genomes with dot characters in their names, and it gave the proper Config Error :)

meren:
This looks good to me, too. Thank you very much, @mschecht.

PR #1977, opened 2022-09-06
Anvi'o docs += 'workflows'

meren:
Meren's attempt to add 'workflows' as a concept next to 'artifacts' and 'programs' in anvi'o docs.

Making the most of my time at the Roseobacter symposium.

_(needs more work, don't review yet :p)_


meren:
This is now ready, and the addition of the workflows to the anvi'o docs system can be seen here:

https://anvio.org/help/main/

Merging it to main, so we can continue working there.

PR #1994, opened 2022-10-14
HMI classification script

ivagljiva:
This PR adds the script `anvi-script-classify-hmi-genomes`, which can tell the user if their provided genome has high metabolic independence (HMI) or low metabolic independence (LMI). 

The way we score the level of metabolic independence is by adding up the completeness of the KEGG modules in the provided list (specifically, pathwise completeness). The script reports this score as well as the genome class, and the genome class depends on the provided threshold score.

It's a very basic script without many sanity checks, but it should be fairly straightforward for people to use. This PR includes the documentation page for the script which is hopefully clear enough for people without any knowledge of our FMT study to follow, while emphasizing that people shouldn't randomly use it without really thinking about the ecological concepts behind it. It was a tricky balance to strike, but I hope it makes sense :)

@meren, could you please take a look and make sure it's good enough? :)


ivagljiva:
Thank you very much for your changes, @meren :) It is much more simple now.

ivagljiva:
Thank you very much for your changes, @meren :) It is much more simple now.

Issue #1920, opened 2022-04-13
anvi-gen-contigs-database: gzip'ed fasta input


nick-youngblut:
## The need

It appears that `anvi-gen-contigs-database` cannot load gzip'ed fasta files:

```
File/Path Error: Someone is not happy with your FASTA file
                 '/path/to/file/assembly.fasta.gz' (this is
                 what the lib says: 'Fasta Lib Error: File
                 '/path/to/file/assembly.fasta.gz' does not
                 seem to be a FASTA file.'
```

## The solution

This should be a simple feature addition, depending on the fasta parsing code

## Beneficiaries

Most people gzip/bzip2 compress their genome fasta files, given the space savings

## Notes

I'm using anvio 7.1

nick-youngblut:
The same goes for `anvi-script-reformat-fasta`:

```
$ anvi-script-reformat-fasta -o assembly.fasta assembly.fasta.gz
Traceback (most recent call last):
  File "/tmp/global2/nyoungblut/code/Lreuteri_comp-genomics/conda_envs/Lreuteri-anvio/bin/anvi-script-reformat-fasta", line 216, in <module>
    reformat_FASTA(args)
  File "/tmp/global2/nyoungblut/code/Lreuteri_comp-genomics/conda_envs/Lreuteri-anvio/bin/anvi-script-reformat-fasta", line 75, in reformat_FASTA
    fasta = u.SequenceSource(args.contigs_fasta)
  File "/tmp/global2/nyoungblut/code/Lreuteri_comp-genomics/conda_envs/Lreuteri-anvio/lib/python3.6/site-packages/anvio/fastalib.py", line 104, in __init__
    raise FastaLibError("File '%s' does not seem to be a FASTA file." % self.fasta_file_path)
anvio.fastalib.FastaLibError: Fasta Lib Error: File 'assembly.fasta.gz' does not seem to be a FASTA file.
```

It appears that you are not including a decode for `if not self.file_pointer.read(1) == '>':` in `fastalib.py`.

For example:

```
def decode(x):
    try:
        return x.decode()
    except AttributeError:
        return x

[...]

if not decode(self.file_pointer.read(1)) == '>':
   [...]
```

meren:
Great idea. I will take care of this soon.

vinisalazar:
@meren would it be okay if I attempt a PR for this one?

meren:
[...]

PR #1997, opened 2022-10-25
Use 'rt' mode when reading compressed file

vinisalazar:
Hi,

This is an attempt at #1920.

In `SequenceSource.__init__`, use `gzip.open(fname, mode='rt')` when reading a compressed file. This prevents errors when looking at the first character of the file (which should be '>'). Both `anvi-gen-contigs-database` and `anvi-script-reformat-fasta` seem to be working with `gz` files now.

An alternative way would be to decode the first character to ensure it's a string, and not a byte, when comparing it to `'>'`.

Please let me know if you'd like me to modify/add anything.

Thanks,
V

### DIFF
--- a/anvio/fastalib.py
+++ b/anvio/fastalib.py
@@ -95,7 +95,7 @@ class SequenceSource():
         self.unique_next_hash = 0
 
         if self.compressed:
-            self.file_pointer = gzip.open(self.fasta_file_path)
+            self.file_pointer = gzip.open(self.fasta_file_path, mode="rt")
         else:
             self.file_pointer = io.open(self.fasta_file_path, 'rU', newline='')
 


meren:
I think this works perfectly :) Thank you so much for this change, @vinisalazar.

I will merge this branch now. But since I have you here, would you consider sending another PR for this file in the anvi'o repository?

https://github.com/merenlab/anvio/blob/master/anvio/data/misc/PEOPLE/CONTRIBUTORS.yaml

This information is rendered at anvio.org to generate the [anvi'o people](https://anvio.org/people/) page.

If you agree to add yourself, the CONTRIBUTORS.yaml file now has enough examples to follow, so it should be simple. The only thing in addition to the entry in that file is a **900x900px profile photo** of yours with your GitHub username under the AVATARS directory. There is more information here:

Thanks again for your help and consideration!

PR #1999, opened 2022-10-25
Write gzip output when input is also gzipped

vinisalazar:
Hi, 

this is a follow-up to #1920. When reformatting a gzip FASTA, also write the output as gzip.

**Summary of changes:**
  - Use mode 'wt' when creating 'output_file_obj' in class FastaOutput
  - Append '.gz' to output file if input file name is gzipped

Best,
V

### DIFF
--- a/anvio/fastalib.py
+++ b/anvio/fastalib.py
@@ -27,7 +27,7 @@ class FastaOutput:
         self.compressed = True if self.output_file_path.endswith('.gz') else False
 
         if self.compressed:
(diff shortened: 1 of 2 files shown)


meren:
🎉

vinisalazar:
That is a very fair point, do you have any suggestions on what could be a good way to address it? Would a warning message do?

meren:
I feel like the simplest way to solve it would be to exclusively work with the user output file name, rather then changing it :)

vinisalazar:
brb

PR #2000, opened 2022-10-25
Avoid unnecessarily changing the output file extension

vinisalazar:
Hi,

This is a follow-up to #1999. It should get rid of any unexpected behaviour. It worked on all the different scenarios I tried.

Best,
V


meren:
This is great. Thank you very much!

PR #2012, opened 2022-11-16
Fix help in anvi-run-hmms argument

vinisalazar:


meren:
Thank you very much for this change, @vinisalazar!

PR #2016, opened 2022-11-28
Pfam accession sanity checks

mschecht:
Hi everyone! I added the function `sanity_check_pfam_accessions` to `utils.py` to address an aspect of this [thread in discord today](https://discord.com/channels/1002537821212512296/1046482126490583190). I was looking for a definition online for Pfam accessions, but all I know for sure is that they start with "PF". To leave room to expand the sanity check, I threw it in `utils.py`. Let me know what you think!

Here is a quick test:

```bash
anvi-script-pfam-accessions-to-hmms-directory --pfam-accessions-list PF16867 asdf lkj -O asdf
```

### DIFF
--- a/anvio/utils.py
+++ b/anvio/utils.py
@@ -3969,6 +3969,21 @@ def sanity_check_hmm_model(model_path, genes):
                           "Here is a list of missing gene names: %s" % ', '.join(list(genes_in_model.difference(genes))))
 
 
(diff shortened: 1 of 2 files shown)


meren:
The beat approach in these kinds of situations is also to add a `--skip-sanity-check` optional and mention it as an option in the error message :) I'll merge it though since this is quite reasonable.

PR #2029, opened 2023-01-11
Sra download workflow

mschecht:
Hey everyone! I made a short but useful Snakemake workflow to download FASTQ files from SRA-accessions from NCBI. Briefly, it uses [NCBI sra-tools wiki](https://github.com/ncbi/sra-tools/wiki/08.-prefetch-and-fasterq-dump) to download the FASTQ files then gzips them using [pigz](https://zlib.net/pigz/). 

Please note that you will need to download the [NCBI SRA download tools](https://github.com/ncbi/sra-tools/wiki/01.-Downloading-SRA-Toolkit) and [pigz](https://zlib.net/pigz/). Luckily for pigz there is a [conda option](https://anaconda.org/conda-forge/pigz). However, the NCBI SRA toolkit is not very straightforward.

You test the workflow like this:

```bash
anvi-self-test --suite workflow-sra-download
```

Thanks for taking a look and let me know if you have any questions! I also wrote up documentation in our lab wiki [here](https://github.com/merenlab/lab-fiesta/wiki/Downloading-metgenomes-onto-Midway).


meren:
This is such an amazing contribution, @mschecht!

What happens when the SRA toolkit or pigz is not installed? I didn't see any checks or error messages, but if it is obvious to the user from the error messages that they need to install something, it is good enough :)

meren:
We can at pigz to the list of conda packages that are installed by anvi'o, or leave it as is and mention on the `workflow` page.

meren:
I think the *only* thing that is missing in this PR is a file called,

```
anvio/docs/workflows/sra-download.md
```

Which simply could be a version of the text in the internal wiki :)

mschecht:
Thanks for taking a look @meren!

> We can at pigz to the list of conda packages that are installed by anvi'o, or leave it as is and mention on the workflow page.

I think adding `pigz` to the anvi'o installation page would be the easiest solution. There could be other use cases for the tool elsewhere in the codebase.

> What happens when the SRA toolkit or pigz is not installed? 

Great point, I will add a sanity check for the SRA toolkit installation

mschecht:
@meren I've addressed all of your comments so far. Please take a look at `anvio/docs/workflows/sra-download.md` then I think this PR is ready.

meren:
I love Lorem ipsum, but I guess the one paragraph description needs some English words :p Apart from that, I think this is good to merge as far as I'm concerned. Thank you so much for taking the initiative in this, Matt.

PR #2031, opened 2023-01-20
anvi-run-cazymes

mschecht:
Hi everyone! I implemented two programs, `anvi-setup-cazymes` and `anvi-run-cazymes`, to annotate genes in contig-dbs with [dbCAN CAZymes](https://bcb.unl.edu/dbCAN2/download/Databases/). Unfortunately, CAZyme HMMs don't have `--cut-ga` values so the user will need to input an `e-value` to filter the hits via `anvi-run-cazymes --noise-cutoff-terms`.

```bash
anvi-self-test --suite run-cazymes
```

Thanks for taking a look!

### DIFF
--- a/anvio/cazymes.py
+++ b/anvio/cazymes.py
@@ -100,7 +100,7 @@ class CAZymeSetup(object):
 
     def is_database_exists(self):
         """Determine if CAZyme database has already been downloaded"""
-        if os.path.exists(os.path.join(self.cazyme_data_dir, f"dbCAN-HMMdb-{self.page_index}.txt")):
+        if os.path.exists(os.path.join(self.cazyme_data_dir, f"CAZyme_HMMs.txt")):
             raise ConfigError(f"It seems you already have CAZyme database installed in {self.cazyme_data_dir}, please use --reset flag if you want to re-download it.")
 
     def download(self, hmmpress_files=True):
--- a/anvio/cazymes.py
+++ b/anvio/cazymes.py
@@ -105,7 +105,11 @@ class CAZymeSetup(object):
         """Download CAZyme database and compress with hmmpress"""
         self.run.info("Database URL", self.database_url)
 
-        utils.download_file(self.database_url, os.path.join(self.cazyme_data_dir, os.path.basename(self.database_url)) , progress=self.progress, run=self.run)
+        utils.download_file(self.database_url, os.path.join(self.cazyme_data_dir, "CAZyme_HMMs.txt"), progress=self.progress, run=self.run)
+
+        message = f"CAZyme_HMMs.txt was downloaded from this URL: {self.database_url}"
+        with open(os.path.join(self.cazyme_data_dir, "version.txt"), 'w') as f:
+            f.write(message)
 
         if hmmpress_files:
             self.hmmpress_files()
@@ -113,7 +117,7 @@ class CAZymeSetup(object):
     def hmmpress_files(self):
         """Runs hmmpress on CAZyme HMM profiles."""
 
-        file_path = os.path.join(self.cazyme_data_dir, os.path.basename(self.database_url))
+        file_path = os.path.join(self.cazyme_data_dir, "CAZyme_HMMs.txt")
         cmd_line = ['hmmpress', file_path]
         log_file_path = os.path.join(self.cazyme_data_dir, '00_hmmpress_log.txt')
         ret_val = utils.run_command(cmd_line, log_file_path)
@@ -167,6 +171,16 @@ class CAZyme(object):
         if not self.cazyme_data_dir:
             self.cazyme_data_dir = os.path.join(os.path.dirname(anvio.__file__), 'data/misc/CAZyme')
 
+        self.hmm_file =  os.path.join(self.cazyme_data_dir, "CAZyme_HMMs.txt")
+        self.version_txt = os.path.join(self.cazyme_data_dir, "version.txt")
+
+        # Read version_txt to get CAZyme database version
+        with open(self.version_txt, 'r') as f:
+            output = f.read()
+        
+        URL = output.split(' ')[-1]
+        self.run.warning(f"Running CAZyme version: {os.path.basename(URL)}", lc='green', header="VERSION")
+
         self.is_database_exists()
 
         # reminder to be a good citizen (Good idea Iva!)
@@ -177,7 +191,7 @@ class CAZyme(object):
     def is_database_exists(self):
         """Checks if decompressed database files exist"""
 
-        if not glob.glob(os.path.join(self.cazyme_data_dir, "dbCAN-HMMdb-*.txt")):
+        if not glob.glob(self.cazyme_data_dir):
             raise ConfigError(f"It seems you do not have the CAZyme database installed in {self.cazyme_data_dir}, "
                               f"please run 'anvi-setup-cazymes' download it.")
 
@@ -194,10 +208,6 @@ class CAZyme(object):
 
     def process(self):
         """Search CAZyme HMMs over contigs-db, parse, and filter results"""
-
-        #FIXME: need a smarter way to find the CAZyme HMM file for when users 
-        # want a specific version. This will break if it's not V11
-        hmm_file = os.path.join(self.cazyme_data_dir, "dbCAN-HMMdb-V11.txt")
 
         # initialize contigs database
         class Args: pass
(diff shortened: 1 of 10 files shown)


anvio/cazymes.py
```diff
@@ -0,0 +1,252 @@
...
+        hmmer = HMMer(target_files_dict, num_threads_to_use=self.num_threads, program_to_use=self.hmm_program)
+        hmm_hits_file = hmmer.run_hmmer('CAZymes', 'AA', 'GENE', None, None, len(self.function_catalog), hmm_file, None, self.noise_cutoff_terms)
+
+        if not hmm_hits_file:
+            run.info_single("The HMM search returned no hits :/ So there is nothing to add to the contigs database. But "
+                            "now anvi'o will add CAZymes as a functional source with no hits, clean the temporary directories "
+                            "and gracefully quit.", nl_before=1, nl_after=1)
+            shutil.rmtree(tmp_directory_path)
+            hmmer.clean_tmp_dirs()
+            gene_function_calls_table.add_empty_sources_to_functional_sources({'Pfam'})
```
meren, line 217: This will clean up all the Pfam annotations from a contigs-db if there are no CAZyme hits, and then make it look like CAZymes were never run on that db :)

I understand that you used the template for anvi-run-pfams for this, which makes a lot of sense, but please search for Pfam in this file and make sure there are no residuals.

anvio/docs/programs/anvi-run-cazymes.md
```diff
@@ -0,0 +1,34 @@
...
+{:.notice}
+Unsure what an HMM is? Check out [our vocab page](http://merenlab.org/vocabulary/#hmm)
+
+To run, you'll need to provide a %(contigs-db)s. If you stored the %(cazyme-data)s that you got from running %(anvi-setup-cazymes)s in a custom location, you'll need to provide that path as well. The output is a %(functions)s artifact. 
+
+Here is a default run: 
+
+{{ codestart }}
+anvi-run-cazymes -c %(contigs-db)s \
+            --pfam-data-dir %(cazyme-data)s
```
meren, line 16: I think these are supposed to be `--cazyme-data-dir` in the docs :)

bin/anvi-run-cazymes
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == '__main__':
+    from anvio.argparse import ArgumentParser
+
+    parser = ArgumentParser(description=__description__)
+
+    parser.add_argument(*anvio.A('contigs-db'), **anvio.K('contigs-db'))
+    parser.add_argument('--noise-cutoff-terms', default=None, help="Choose a threshold for the CAZyme HMMs e.g. --noise-cutoff-terms \"-E 1\". \
+                        If you are not sure what to pick run `hmmsearch -h` or `hmmscan -h` and check out your options.")
```
meren, line 35: is `-E 1` a good idea as the default cutoff? I feel like we could be more stringent because I'm afraid no one will pay attention, and we will make people include too much garbage in their hits due to our defaults :(

bin/anvi-setup-cazymes
```diff
@@ -0,0 +1,48 @@
...
+
+from anvio.errors import ConfigError, FilesNPathsError
+
+
+__author__ = "Developers of anvi'o (see AUTHORS.txt)"
+__copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
+__license__ = "GPL 3.0"
+__version__ = anvio.__version__
+__authors__ = ['mschecht']
+__provides__ = ['pfams-data']
```
meren, line 17: `cazyme-data`? :)

anvio/tests/run_component_tests_for_cazymes.sh
```diff
@@ -0,0 +1,21 @@
...
+cd $output_dir/cazyme_test
+
+INFO "Running anvi-run-cazymes help menu"
+anvi-run-cazymes -h
+
+INFO "Running anvi-run-cazymes"
+anvi-run-cazymes -c B_thetaiotamicron_VPI-5482.db
+
+INFO "Running anvi-run-cazymes with --noise-cutoff-terms"
+anvi-run-cazymes -c B_thetaiotamicron_VPI-5482.db --noise-cutoff-terms "-E 1"
```
meren, line 21: I think this file should be accessible via the `anvi-self-test` program as well.

anvio/cazymes.py
```diff
@@ -0,0 +1,252 @@
...
+        ret_val = utils.run_command(cmd_line, log_file_path)
+
+        if ret_val:
+            raise ConfigError("Hmm. There was an error while running `hmmpress` on the Pfam HMM profiles. "
+                                "Check out the log file ('%s') to see what went wrong." % (log_file_path))
+        else:
+            # getting rid of the log file because hmmpress was successful
+            os.remove(log_file_path)
+
+class CAZyme(object):
```
meren, line 128: We don't have a citation statement for Pfams either, but I think it would be great to add something similar to our citation statements for other classes/programs so people would remember or see it in their logs that they should CAZymes in their methods sections etc.

meren (commented):
This is great, Matt! Thank you very much for creating this. Here are a few comments. I am very happy that this will make it into v8, but we need to make sure it is well tested and the default noise cutoff is appropriate.

anvio/cazymes.py
mschecht: Woops, good catch! I'll sift through the rest of the PR and purge `anvi-run-pfams`!

bin/anvi-run-cazymes
mschecht: At the moment there is no default cutoff, but I like the idea of making a stringent default. For now I will make the default:  `-E 1e-12`

anvio/cazymes.py
```diff
@@ -0,0 +1,259 @@
...
+        - self.page_index : string
+            version of CAZyme database
+
+        """
+        if self.args.cazyme_version:
+            self.page_index = self.args.cazyme_version 
+            self.run.info('Attempting to use version', self.args.cazyme_version)
+        else:
+            self.page_index = 'V11'
+            self.run.info_single('No CAZyme version specified. Using current release.')
```
ivagljiva, line 95: Is the current release always going to be 'V11'? :)

I think you should change this to 
```
self.run.info_single(f"No CAZyme version specified. Using {self.page_index}.")
```
to be more explicit in case we lag behind updating the default when CAZymes has an update.

anvio/cazymes.py
```diff
@@ -0,0 +1,259 @@
...
+        if hmmpress_file_extensions.sort() != extant_extensions.sort():
+            raise ConfigError("Anvi'o detected that the CAZyme database was not properly compressed with hmmpress. "
+                              "Please 'anvi-setup-cazymes --reset'")
+
+    def process(self):
+        """Search CAZyme HMMs over contigs-db, parse, and filter results"""
+
+        #FIXME: need a smarter way to find the CAZyme HMM file for when users 
+        # want a specific version. This will break if it's not V11
+        hmm_file = os.path.join(self.cazyme_data_dir, "dbCAN-HMMdb-V11.txt")
```
ivagljiva, line 200: Idea: you could rename the HMM file when you download it so that the name is always the same and not version-dependent. 

The only problem with this is that version information is then lost. Which you could potentially solve by adding a `version.txt` file within the CAZymes data folder so it can be looked up as needed.

anvio/docs/artifacts/anvi-setup-cazymes.md
```diff
@@ -0,0 +1,17 @@
...
+### Set up CAZyme data
+{{ codestart }}
+anvi-setup-cazymes 
+{{ codestop }}
+
+By default, this data is stored at `anvio/data/misc/CAZyme/`. To set up this data in a non-default location, run 
+{{ codestart }}
+anvi-setup-cazymes --cazyme-data-dir path/to/location
+{{ codestop }}
+
```
ivagljiva, line 12: I suggest adding a note in this section that if people want to download multiple versions of the CAZyme database, they can use different `--cazyme-data-dir`s to do it

ivagljiva (approved):
It looks great to me @mschecht :) I will test it on my computer and report to you if there are any issues (though I'm not expecting any)

anvio/cazymes.py
mschecht: Great idea, just implemented :)

anvio/cazymes.py
meren: `THIRD_PARTY_PROGRAMS` are only relevant for workflows, so it wouldn't be necessary unless it will appear in a workflow.

mschecht:
I think this is ready to go. Any more comments or suggestions? :)

ivagljiva:
It works for me :) in my opinion, it is ready to go.

meren:
We need to advertise it in the release notes, and everywhere.

PR #1892, opened 2022-02-21
anvi-report-inversions

FlorianTrigodet:
This PR introduces a WEALTH of new features added into the anvi'o framework to study active inversions in metagenomes. While it is close to being merged to the main repository, we have the following outstanding issues to be addressed in it:

 - [x]  make `--only-report-from` more flexible to use any contig's position. Right now, I can only use the exact start and stop positions that already match the parameter to find region of interest (min cov, min length, min dist between stretches, nts to pad). 
 - [x] compute inversion activity: if any reads are found for a given pair of inv/sample, then there should alway be a row for the true oligo (match reference contig) with min 0 reads and min 0% if no reads where found. 
- [x] Create anvi'o program docs for under `anvio/docs/programs/anvi-report-inversions.md` to briefly describe the major steps, key parameters, input file formats, example runs, and the resulting output files related to the `anvi-report-inversions`.

PR #2048, opened 2023-02-20
Anvi run batch

metehaansever:
I'm opening PR but if you see something missing I will fix it also.
My only question is, should we do another sanity check for the commands we receive from the user?
If so, should the sanity check here be the same as in the interactive.py?

### DIFF
--- a/anvio/batchwork.py
+++ b/anvio/batchwork.py
@@ -3,6 +3,7 @@
 """This is the main script for anvi-run-batch program."""
 
 import os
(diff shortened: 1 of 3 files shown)


meren:
you don't need to close the PR, @metehaansever, you can continue working on this branch, and the PR would be automatically updated. It is a good way to get the PRs evolve (like we did here, for instance: https://github.com/merenlab/anvio/pull/1892

PR #2059, opened 2023-03-22
A few functions for generic KEGG downloads (API only)

ivagljiva:
This PR provides a few new functions in the `KeggSetup` class for downloading 1) generic htext files, 2) generic flat files, and 3) all flat files within a given hierarchy from KEGG. They are only accessible when using anvi'o as an API (that is, no anvi'o programs make use of these functions). I implemented this at the request of @jessika-fuessel, who needed it for downloading KEGG COMPOUND entries, but I'm sure it will be more broadly useful. :)

Some documentation about using these functions is also included, as part of the `anvi-setup-kegg-kofams` help page (couldn't think of a better place for it at the moment). At some point if we get around to making API-specific docs, we should probably move it there.

Here is a brief example for using function (3) to get all flat files given a hierarchy accession:
```
python
import anvio
import argparse
from anvio import kegg
args = argparse.Namespace(reset=False)
setup = kegg.KeggSetup(args, skip_init=True)
setup.download_kegg_files_from_hierarchy('br08001', download_dir='KEGG_COMPOUND')
```

### DIFF
--- a/anvio/docs/programs/anvi-estimate-metabolism.md
+++ b/anvio/docs/programs/anvi-estimate-metabolism.md
@@ -94,7 +94,7 @@ If you have an unbinned metagenome assembly, you can estimate metabolism for it 
 anvi-estimate-metabolism -c %(contigs-db)s --metagenome-mode
 {{ codestop }}
 
-{: .notice}
+{:.notice}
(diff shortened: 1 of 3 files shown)


meren:
❤️ ❤️ ❤️

PR #2081, opened 2023-05-22
Ancient DNA Profiler

meren:
This PR adds a new functionality to `anvi-profile` to make it more suitable to report reliable variants from short reads that originate from ancient DNA samples.

By default, anvi'o will report variable nucleotides and their allele frequencies from **any nucleotide position** in a given short read found in the BAM file. Although, there are some applications where the observed variation in short reads depend on the location of the nucleotide positions in the read. For instance, in ancient DNA sequencing, the start and the end of short reads are often suffer from increased DNA damage, leading to a larger number of single-nucleotide variants when they are aligned to a reference sequence.

The parameter introduced in this PR, `--skip-edges`, enables the user to ask anvi'o to ignore a few number of nucleotides from the beginning and the end of any given short read during the calculation of SNVs. For instance, if you provide `5` as a parameter, the mismatches of a given read to the reference at the first and the last 5 nucleotides will **not** be reported in the variability table. The coverage of the contig, however, will not be impacted by the use of this parameter -- only what is reported as variants will differ, decreasing the impact of noise in specific applications.

For instance, here is an example with simulated data. I simulated a set of short reads where errors only occurred at the beginnings and ends of the reads, but I also included bona fide variants that occurred every 1000 nucleotides using https://github.com/merenlab/reads-for-assembly. Then, I recruited resulting simulated reads to the reference sequence, and generated a single BAM file.

This is the profiling of this BAM file with `anvi-profile` _without_ ignoring the extremities of the reads:

```
anvi-profile -i ANCIENT_DNA_SIMULATION.bam \
             -c contig.db \
             -S ANCIENT_DNA_SIMULATION_DEFAULT \
             --report-variability-full
```

image

And this is the profiling of the sae BAM file with `anvi-profile` that ignores the extremities of the reads `--skip-edges 5`:

```
anvi-profile -i ANCIENT_DNA_SIMULATION.bam \
             -c contig.db \
             -S ANCIENT_DNA_SIMULATION_TRIMMED \
             --report-variability-full
             --skip-edges 5
```

image

The coverages are identical between the two cases, and despite the poor choice in naming things, in fact nothing is trimmed :)

I'm pushing this PR so @genomewalker can take a look at it, and I'm happy to change the parameter and so on.

---

PS: we updated the flag from `--num-nts-to-ignore-from-short-read-ends` to `--skip-edges`, and the PR text is updated accordingly.


meren:
@genomewalker, I just realized that I haven't pushed this to the main branch. Should I merge this PR?

There is more work to do, such as 

- [ ] updating the `self` table,
- [ ] writing a migration script,
- [ ] and updating the docs for anvi-profile

but if the rest is OK as far as you're concerned, I'll start working on them whenever I can.

genomewalker:
@meren yes, please this is very useful, and it would be great to have it into the main branch :-)

meren:
All done and merged. Thank you!

PR #2087, opened 2023-06-21
UI improvements

metehaansever:
Blast Search Dropdown button was created.

### DIFF
--- a/anvio/data/interactive/css/inspection.css
+++ b/anvio/data/interactive/css/inspection.css
@@ -1,4 +1,6 @@
-
+.dp-button {
+    padding-left: 5px;
+}
 
 .dropdown-submenu {
     position: relative;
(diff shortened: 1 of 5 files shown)


meren:
This looks good, @metehaansever! Great job. I've been playing with it, and it seems to be working well.

image

I think it would be a good time to also fix those confusing function names: `get_sequence_and_blast` (which searches the nucleotide sequence of the gene call) vs `aa_sequence_blast` (which searches the amino acid sequence of the gene call). I think they should be renamed to `get_nt_sequence_and_search` and `get_aa_sequence_and_search`, respectively.

Of course, the next step to improve that is to consolidate them in a single function, say, `search_gene_sequence_in_remote_dbs`, which takes `NT` or `AA` as a parameter to determine what to search.

Can you also please test what happens when you want to search the amino acid sequence of a gene call that does not have an amino acid sequence (such as tRNAs)? @FlorianTrigodet can show you how to find an example gene for that if you can't figure it out by yourself :)

metehaansever:
I fixed the tRNA version of blast.
image

metehaansever:
Latest version.
image

PR #2097, opened 2023-07-27
add @ctb to anvi'o developers

ctb:
per @meren request


meren:
Thank you very much, @ctb! I'm about to merge it, but I wanted to ask if this is intentional (it seems a lot is being left for imagination here):

image

ctb:
ooops! fixed!

PR #2101, opened 2023-08-03
Updating the Python environment from 3.7 to 3.10

meren:
This PR attempts to update the Python dependency of anvi'o to 3.10.


anvio/homogeneityindex.py
```diff
@@ -46,7 +46,7 @@ def compute_functional_index(self, gene_cluster_sequences):
         max_score = 0
         for residue_number in range(0, num_residues):
             residues = []
-            for gene_sequence in range(0, num_sequences):
+            for gene_sequence in range(0, (num_sequences)-1):
```
meren, line 49: I'm not sure why this was necessary. We shouldn't have to make this change since Python's way of counting items in arrays could not have changed between 3.7 to 3.10. If this part of the code is exploding after the update to 3.10, there must have been something else gone wrong I thin.

meren:
When I change the Python dependency to 

``` diff
 dependencies:
-- python=3.7
+- python=3.10
```
I can't run

```
mamba env create -f .conda/environment.yaml
```

without getting this error:

```
Could not solve for environment specs
The following packages are incompatible
├─ fastani is installable with the potential options
│  ├─ fastani 1.33 would require
│  │  ├─ gsl >=2.7,<2.8.0a0 , which can be installed;
│  │  └─ openmp >=4.0  with the potential options
│  │     ├─ openmp 4.0.0 would require
│  │     │  └─ clangdev [4.0.0 |4.0.0|4.0.0.* ], which requires
│  │     │     └─ libxml2 2.9.*  with the potential options
│  │     │        ├─ libxml2 2.9.10 would require
│  │     │        │  └─ icu >=67.1,<68.0a0 , which can be installed;
│  │     │        ├─ libxml2 [2.9.10|2.9.9] would require
│  │     │        │  └─ icu >=64.2,<65.0a0 , which conflicts with any installable versions previously reported;
│  │     │        ├─ libxml2 [2.9.10|2.9.11|2.9.12] would require
│  │     │        │  └─ icu >=68.1,<69.0a0 , which can be installed;
│  │     │        ├─ libxml2 [2.9.12|2.9.14] would require
│  │     │        │  └─ icu [69.* |>=69.1,<70.0a0 ], which can be installed;
│  │     │        ├─ libxml2 [2.10.0|2.10.1|...|2.9.14] would require
│  │     │        │  └─ icu >=70.1,<71.0a0 , which conflicts with any installable versions previously reported;
```

The only way to avoid this outcome is to remove fastani dependency :/

I think this is a Mac OS associated problem.

meren:
This is how our help menus look like in Python 3.7:

image

And this is how they look like in Python 3.10:

image

MUST BE SARUMAN.

anvio/homogeneityindex.py
metehaansever: > I just tested the pangenomics workflow after removing this change, and it seems to be working fine, @metehaansever. Do you remember why you've made this change?

I made a change here because I got a list index out-of-range error.

metehaansever:
> When I change the Python dependency to
> 
> ```diff
>  dependencies:
> -- python=3.7
> +- python=3.10
> ```
> 
> I can't run
> 
> ```
> mamba env create -f .conda/environment.yaml
> ```
> 
> without getting this error:
> 
> ```
> Could not solve for environment specs
> The following packages are incompatible
> ├─ fastani is installable with the potential options
> │  ├─ fastani 1.33 would require
> │  │  ├─ gsl >=2.7,<2.8.0a0 , which can be installed;
> │  │  └─ openmp >=4.0  with the potential options
> │  │     ├─ openmp 4.0.0 would require
> │  │     │  └─ clangdev [4.0.0 |4.0.0|4.0.0.* ], which requires
> │  │     │     └─ libxml2 2.9.*  with the potential options
> │  │     │        ├─ libxml2 2.9.10 would require
> │  │     │        │  └─ icu >=67.1,<68.0a0 , which can be installed;
> │  │     │        ├─ libxml2 [2.9.10|2.9.9] would require
> │  │     │        │  └─ icu >=64.2,<65.0a0 , which conflicts with any installable versions previously reported;
> │  │     │        ├─ libxml2 [2.9.10|2.9.11|2.9.12] would require
> │  │     │        │  └─ icu >=68.1,<69.0a0 , which can be installed;
> │  │     │        ├─ libxml2 [2.9.12|2.9.14] would require
> │  │     │        │  └─ icu [69.* |>=69.1,<70.0a0 ], which can be installed;
> │  │     │        ├─ libxml2 [2.10.0|2.10.1|...|2.9.14] would require
> │  │     │        │  └─ icu >=70.1,<71.0a0 , which conflicts with any installable versions previously reported;
> ```
> 
> The only way to avoid this outcome is to remove fastani dependency :/
> 
> I think this is a Mac OS associated problem.

Actually, this installation not totally working with Ubuntu. We realize this with Alex today. 
```
mamba env create -f .conda/environment.yaml
 ```

File: anvio/homogeneityindex.py (hunk 1 above)
Comment (meren) [reply anvio/homogeneityindex.py:49]: Can you try again running

```
anvi-self-test --suite pangenomics -o PAN-TEST
```

after reverting this change and see if you are getting that error again? There must be something else going wrong before this.

Comment (metehaansever) [reply anvio/homogeneityindex.py:49]: I receive another error right now, 

```
Config Error:: An anvi'o function needs 'average_nucleotide_identity.py' to be installed on   
              your system, but it doesn't seem to appear in your path :/ If you are certain  
              you have it on your system (for instance you can run it by typing              
              'average_nucleotide_identity.py' in your terminal window), you may want to send
              a detailed bug report. Sorry! 
```
I'll test and fix this change as soon as I fix this problem.

PR #2108, opened 2023-08-25
Sra workflow update

FlorianTrigodet:
I updated the workflow to be able to handle two issues:
 - Single read samples
 - SRAlite format

The workflow was expected specific output, like and SRRXXXXX_1.sra SRRXXXXX_1.fastq.gz + SRRXXXXX_2.fastq.gz. 
I mostly just changed the input and output to .done files. Now if an SRA is only available in SRAlite format, it will still be processed. And if the SRA is single reads, this format can be handled too. 

I have also added a final rule which creates either a samples.txt and/or samples_single_reads.txt (only has the column sample and r1).

And I updated the test to include a SRA accession for a single-reads sample.


meren:
Thank you for making these changes, @FlorianTrigodet. They look good to me, but I will let @mschecht to merge it.

mschecht:
Thanks a lot, @FlorianTrigodet! Great idea adding the `samples.txt` output.

I did a little polishing and we are ready to merge.

Issue #1738, opened 2021-05-13
anvi-setup-ncbi-cogs issue


meren:
# Why are you here?

Probably anvi'o sent you here so you can help us address this issue.

# Summary

This is a problem we have not been able to address, so we decided to collect more data from people to understand this enigmatic error ([example](https://groups.google.com/g/anvio/c/t-my0SUq4JY/m/tvIOT8-eBAAJ), or these: #1686, #1671, #1647) that usually happens around here in the code while running `anvi-setup-ncbi-cogs`:

```
Traceback (most recent call last):
  File "/home/australomics/anaconda3/envs/anvio-7/bin/anvi-setup-ncbi-cogs", line 47, in <module>
    setup.create()
  File "/home/australomics/anaconda3/envs/anvio-7/lib/python3.6/site-packages/anvio/cogs.py", line 617, in create
    self.setup_raw_data()
  File "/home/australomics/anaconda3/envs/anvio-7/lib/python3.6/site-packages/anvio/cogs.py", line 831, in setup_raw_data
    self.files[file_name]['func'](file_path, J(self.COG_data_dir, self.files[file_name]['formatted_file_name']))
  File "/home/australomics/anaconda3/envs/anvio-7/lib/python3.6/site-packages/anvio/cogs.py", line 660, in format_p_id_to_cog_id_cPickle
    COG = fields[6]
IndexError: list index out of range
```

# Do we still need your files?

Yes, please send the full error message and the files anvi'o requested you to send in its error message below as a comment. Thank you!

watsonar:
Hello there. Anvi'o did indeed send me here, with this error message as a gift!

Full command and error message, using anvio-dev:
```
$ anvi-setup-ncbi-cogs --reset

COG version ..................................: COG20
COG data source ..............................: The anvi'o default.
COG base directory ...........................: /Users/andrea/github/anvio/anvio/data/misc/COG

WARNING
===============================================
This program will remove everything in the COG data directory, then download and
reformat everything from scratch.

Downloaded successfully ......................: /Users/andrea/github/anvio/anvio/data/misc/COG/COG20/RAW_DATA_FROM_NCBI/cog-20.cog.csv
Downloaded successfully ......................: /Users/andrea/github/anvio/anvio/data/misc/COG/COG20/RAW_DATA_FROM_NCBI/cog-20.def.tab
[...]

PR #2110, opened 2023-08-28
Add checksum checking for NCI downloads in `anvi-setup-ncbi-cogs`

Ge0rges:
I added checksum verification to `anvi-setup-ncbi-cogs`. If fails, it prints an error. This handles the fact that `COG2014` does not have a checksum file as well. This feature is to handle the incorrect downloads often placing this command due to NCBI servers. See #1738. 

While not added here, I find it surprising that the function `download_file` in `utils.py` does not ever check if `file_size == downloaded_size`. I leave it up to the maintainer to determine if that check is necessary, or if it exists elsewhere. 

PS: I was able to test both successful and unsuccesful.

### DIFF
--- a/anvio/cogs.py
+++ b/anvio/cogs.py
@@ -853,7 +853,7 @@ class COGsSetup:
         checksums = {}
         for line in open(input_file_path, 'rU').readlines():
             stripped = line.strip('\n').split(' ')
-            file_name = stripped[-1].strip('*')[-1]
+            file_name = stripped[-1].strip('*')
             checksums[file_name] = stripped[0]
 
         # Print warning if checksums are not provided by NCBI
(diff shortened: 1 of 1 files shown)


meren:
Thank you very much for this PR, @Ge0rges! It looks reasonable to me and I will merge it immediately so we can test it a bit. It is a shame that COG14 does not have checksums. Since those files never change, in theory, one could generate a checksums file from a successful download, store that file in anvi'o codebase (somewhere under `data/misc`, I guess), and set the access for that file for the `COG14` entry in `self.cog_files`. This way even if people are working with COG14, they still could benefit from the same solution.

> While not added here, I find it surprising that the function `download_file` in `utils.py` does not ever check if `file_size == downloaded_size`. I leave it up to the maintainer to determine if that check is necessary, or if it exists elsewhere.

That is true. Content-Length is often not in response headers, so such a check would only work for a fraction of files that function downloads, but still, that could have been useful at least in some instances :)

Thank you again!

PR #2112, opened 2023-08-29
Checksum checking for COG14 and addition to developers.yaml

Ge0rges:
Following up on #2110, this PR adds a checksum file I generated for COG14. The checksum file should be checked with somebody else's existing download of COG14. I tested it with two downloads and it seems to have worked.

It also adds me onto the `developers.yaml` file.


meren:
Thank you very much for these changes, @Ge0rges. I will move the file somewhere else after merging and change the name to avoid confusion, but this is a great solution that cover all bases :)

PR #2134, opened 2023-09-29
gene popover functionality

metehaansever:
Since the length value was not returned from the meta.json file, I left it as static. I guess there is no Annotations value either. In addition, for missing values in some genes, for example, id: 28, we can return the values to none.

image

### DIFF
--- a/anvio/data/misc/PEOPLE/DEVELOPERS.yaml
+++ b/anvio/data/misc/PEOPLE/DEVELOPERS.yaml
@@ -332,6 +332,10 @@
   avatar: efogarty11.jpg
   bio: "Microbiologist developing computational and wet lab strategies to study horizontal gene transfer in the human gut microbiome."
   affiliations:
+    - title: Clinical Affairs Associate
+      inst: Theralink Technologies, Inc.
+      inst_link: https://theralink.com/
+      current: true
     - title: Graduate Student
       inst: Committee on Microbiology at the University of Chicago
       inst_link: https://biomedsciences.uchicago.edu/page/microbiology
--- a/anvio/data/static/template/inversions.tmpl
+++ b/anvio/data/static/template/inversions.tmpl
@@ -163,7 +163,7 @@
                                                 {% for function in gene|lookup:"functions" %}
                                                 <tr>
                                                     <td class="long">{{ function|lookup:"source"|pretty }}</td>
-                                                    <td class="short">{{ function|lookup:"accession"|pretty }}</td>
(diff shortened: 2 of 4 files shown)


FlorianTrigodet:
I'll add length in the json. 
I'm also working on adding DNA and AA sequences into the json. But that requires me learning more about anvi'o class etc.

metehaansever:
> I'll add length in the json. I'm also working on adding DNA and AA sequences into the json. But that requires me learning more about anvi'o class etc.

In that case, we can merge this branch after we finish the DNA and AA functionality.

meren:
How come there is no annotation but there is an accession for the annotation? :)

FlorianTrigodet:
Now they do! Annotations are called 'function' in the json

meren:
@metehaansever, please make sure the popovers has a fixed width of 900px. The contents of the annotation column should wrap in each `td`.

FlorianTrigodet:
DNA and AA sequence are available for each gene. 
They are accessible as `DNA_sequence` and `AA_sequence` in the gene part of the json.
Now we can make a button to display these sequences :D

FlorianTrigodet:
Oh, should I provide a fasta header like in inspect page?

metehaansever:
> Oh, should I provide a fasta header like in inspect page?

I guess, this is a good idea because they'll copy and paste somewhere else. But I guess I can handle that on the UI :)

FlorianTrigodet:
Now there is a fasta header in the summary (called 'header' because I have a lot of imagination this morning). 
I also fix the DNA reporting part because if the gene is in rev direction, the sequence should be reverse complemented. 
Ready for the DNA and AA button :D

metehaansever:
we replaced the popover top to the bottom. The reason was the Function's width. And I added AA/DNA buttons.
image

meren:
Some attention to detail please: can anyone answer why the borders for the inner div are not equally distant to the borders of the popover at all edges? No. No one can answer that. But they can make sure it is not the case :)

image

metehaansever:
I hope, it's not the case anymore :+1: 
image

FlorianTrigodet:
Shame on WHOEVER could have forgotten a print statement in the code ...

Looks nice to me with the latest changes. Ready for merge do you think?

meren:
Yes! Thank you both :) Merge whenever, boss.

meren:
(but please don't forget to update the documentation with some screenshots or whatever, so it is clear what we're generating)

PR #2138, opened 2023-10-04
Inversion motif summary

FlorianTrigodet:
Hello, what about a new branch about inversions? 🙃 

I have updated the code to parse the motif result and add information to the consensus summary and the html summary. 

image

I have also included updated documentation, especially regarding the html output. 
I need to update the inversion-txt artifact.

The code is functional, but probably not elegant and for sure not fool proof. 

Thanks @metehaansever for your life saving help 🙏


meren:
Looks like a great start!

PR #2155, opened 2023-10-27
Added support for AA files for anvi-run-ncbi-cogs

Ge0rges:
Hello,

I wanted to allow `anvi-run-ncbi-cogs` to be given just an amino acid sequence from a FASTA file. I believe I've done this successfully. 

However this pull request requires still one major change which is the output. I was unsure what the best output format/options to offer would be and wanted to seek the advice of the anvi'o team.


Ge0rges:
Perhaps a TSV would be a useful output , where the first column is the gene ID which would match back to the FASTQ sequence, and every column is another of the variables present in the dict? So final product would have one row per sequence in the FASTA file given.

Ge0rges:
Ok after discussion with @ivagljiva the output is now a TSV. I think this is good.

bin/anvi-run-ncbi-cogs
```diff
@@ -19,9 +19,10 @@ __copyright__ = "Copyleft 2015-2018, the Meren Lab (http://merenlab.org/)"
 __license__ = "GPL 3.0"
 __version__ = anvio.__version__
 __authors__ = ['meren']
-__requires__ = ['contigs-db', 'cogs-data',]
+__requires__ = ['cogs-data',]
 __provides__ = ['functions',]
```
meren, line 23: Provides should now include `functions-txt`, as well :)

bin/anvi-run-ncbi-cogs
```diff
@@ -47,12 +49,14 @@ if __name__ == '__main__':
...
 
-    parser.add_argument(*anvio.A('contigs-db'), **anvio.K('contigs-db'))
+    parser.add_argument('-c', '--contigs-db', metavar='CONTIGS_DB', required=False,
+                        help="Anvi'o contigs database generated by 'anvi-gen-contigs-database'")
     parser.add_argument(*anvio.A('cog-version'), **anvio.K('cog-version'))
     parser.add_argument(*anvio.A('cog-data-dir'), **anvio.K('cog-data-dir'))
     parser.add_argument(*anvio.A('num-threads'), **anvio.K('num-threads'))
     parser.add_argument(*anvio.A('temporary-dir-path'), **anvio.K('temporary-dir-path'))
 
+    parser.add_argument('--aa-file', default=None, metavar="AA_FILE", help="An amino acid fasta file to annotate.")
```
meren, line 59: With this PR you in fact implement two different entry routes to the program that are not compatible. One is the old way of providing a contigs-db, and the other is used when the user provides a FASTA file. I think it would be most helpful to change the help menu in such a way it is clear. For some inspiration, please tak ea look at the output of,

```
anvi-estimate-metabolism -h
```

and input options there.

I think the input option for FASTA should also require an output file name, so the output is not always named `cog_functions.tsv`, and give user control.

anvio/cogs.py
```diff
@@ -198,6 +199,16 @@ def process(self, aa_sequences_file_path=None):
         if self.contigs_db_path:
             utils.is_contigs_db(self.contigs_db_path)
 
+        if aa_sequences_file_path:
+            filesnpaths.is_file_exists(aa_sequences_file_path)
+
+            fasta = u.SequenceSource(aa_sequences_file_path); next(fasta) 
+
+            sequence = fasta.seq
+            utils.is_gene_sequence_clean(sequence, amino_acid=True, must_start_with_met=False)
```
meren, line 208: Thank you very much for finding these functions in utils and making good use of them! Great investigative work :)

Here we are only checking the first sequence. I wonder if it is good enough scrutiny :p I know the trade-off here is compute time for 1 million sequences, but we did learn the hard way quick&dirty checks often lead to a lot of cryptic messages later when things fail.

anvio/cogs.py
```diff
@@ -198,6 +199,16 @@ def process(self, aa_sequences_file_path=None):
         if self.contigs_db_path:
             utils.is_contigs_db(self.contigs_db_path)
 
+        if aa_sequences_file_path:
+            filesnpaths.is_file_exists(aa_sequences_file_path)
+
+            fasta = u.SequenceSource(aa_sequences_file_path); next(fasta)
```
meren, line 205: `fasta.close()` should be run at some point after this for best practice.

meren:
Hey @Ge0rges! Thank you very much for this PR! I added some 2 cents here and there :)

Please add yourself to the authors column of the program, too!

bin/anvi-run-ncbi-cogs
```diff
@@ -33,7 +34,8 @@ pp = terminal.pretty_print
 @time_program
 def main(args):
     cogs = COGs(args)
-    cogs.process()
+    aa_file = args.aa_file or None
+    cogs.process(aa_sequences_file_path=aa_file)
```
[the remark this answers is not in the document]
Ge0rges: The process function might become confusing now however as the case where "aa_sequences_file_path" and "--aa-file" arguments are both passed may be ambiguous. What was the role of this argument prior to my commit? I didn't identify it as doing anything other than allowing for cogs to consider a file input directly, thus my use of it. My solution to this comment would be to add a class variable that is set to the "--aa-file" argument in init, and add a check if both the argument and the class variable are provided, error out. Would that work?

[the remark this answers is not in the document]
meren: Ah, yes, that is indeed confusing. I'm very sorry I missed this.

I now realize the `aa_sequences_file_path` is produced by the class when a contigs-db is provided because BLAST or DIAMOND must be called to run on _that_ file. What you did is the right solution, and please ignore my comment on this item :)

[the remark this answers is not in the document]
meren: I will have to come back to this later to change the design a little so everything is a bit more tidy, but obviously that is independent of your PR. Thanks for your patience.

bin/anvi-run-ncbi-cogs
```diff
@@ -47,12 +49,14 @@ if __name__ == '__main__':
 
     parser = ArgumentParser(description=__description__)
 
-    parser.add_argument(*anvio.A('contigs-db'), **anvio.K('contigs-db'))
+    parser.add_argument('-c', '--contigs-db', metavar='CONTIGS_DB', required=False,
+                        help="Anvi'o contigs database generated by 'anvi-gen-contigs-database'")
```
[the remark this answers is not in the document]
Ge0rges: Much cleaner. Fixed.

bin/anvi-run-ncbi-cogs
Ge0rges: Done.

bin/anvi-run-ncbi-cogs
Ge0rges: Cleaned up the help menu as suggested (much better). An added what should've been an obvious output file path.

anvio/cogs.py
Ge0rges: Done.

anvio/cogs.py
Ge0rges: Don't want to waste downstream computation, checked every sequence.

Ge0rges:
@meren incorporated the very logical feedback. Let me know if can do anything more.

meren:
Thank you very much for these changes, @Ge0rges! And apologies for the delayed responses here. The last few days have been hectic :) I am merging the PR now. Thank you again!

meren:
Looks nice :)

image

It would have been nice to update anvio/docs/programs/anvi-run-ncbi-cogs.md to mention that it can also be run on a FASTA file directly.

Ge0rges:
Yes! Sorry I'm still learning about all the different places Anvi'o stores information.

meren:
No worries at all. Feel free to add to this one instead of making a new one. Completely up to you. If you don't want to do it, I can add the information there at some point, too :)

Ge0rges:
I didn't know I could contribute to a closed pull request. However, I have pushed the change to my fork.

meren:
Hmm. Then feel free to send another PR, @Ge0rges. Thank you!

PR #2162, opened 2023-10-31
Updated docs for anvi-run-ncbi-cogs

Ge0rges:
An addendum to the previous pull request.


anvio/docs/programs/anvi-run-ncbi-cogs.md
```diff
@@ -2,7 +2,7 @@ This program **annotates genes in your %(contigs-db)s with functions using NCBI'
 
 This program assumes that the user has successfully set up the COGs database on their computer using the anvi'o program %(anvi-setup-ncbi-cogs)s.
 
-The only critical parameter to %(anvi-run-ncbi-cogs)s is a %(contigs-db)s. The program will store its output in the %(contigs-db) sas a %(functions)s artifact.
+This program requires one of two possible inputs a %(contigs-db)s or a %(fasta). If a %(contigs-db) is provided, the program will store its output in the %(contigs-db) as a %(functions)s artifact. Otherwise you must specify an outputfile using `-o`, to which the program will write a tab delimited file containing all the annotatons found.
```
meren, line 5: I know this is a bit cryptic so I thought I'd mention here to shed some light on something subtle:

The correct notation is `%(fasta)s` and `%(contigs-db)s` and so on. This is because these markdown files are rendered twice -- one, when we run `anvi-script-gen-help-pages` to generate a markdown outputs, and two, when those markdown files are rendered for web.

For instance, if you run the following command on your computer after activating the anvi'o environment, you can see the output:

```
anvi-script-gen-help-pages -o anvio-docs
```

Note: if you do that you will also see that Python will complain that `%(fasta)` causes a syntax error, but `%(fasta)s` works smoothly simply because at that stage of rendering these named variables are populated with content from a dictionary (i.e., see them as `%s` vs %(key)s when a dictionary template are accessed).

I'll change these in the main branch after merging so you don't have anything to worry about. I'm just over explaining things that are not adequately explained elsewhere :)

meren: Just FYI, the exact place in the code where this first step of rendering takes place is here:

https://github.com/merenlab/anvio/blob/master/anvio/programs.py#L728

Ge0rges: Thank you for explaining that!

PR #2174, opened 2023-11-15
Allow pre-computed inversions to be searched in new metagenomes without profile-db and bam-file artifacts

### LINKED ISSUE #2161: [BUG] Issue with --pre-computed-inversions
`anvi-report-inversions --pre-computed-inversions` does not work if you don't have associated bams and profiles. But they are not needed to recompute activity. 

## anvi'o version

anvio dev version.

The command `anvi-report-inversions` has a flag `--pre-computed-inversions` which allow user to use a precomputed inversions summary and only run the activity part of the code which uses R1 and R2 short reads. 

For instance I ran the code using a first set of sample and I want to use `--pre-computed-inversions` using another set of samples, but I don't have bams and profiles associated with these samples. And I don't want to make them: they are not needed to compute the activity.

Then I get an error saying that my bam and profile.db are (of course) missing from my bams-and-profiles.txt

I am not sure what is the best thing to do here. Should the code skip checking for bam and profiles, should the user just not fill the bam and profile column of the bams-and-profiles.txt. SHOULD WE CHANGE THE ENTIRE CODE and put the path to R1 an R2 in a samples.txt??? 

Any input is welcome :D

meren:
This PR closes #2161. Much progress.

### DIFF
--- a/anvio/docs/programs/anvi-report-inversions.md
+++ b/anvio/docs/programs/anvi-report-inversions.md
@@ -124,6 +124,9 @@ anvi-report-inversions -P %(bams-and-profiles-txt)s \
 
 In this mode, %(anvi-report-inversions)s will not recalculate inversions, and only report the activity of inversions found in the input file across samples listed in the %(bams-and-profiles-txt)s file.
 
+{:.notice}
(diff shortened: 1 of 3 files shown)


FlorianTrigodet (commented):
🫠

PR #2175, opened 2023-11-15
Fix `anvi-script-reformat-fasta` amino acid check

Ge0rges:
As discussed on discord, the script was not using the correct AA alphabet. That is fixed here along with another bug regarding the existence of flags causing the script to skip over some reads. Before this is merged we should also consider whether the addition of ambiguous characters to the AA alphabet is warranted (e.g. `BZJ`).

### DIFF
--- a/sandbox/anvi-script-reformat-fasta
+++ b/sandbox/anvi-script-reformat-fasta
@@ -98,19 +98,19 @@ def reformat_FASTA(args):
         exclude_ids = set([l.split('\t')[0].strip() for l in open(args.exclude_ids, 'r').readlines()])
         run.info('Input IDs to remove', '%d found' % len(exclude_ids))
     else:
-        exclude_ids = set([])
+        exclude_ids = None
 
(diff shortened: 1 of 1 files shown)


meren:
You're the best, @Ge0rges! Thank you very much for catching the bug and fixing it :)

Ge0rges:
@meren what do you think about the ambiguous character question?

meren:
Ah, sorry, I completely misread that sentence. I am not sure what to suggest for that. Under what circumstances it becomes a necessity? When we convert single-letter alphabets to verbose names?

Ge0rges:
The case I ran into is that I have protein sequences with these characters (so they are still valid but not seen as so), that I give to `anvi-run-ncbi-cogs` which it also sees as invalid due to the use (I think) of `utils.utils.is_gene_sequence_clean` (and/or another part of the script). But I think these characters should be deemed valid inputs for anvio programs.

PR #2193, opened 2023-12-22
Structure fix

metehaansever:
First of all, 
we were receiving http 301 from ngl package (https://cdn.rawgit.com/arose/ngl/v0.10.4/dist/ngl.js) source is updated. 

Second problem is look like backend problem atleast Dev console says so.
image
I just fix with work around solution for now and it is **working**.

I couldnt find any error on the backend code (Interactive>get_model_info || bottleroutes.py>get_model_info).
I'll merge that and hopefully fix the problem on the new UI.


meren:
Thanks for fixing this, @metehaansever!! :)

metehaansever:
I just fix the Backend issue :)

PR #2194, opened 2023-12-22
Fix protein structure problem

metehaansever:
**Error Message:** 
    raise KeyError(f"{not_found} not in index")
KeyError: "['percent_similarity', 'align_fraction'] not in index"

I just handle the case when columns are not present.

### DIFF
--- a/anvio/data/interactive/js/structure.js
+++ b/anvio/data/interactive/js/structure.js
@@ -723,9 +723,8 @@ function load_model_info() {
             defer.resolve();
         },
         error: function(xhr, status, error) {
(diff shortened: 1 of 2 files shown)


meren:
Why are they not present in our test case though? I'm just curious to know what led to this error at the first place when we didn't change anything in the codebase that should effect this section.

metehaansever:
> Why are they not present in our test case though? I'm just curious to know what led to this error at the first place when we didn't change anything in the codebase that should effect this section.

To be honest, I have no idea. I asked the same question and focused on Front-end to solve the problem but it wasn't the case. I couldn't find any clue how it didn't occur before.

PR #2198, opened 2024-01-16
Inversions activity issue

FlorianTrigodet:
I added a check if activity is missing for one oligo_primer (first or second).
The main fix is the second commit. We need to always have an entry for the reference oligo in the activity table. And if there is no activity (short reads) matching the reference oligo, we create an entry with frequency 0. We also have a filter with a min frequency value to report the activity (avoid very rare and likely irrelevant oligo). In rare cases, a ref oligo can have a frequency below the threshold and is not included in the final table, and no entry with frequency 0 is created. 
The html part of the code expect an entry for the ref oligo. I corrected and tested the fix and it works 🎉

### DIFF
--- a/anvio/inversions.py
+++ b/anvio/inversions.py
@@ -938,18 +938,20 @@ class Inversions:
 
             for inversion_id in self.summary['inversions']:
                 for sample in self.summary['inversions'][inversion_id]['activity']:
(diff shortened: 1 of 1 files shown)


meren:
Excellent! :)

PR #2222, opened 2024-02-15
Convert photo to 900 x 900 px

Kekananen:
- Fixing Aspect ratio of photo


ivagljiva:
Thanks for this, @Kekananen!

Issue #2043, opened 2023-02-09
[BUG] Support values can be misplaced in the interactive interface


tdelmont:
## Short description of the problem

Support values in the interactive interface can be wrong, due to a well known problem.

 
## anvi'o version

anvi'o v7.1

## System info

Mac

## Detailed description of the issue

We realised that bootstrap values can be misplaced in the interactive interface. This is especially true when playing around with re-rooting. This is a well know phenomenon, described in this paper: https://academic.oup.com/mbe/article/34/6/1535/3077051

## Files to reproduce

Here is are files that can be used for testing:
[Archive.zip](https://github.com/merenlab/anvio/files/10694187/Archive.zip)

meren:
You are the best, Tom. Thank you for reproducible examples.

meren:
I hope you would be available for questions later, too.

metehaansever:
I can't see all the things in the Branch Support section.
I try with that piece of command:

`anvi-interactive --manual -p PROFILE.db -t DNApolB_NCLDV_marine_db_v1.Darius_EukaOnly_exceptAlpha.wMedusa.noLB5.Duplodna.Baculo.nr95.corrected.removedSeqTreemmed_Tom.noLoneSeq2.fftnsi.trim50.fa.treefile -d metadata_duplodna.txt 
`

![Screenshot from 2023-02-09 11-50-49](https://user-images.githubusercontent.com/24258998/217792575-88ce1864-2412-4438-ab29-ed7e9de4ed9f.png)

tdelmont:
You should be able to see the support values using the mouse (right side of the interface) and pointing it to the branche splits within the tree. 

The problem is that the scores provided by anvi'o do not necessarily match to those in the tree itself (named DNApolB_NCLDV_marine_db_v1.Darius_EukaOnly_exceptAlpha.wMedusa.noLB5.Duplodna.Baculo.nr95.corrected.removedSeqTreemmed_Tom.noLoneSeq2.fftnsi.trim50.fa.treefile)

tdelmont:
I just understood. In our case, we have 2 support values and because of that none are displayed, which is for another request.
But again, what we have done is explore the values using the mouse in the tree, interactively. 

Sorry, since all our work is done using 2 support values, I do not have a better example for you.

metehaansever:
> You should be able to see the support values using the mouse (right side of the interface) and pointing it to the branche splits within the tree.
> 
[...]

PR #2224, opened 2024-02-16
Newick rooting with branch support value

FlorianTrigodet:
Hello everyone and welcome for our new and final episode of the newick tree branch support value problem!

# Background
The problem: when rooting a tree, some branch support values have to be moved around. See the issue #2043 for more information and link to a paper explaining the situation. 

The origin of the problem is that the "support values" were stored as node.name and not node.support in bottleroutes.py's reroot_tree. ete3 is perfectly capable of replacing branch support values (and that's a good news!).

# The solution
You could be tempted to just copy node.name (which contains the real support value) into node.support, root the tree and maybe rename node.name with node.support after they were moved appropriately by ete3. And it would work. As long as your support values are float. 

But the world is complex and Tom's support values looks like that: '100/99.6'. Basically a string. It is ok, in the code below you will see a hack™ that pair unique integer to node.name, use unique integer as support value, track their movement after rooting and rename node.name accordingly. Now, support value format is not a problem and Tom's tree can be properly rooted and rerooted, the support value will be rightfully placed. 

# Test case
You don't need to trust me, you just need to test with the following example which is directly copied from the paper mentioned earlier. The fist digit of the string test case matches the 1, 2, and 3 support value of the original test case:
```
# support value as number
echo "((C,D)1,(A,(B,X)3)2,E)R;" > tree_support_as_number.newick
anvi-interactive --manual -t tree_support_as_number.newick -p profile_support_as_number.db

# or support value as string
echo "((C,D)1/98.5,(A,(B,X)3/777)2/5546,E)R;" > tree_support_as_string.newick
anvi-interactive --manual -t tree_support_as_string.newick -p profile_support_as_string.db
```

Original anvi'o (WRONG) behavior after rooting using X:

Screenshot 2024-02-16 at 10 45 06

Now:
Screenshot 2024-02-16 at 10 45 32

Also works if support value is a string, but we cannot display it (more on that later), but using the panel Data, one can see that there is no problem and the string starting with "2" is in the lower branch compared to the wrong behavior:
Screenshot 2024-02-16 at 10 50 42

# Improvement needed
## The zeros
We can see some 0 where we don't have support value. This is a default behavior in anvi'o and as nothing to do with the rooting. To be fair, if one has support values for their tree, they have values for every branch and it shouldn't be a problem. Anyway, there should be nothing displayed.

## What is node.name is REALY the node's name
Let's imagine the user has a tree with no support values, but the non-leaf node have a name. Then you don't want to move the node's name like we move the support values after rooting!!!

I propose the following: 
Screenshot 2024-02-16 at 10 21 47

If you click on the first option, the string associated with each node is considered support value and will be reassigned after rooting. The second option will not move the node's name. 

## What if I have BOTH node name and support value
Then I don't like you and it is personal at this point. 
But seriously, it is not possible. These info are stored in the same location in the newick tree structure. You either treat it as node name OR support value. So we are good. 

## I want to display my support value when they look like Tom's (100/99.6)
Me too. I have checked the code a little bit, and I have noticed that there is a js function that check if branch support value as a min and max value. If not, you get this warning when you want to display the support values:
Screenshot 2024-02-16 at 11 10 45

The need to check for max and min are for two things: define range of value to display and display a symbol (round shape) proportional to the support value:
Screenshot 2024-02-16 at 11 12 07

Now, I would like the front end to check if branch support is number or string. If number, show the same as above. But if non digit character, display the text only (no range filtering, no display symbol):
Screenshot 2024-02-16 at 10 26 29
Screenshot 2024-02-16 at 11 15 27

# Summary TODO
- [ ] Right click option to root with branch support or node name
- [ ] adapt reroot_tree in bottle root to behave according to above option
- [ ] diplay support value/node name with non-digit character
- [ ] [optional] get rid of default '0' value when no support value given in the original tree

### DIFF
--- a/anvio/bottleroutes.py
+++ b/anvio/bottleroutes.py
@@ -1425,15 +1425,32 @@ class BottleApplication(Bottle):
     def reroot_tree(self):
         # Get the Newick tree string from the form data
         newick = request.forms.get('newick')
-        
+
         # Create an Ete3 Tree object from the Newick string, specifying format=1
-        # since you are using a different Newick format (check Ete3 documentation for details)
+        # which means that branch support values are stored as node.name and
+        # a default support value is made for each node ('1')
+        # When rooting, some branch support values have to be moved around (see issue #2043)
+        # but node.name won't! We trick the system by replace the default support value with a unique
+        # number (unique index) and we keep track of node.name + its unique number in branch_support_value dict.
+        # After rooting, we simply go through every node, look at support value and rename the node accordingly.
+        # Wonderful you say? YES. But one big problem that is not really a problem: if the user provide a tree
+        # where the node name are ACTUALLY node names and not support value. Those are not supposed to be moved after rooting.
+        # It would be very bad. Most people would use support value and not node name, but we should be ready for any
+        # situation. TO-DO: either 1) update the help page to explain the situation and NO ONE can use trees with node name
+        # or 2) when someone click on rooting in the interface, we have a popup asking if they want their node name to be
+        # considered support values (therefore moved around) or accutal node name and they should not move.
+        # Then we change the code here to root according to what the user wants.
         tree = Tree(newick, format=1)
 
-        # Assign node names to their support values for non-leaf and non-root nodes
+        branch_support_value = {1:''}
+        unique_index = 1
+
+        # for every node that is not leaf or root, associate with a unique index
         for node in tree.traverse():
             if not node.is_leaf() and not node.is_root():
-                node.support = node.name
+                unique_index += 1
+                branch_support_value[unique_index] = node.name
+                node.support = unique_index
 
         # Find the leftmost and rightmost nodes based on the provided names
         left_most = tree.search_nodes(name=request.forms.get('left_most'))[0]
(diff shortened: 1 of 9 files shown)


meren:
FANTASTIC. Thank you both, @FlorianTrigodet and @metehaansever, and thank you again, @FlorianTrigodet, for the extensive reporting.

First,

> These info are stored in the same location in the newick tree structure. You either treat it as node name OR support value. So we are good.

Thank you, STUPID DUMB FILE FORMATS OF BIONFORMATICS. It is great.

Second, the right way to show multiple support values is the following, IMO.

If you don't find integer/float in there,

- You show a checkbox that says "[ ] I have multiple support values".
- If clicked, you show an input box that says "Character that separates support values: ____" (is it a `/`? is it `;`? one can write a smart piece of code to figure it out, too, and pre-fill the input box wink wink).
- Then you parse each sets of support values in distinct arrays, calculate min/max values for each one of them separately, and then display X number of symbols or text items on the tree depending on how many support values you parsed given the separator, and have individual color pickers for each one of them.

This way people with multiple support values can have their cake and eat it too.

metehaansever:
image
We can display 2 branch support values instead of label. This changes directly effect display value range calculation process. Due to this changes, we can't calculate percentile Symbols. I also noticed that the display value range part of the code doesn't work as it should.

For better calculation, We decided to take min and max values from all list of branch support values. For example; 
| A Tool Result   | B Tool Result|
| ------------- | ------------ |
| 100  | 99    |
| 89 | 97     |
| 88    | 93    |

Min value will be 88 and max value will be 100

metehaansever:
IMG_0004

When I look at the complex tree like above, Branch support values looks like a mess and its not giving clue about value.  :cry: 
Instead of using `/` between two values, should i try to seperate them on the branch and do color coding?  @FlorianTrigodet

metehaansever:
image
I have finally managed to display two different support values with proportionally calculated sizes. However, we encounter an issue where both of them share the same X and Y coordinates. This solution works fine for the Phylogram view, but when it comes to the Circlephylogram, we need to calculate the midpoint between the left and right branches and update the X and Y coordinates accordingly. For this specific task, I am eager to tap into @ahenoch 's knowledge regarding Circlephylogram.

ahenoch:
Just a small idea to that problem, is it possible to get the x and y coordinates of the not circular version and then do a cirlce transformation? That way the blue and red dots should be placed on one vector on the circle with the start point being the middle point of the circle.
image
image

metehaansever:
@ahenoch Thanks for insights! I guess, I got your point!

metehaansever:
image
image

PR #2236, opened 2024-03-08
Update h5py installation error message

Vlasovets:
Update the error message that is raised when the `h5py` module is not found. 
The new message suggests installing the newest version of `h5py` instead of a specific version.


meren:
Thank you for this change, @Vlasovets!

meren:
Now we will also need you to send a PR for the `DEVELOPERS.yaml` file :) Instructions are here:

Issue #2238, opened 2024-03-12
[BUG] Key Error when running anvi-metabolism


lexikazen:
## Short description of the problem

I keep getting KeyError: 1766176 when trying to run anvi-estimate metabolism, and I'm not really sure what's going on. 

## anvi'o version

Anvi'o .......................................: marie (v8)
Python .......................................: 3.10.13

Profile database .............................: 38
Contigs database .............................: 21
Pan database .................................: 16
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 4
tRNA-seq database ............................: 2

```
anvi-self-test --version
```

## System info

Anvio was installed through our institution's RCC. 

## Detailed description of the issue

I'm trying to generate a kegg-metabolism-modules.txt file so I can run anvi-compute-metabolic-enrichment, but I end up with this output when I try to run anvi-estimate-metabolism:

==========================================
SLURM_JOB_ID = 2236159
SLURM_NODELIST = hm01
==========================================

WARNING
===============================================
ProfileSuperClass found a collection focus, which means it will be initialized
using only the splits in the profile database that are affiliated with the
collection METABAT and all bins it describes.

Auxiliary Data ...............................: Found: AUXILIARY-DATA.db (v. 2)
Profile Super ................................: Initialized with 119182 of 268661 splits: PROFILE_Umfleet.db (v. 38)
Contigs DB ...................................: contigs_Umfleet_anvio8.db
Profile DB ...................................: PROFILE_Umfleet.db
Collection ...................................: METABAT
Metagenome mode ..............................: True

CITATION
===============================================
Anvi'o will reconstruct metabolism for modules in the KEGG MODULE database, as described in Kanehisa and Goto et al (doi:10.1093/nar/gkr988). When you publish your findings, please do not forget to properly credit this work.

Modules database .............................: An existing database, /hpc/apps/miniconda3/4.9.2/envs/anvio-8.0.0/lib/python3.10/site-packages/anvio/data/misc/KEGG/MODULES.db, has been loaded.
Modules ......................................: 479 found
BRITE KO hierarchies .........................: 57 found
Metabolism data ..............................: KEGG only
Output file for modules mode .................: kegg-metabolism_modules.txt
Annotation sources used ......................: KOfam
Gene calls from these sources ................: 1031010 found
* Since the --add-coverage flag was provided, we are now loading the relevant
  coverage information from the provided profile database.

WARNING
===============================================
[...]

PR #2242, opened 2024-03-15
Fix metagenome mode with collection focus

ivagljiva:
This PR addresses issue #2238. Now, when users provide a collection name to `anvi-estimate-metabolism`, we only utilize gene calls that belong to the splits in the collection. And I also updated the Mode output when users provide both `--metagenome-mode` and a collection, so now it shows this:
```
Mode (what we are estimating metabolism for) .: Individual contigs within a collection in a metagenome
```

I tested it using the Infant Gut dataset and a collection including only a few splits:
```
head -n 30 additional-files/collections/merens.txt > meren_partial_collection.txt
anvi-import-collection -p PROFILE.db -C MEREN meren_partial_collection.txt -c CONTIGS.db
anvi-estimate-metabolism -c CONTIGS.db -p PROFILE.db -C MEREN --add-coverage -O test_coverage --metagenome-mode
```

It shows the following new warning when reducing the number of relevant splits to use, and it works with only 181 gene calls that belong to those splits:
```
WARNING
Since a collection name was provided, we will only work with gene calls from the
subset of 30 splits in the collection for the purposes of estimating metabolism.

Gene calls from these sources ................: 181 found
* Since the --add-coverage flag was provided, we are now loading the relevant
  coverage information from the provided profile database.

WARNING
A subset of splits (30 of 4784, to be precise) are requested to initiate gene-
level coverage stats for. No need to worry, this is just a warning in case you
are as obsessed as wanting to know everything there is to know.
```

I also tested these cases:
```
 anvi-estimate-metabolism -c CONTIGS.db -p PROFILE.db -C E_faecalis -O test_collection
anvi-estimate-metabolism -c CONTIGS.db -p PROFILE.db -C E_faecalis --add-coverage -O test_collection_w_cov
```

And I also tested it with `anvi-self-test --suite metabolism -T 6` to make sure nothing else broke from these changes (and everything was fine) :)

### DIFF
--- a/anvio/kegg.py
+++ b/anvio/kegg.py
@@ -3278,6 +3278,13 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
                                      f"{min_contig_length_in_profile_db} nts. Anvi'o hopes that this explains some things.")
                     splits_to_use = split_names_in_profile_db
 
+            # second, if we are working with a collection, we can limit the splits to use with those from the collection
+            if self.collection_name:
+                splits_to_use = ccollections.GetSplitNamesInBins(self.args).get_split_names_only()
+                self.progress.reset()
+                self.run.warning(f"Since a collection name was provided, we will only work with gene calls "
+                                 f"from the subset of {len(splits_to_use)} splits in the collection for the "
+                                 f"purposes of estimating metabolism.")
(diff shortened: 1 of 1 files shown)


meren:
🎉

PR #2271, opened 2024-05-23
Newick rerooting

metehaansever:
**DO NOT MERGE**

Scale bar works with ExportSvg.
I found a bug on ExportSvg -> contig_length and contig_count looks undefined no matter what. :/ 
I'll merge when I fix that bug.

image


metehaansever:
image
contig_length and contig_count problems solved :100:

PR #2276, opened 2024-06-03
add GenBank file check for identical locus names

treitlis:
Adds a check and throws an error if the GenBank file used as input for anvi-script-process-genbank contains sequences with the same locus name, as this will cause the sequences to be overwritten in the output fasta.


meren:
Thank you very much for this change, @treitlis. It makes a lot of sense!

PR #2278, opened 2024-06-03
Tree scale bar on Sidebar

metehaansever:
**Do Not Merge**

- [x] Items scale bar 
- [ ] Layer scale bar (Dendogram)
- [ ] Update export func with other modes
- [ ] Polish the scale bar

image

I was using same approach with Ete3 but I realize they have some problem with their calculation.
Due to that, I open a issue to ETE to understand what is wrong with that calculation: 
https://github.com/etetoolkit/ete/issues/753


metehaansever:
I'm looking for tools that produce the same scale bar results for the same newick tree.
Below, I show visualization tools that perform correct rerooting for newick trees and have a scale bar(That doesn't mean they calculated scale bar correctly).

| Tool                 | Nodes | Branches | Default behavior | Correct rerooting | Scale bar |
|----------------------|-------|----------|------------------|-------------------|-----------|
| Archaeopteryx        | ✓     | ✓        | Nodes            | ✓                 |           |
| ATV                  |       | ✓        | Branches         | ✓                 |           |
| Dendroscope          | ✓     | ✓a       | Dialoga          | ✓a                |           |
| ETE (GUI)            | ✓     | ✓        | Branches         | ✓                 |     ✓      |
| EvolView             |       | ✓        | Branches         |                   |      ✓     |
| FigTree              | ✓     | ✓        | Both             |                   |           |
| iTOL                 | ✓     | ✓        | Input dependent  | ✓                 |      ✓     |
| PhyloWidget          | ✓     |          | Nodes            |                   |           |
| TreeView             |       | ✓        | Branches         | ✓                 |           |
| T-REX                |       | ✓        | Branches         | (✓)               |           |
| APE                  | ✓     | ✓a       | Nodes            | ✓a                |           |
| BioPerl              | ✓     | ✓        | Nodes            | ✓                 |           |
| BioPython            | ✓     |          | Nodes            |                   |           |
| Dendropy             | ✓     | ✓a       | Nodes            | (✓)               |           |
| ETE (API)            | ✓     | ✓        | Branches         | ✓                 |      ✓     |
| Geneious             | ✓     | (✓)      | Nodes            |                   |           |
| MEGA                 |       | ✓        | Branches         | ✓                 |           |
| Mesquite             | ✓     | ✓        | Nodes            | ✓                 |           |
| Newick Utilities     | ✓     | ✓a       | Nodes            | ✓a                |           |
| Pycogent/scikit-bio  |       | ✓        | Branches         | ✓                 |           |

meren:
This is excellent work, Mete! Thanks for evaluating other tools, as well.

One quick question. Does this bar here will change if people zoom-in or zoom-out?

image

If not, we can think of a way to ensure its relevance. For instance, when people click Draw it woudl appear, but if they zoom in or zoom out it would disappear and the div would be replaced with "Click Draw to See the Tree Scale" or something (removing the need to adjust it in real time to different levels of zoom behavior).

metehaansever:
> This is excellent work, Mete! Thanks for evaluating other tools, as well.
> 
> One quick question. Does this bar here will change if people zoom-in or zoom-out?
> 
> image
> 
> If not, we can think of a way to ensure its relevance. For instance, when people click Draw it woudl appear, but if they zoom in or zoom out it would disappear and the div would be replaced with "Click Draw to See the Tree Scale" or something (removing the need to adjust it in real time to different levels of zoom behavior).

I didn't add any interactive behavior to avoid breaking the zoom functionality of our chart, but your idea sounds great!

PR #2306, opened 2024-07-13
prodigal to pyrodigal

meren:
This PR changes the default gene caller in anvi'o from prodigal to pyrodigal. Gene calling associated flags in `anvi-gen-contigs-database` remain the same, and `--prodigal-single-mode` continues to modify single/meta procedures in the `pyrodigal` module. I didn't change the names of these flags since pyrodigal uses prodigal code in the background.

The current PR does not implement the `--prodigal-translation-table` parameter. Because I couldn't find a way to do it properly at this time, but we can take a look at that once everything else is working and/or until someone complains (the parameter is there, but it produces 'not implemented' error, so there is no way to miss if/when one wants to use it).

The PR also adds a new parameter, `--full-gene-calling-report`, which generates a separate TAB-delimited text file with everything imaginable about gene calls (which is way more than the amount of data we store for each gene call at this point).

The prodigal driver is still in the codebase and works, although there is no way to invoke prodigal instead of pyrodigal through the CLI at the moment. If we want to keep it for any reason, we can, but I think we should remove it from the codebase.


FlorianTrigodet:
Sounds good to me. We should even switch to pyrodigal-gv. We also need to update the rest of the code base that uses "Prodigal" as default gene caller.

Ge0rges:
Seems like the pulp dependency is really an issue of snakemake that's on the way to being solved. For now we might want to keep it included, and remove it when fixed upstream in the future.

Ge0rges:
@meren what do you think about defaulting to `pyrodigal-gv` as Florian suggested? Looking into it briefly It seems that the viral models are added, so I think this would be a good default, though we could make it a flag to call `GeneFinder` or `ViralGeneFinder`.

meren:
I'm happy with that. We should make the change and wait for @FlorianTrigodet to thumbs up and merge the branch if everything is working out well.

We also should leave behind a small note (perhaps as a blog post on anvio.org) to mark this important change, to thank the developer of pyrodigal, and explain people what it means. It could be very short but I think it would be responsible to do so. Perhaps Florian would like to do that too since he pushed for it all this far? :)

FlorianTrigodet:
Latest updates:
I modified most of the codebase related to Prodigal, and updated the driver for pyrodigal-gv.

There are two flags, `--prodigal-single-mode` and `--prodigal-translation-table`, that I did not modify to something like `--pyrodigal-gv-single-mode`. If you think it is confusing to keep anything with the word 'prodigal' let me know and I'll updated the code. 

I also found this unexpected function in `utils.py`:
```python
def get_missing_programs_for_hmm_analysis():
    missing_programs = []
    for p in ['prodigal', 'hmmscan']:
        try:
            is_program_exists(p)
        except ConfigError:
            missing_programs.append(p)
    return missing_programs
```
It is only used in `anvi-run-hmms`. Any reasons why we needed to check for prodigal here? If there is no good reasons, I'll remove prodigal here. 

Last, I need to update all the contigs.db in the tests directory. I will probably just modify the dbs, unless there is a good reason to re-run `anvi-gen-contigs-database`.

Ge0rges:
I think it's more user friendly to keep the flags named the same way. The underlying engine is still prodigal after all.

Ge0rges:
@FlorianTrigodet I believe the function you removed from utils is also used on [line 15 ](https://github.com/merenlab/anvio/blob/b45d10156ddafd60cdfa0a6f543219aa2495b99b/anvio/data/hmm/__init__.py#L15)in `anvio/data/hmm/__init__.py`

meren:
I can't believe this PR is still open 😂

Ge0rges:
@FlorianTrigodet any chance this could be merged soon?

Ge0rges:
Hmm actually running with `--prodigal-single-mode` gives:

```

Traceback for debugging
================================================================================
  File "/Accounts/gkanaan/github/anvio/bin/anvi-gen-contigs-database", line 64, in <module>
    main(args)
  File "/Accounts/gkanaan/github/anvio/anvio/terminal.py", line 926, in wrapper
  File "/Accounts/gkanaan/github/anvio/bin/anvi-gen-contigs-database", line 31, in main
    a.create(args)
  File "/Accounts/gkanaan/github/anvio/anvio/dbops.py", line 4719, in create
    gene_calls_tables.call_genes_and_populate_genes_in_contigs_table()
  File "/Accounts/gkanaan/github/anvio/anvio/tables/genecalls.py", line 479, in call_genes_and_populate_genes_in_contigs_table
    gene_calls_dict, amino_acid_sequences = self.run_gene_caller(gene_caller)
  File "/Accounts/gkanaan/github/anvio/anvio/tables/genecalls.py", line 516, in run_gene_caller
    raise ConfigError(f"{e}")
================================================================================


Config Error: name 'copy' is not defined
```

meren:
that one is now fixed, @Ge0rges :)

FlorianTrigodet:
We need to merge this branch. But a few things need to be fixed before we can do that. We need a way to handle both prodigal and pyrodigal-gv contigs.db. Both for retro-compatibility and for instances where we could mix contigs.db with different gene callers, like a pangenome made from a mix of contigs.db with prodigal and pyrodigal-gv.

We need both prodigal and pyrodigal-gv to be added as acceptable default gene-caller (in constants.py?).

Here is a small dataset with two E. coli db, on with Prodigal gene calls and the other with pyrodigal-gv. We should be able to make a pangenome with them:
[E_faecalis.tar.gz](https://github.com/user-attachments/files/20758753/E_faecalis.tar.gz)

meren:
We can define a `constants.default_gene_callers` list, and then a utils function (`utils.get_default_gene_caller`) that would return the default gene caller for a given `contigs-db` by looking at the frequency of gene calls for each option present in the `constants.default_gene_callers` list. If none of the gene callers defined in the list have any hits, it would return None (or `prodigal`, IDK). Then we would replace every instance of `constants.default_gene_caller` in the codebase with `utils.get_default_gene_caller()`.

Right?

meren:
I changed all the things, run the component tests for metagenomics. It may explode later, but it will be easy to fix.

Issue #2317, opened 2024-07-30
[FEATURE REQUEST] Using external gene clusters for anvi-pan-genome


FlorianTrigodet:
## The need

Let the user use their own gene clusters for anvi-pan-genome.

## The solution

A flag for anvi-pan-genome to tell the program to not run diamond/MCL, but use a user-defined list of gene cluster.
The final pangenome would have the exact same features with AA alignment, functions, homogeneity layers, etc. 

## Beneficiaries

I would really like that feature, but I'm sure other people would too :D

meren:
FINE.

PR #2322, opened 2024-08-07
User defined gene clusters + AAI estimates

meren:
This PR addresses the feature request made in #2317 and makes it possible to start the anvi'o pangenomics workflow with user defined gene clusters.

It also adds a preliminary average amino-acid identity calculated to have more insights into what is going on in the gene clusters.


meren:
I even updated the documentation because I am a nice person.

PR #2323, opened 2024-08-07
UI custom margins

metehaansever:
The layer margin is displayed on the screen by default.


meren:
Thank you for this, @metehaansever. I tested it, and it works.

PR #2327, opened 2024-08-08
Inspect gc acc table

metehaansever:


meren:
Hey Mete, this looks great, and I like the table template you put in. But I think it would be better if the table is there at all times and not require the user to click the 'GC Accessory' button.

metehaansever:
Default open and scrollable table ready :100: 
I have also updated the design in case more attributes are coming in the future.

image

PR #2350, opened 2024-09-27
Draw KEGG pathway maps with group data

semiller10:
The main addition in this PR is the ability to use `--groups-txt` input in `anvi-draw-kegg-pathways`. With `--ko`, maps show whether groups (of genomes from different taxa, or metagenomes from different locations, for example) are predicted to have different reactions in the map. This is documented extensively in the program artifact with differences observed between and within species groups of *Enterococcus faecalis* and *faecium* genomes. The ability to handle groups lays the groundwork for the visualization of group functional enrichment scores (see [`functional-enrichment`](https://anvio.org/help/main/artifacts/functional-enrichment-txt/)).


meren:
Great additions, Sam! I love the map grids!

PR #2362, opened 2024-10-28
anvi-export-splits-taxonomy now export a table with all taxonomic levels

FlorianTrigodet:
I need to export the split level taxonomy after importing gene taxonomy from kaiju. I realized that `anvi-export-splits-taxonomy` was only reporting the genus information. And it was not even possible to ask for a different taxo level.

Funny, because the documentation for the output says that anvi'o gives all taxo levels: https://anvio.org/help/main/artifacts/splits-taxonomy-txt/

Anyway, here is my attempt to correct that. The user cannot choose the taxo level, but you get a tab-delim file with name of split and then multiple columns for the different taxonomic levels. No header in the output at the moment.

I would be very happy if someone could check that commit before I merge. 
DANKE.

### DIFF
--- a/anvio/dbops.py
+++ b/anvio/dbops.py
@@ -1471,6 +1471,10 @@ class ContigsSuperclass(object):
         taxon_names_table = contigs_db.db.get_table_as_dict(t.taxon_names_table_name)
 
         output = open(output_file_path, 'w')
+        column_list = ['split_name'] + [k for k in taxon_names_table[list(taxon_names_table.keys())[0]].keys()]
+        header = "\t".join(column_list)
+        print(header)
+        output.write(f"{header}\n")
         for split_name in self.splits_basic_info:
             if split_name in splits_taxonomy_table:
                 taxon_id = splits_taxonomy_table[split_name]['taxon_id']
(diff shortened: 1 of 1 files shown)


ivagljiva:
Didn't work for me :( 

I tested it with the Infant Gut Tutorial datapack, after importing centrifuge taxonomy [as in this section](https://merenlab.org/tutorials/infant-gut/#importing-taxonomy-for-genes).

Then I ran:
```
anvi-export-splits-taxonomy -c CONTIGS.db -o test_split_taxonomy_export.txt
```

And this is what I got:
```
  File "/Users/ivvese001/software/anvio/bin/anvi-export-splits-taxonomy", line 38, in <module>
    main(args)
  File "/Users/ivvese001/software/anvio/bin/anvi-export-splits-taxonomy", line 26, in main
    c.gen_TAB_delimited_file_for_split_taxonomies(args.output_file)
  File "/Users/ivvese001/software/anvio/anvio/dbops.py", line 1479, in gen_TAB_delimited_file_for_split_taxonomies
    output.write('{0}\t{1}\n'.format(split_name, '\t'.join(x for x in self.splits_taxonomy_dict[split_name].values())))
TypeError: sequence item 0: expected str instance, NoneType found
```

And it is happening because in some cases the taxonomy at higher levels is a `None` type object:
> taxonomy: {'t_phylum': None, 't_class': None, 't_order': None, 't_family': None, 't_genus': 'Enterococcus', 't_species': 'Enterococcus faecalis'}

I fixed it by converting the taxonomy values to their `str()` representation (which allows None values to become strings and hence be used with string functions like `join()`). At the same time, I also switched to using f-strings instead of the old `.format()` syntax, and I added a file header. 

:)

PR #2364, opened 2024-10-30
Foldseek driver + Structure base Pangenome

metehaansever:


metehaansever:
I need to update md file for anvi-setup-prostt5 and in get_gene_clusters_de_novo we can update the name of blastall_results variable, which is not cover blast, diamond and foldseek.

FlorianTrigodet:
Coupe of notes to discuss:
- I think we don't need structurepan.py. It only contains a class to download the ProstT5 weights using a foldseek command. And ProstT5 itself is, and will not be specific to structure pangenome in the future. Since the download is managed by a foldseek command line, it could be migrated to the foldseek driver. Or somewhere else if you have a better opinion.
- ProstT5 is written with a combination of upper and lowercase. This is not the most important note, but we should write it properly when we can (with the exception of anvi-setup-prostt5 because I can see that all the command line in anvi'o are lower case - with minor exceptions).

meren:
@FlorianTrigodet's note on`structurepan.py` is spot on.

I will leave some additional comments on the code directly :)

I can't wait to test it. Florian, we will have to create a minimal test case for this so it can be run by our nightly action scripts.

anvio/__init__.py
```diff
@@ -3669,7 +3669,27 @@ def TABULATE(table, header, numalign="right", max_width=0):
                      "a comma-separated list. The default stats are 'detection' and "
                      "'mean_coverage_Q2Q3'. To see a list of available stats, use this flag "
                      "and provide an absolutely ridiculous string after it (we suggest 'cattywampus', but you do you)."}
-    )
+    ),
+    'mode': (
```
meren, line 3673: Please change the new community parameter to `pan-mode` so we don't run into trouble in the long run.

anvio/__init__.py
```diff
@@ -3669,7 +3669,27 @@ def TABULATE(table, header, numalign="right", max_width=0):
...
+        ['--mode', '-M'],
+        {'default': None,
+         'metavar': 'structure',
+         'type': str,
+         'help': 'Use this flag to set mode to structure or sequence.'}
+    ),
+   'prostt5-weight-dir': (
```
meren, line 3680: Why not call it `prostt5-data-dir`?

Indentations of new entries in `__init__` do not much each other.

anvio/__init__.py
```diff
@@ -3669,7 +3669,27 @@ def TABULATE(table, header, numalign="right", max_width=0):
                      "a comma-separated list. The default stats are 'detection' and "
                      "'mean_coverage_Q2Q3'. To see a list of available stats, use this flag "
                      "and provide an absolutely ridiculous string after it (we suggest 'cattywampus', but you do you)."}
-    )
+    ),
+    'mode': (
+        ['--mode', '-M'],
+        {'default': None,
+         'metavar': 'structure',
```
meren, line 3676: `metavar` value should be `CHOICE` so it looks good in the help menu.

anvio/__init__.py
meren: please don't forget to change the help files with the new parameter name too.

anvio/panops.py
```diff
@@ -89,12 +90,18 @@ def __init__(self, args=None, run=run, progress=progress):
...
+        self.prostt5_weight_dir = A('prostt5_weight_dir')
+
         self.additional_params_for_seq_search = A('additional_params_for_seq_search')
         self.additional_params_for_seq_search_processed = False
 
         if not self.project_name:
             raise ConfigError("Please set a project name using --project-name or -n.")
 
+        if self.de_novo_compute_mode == 'structure':
+            self.skip_alignments = True
```
meren, line 103: this condition should also set `self.skip_homogeneity` to `True`.

anvio/panops.py
```diff
@@ -89,12 +90,18 @@ def __init__(self, args=None, run=run, progress=progress):
         self.enforce_hierarchical_clustering = A('enforce_hierarchical_clustering')
         self.enforce_the_analysis_of_excessive_number_of_genomes = anvio.USER_KNOWS_IT_IS_NOT_A_GOOD_IDEA
 
+        self.de_novo_compute_mode = A('mode') or 'sequence'
```
meren, line 93: somewhere in `check_params` function we should check for the use of `--pan-mode structure` together with `--user-defined-gene-clusters` parameter (and prevent it basically).

meren (commented):
Some quick suggestions

meren:
This is the error you get when you don't have foldseek:

```
Config Error: An anvi'o function needs 'foldseek' to be installed on your system, but it
              doesn't seem to appear in your path :/ If you are certain you have it on your
              system (for instance you can run it by typing 'foldseek' in your terminal
              window), you may want to send a detailed bug report. Sorry!
```

It should be more instructive and help the user how they can get the foldseek installed.

meren:
And of course this PR should be accompanied by another PR for anvio.org to update the installation instructions so the users who install anvio-dev have all the necessary tools from the get go.

meren:
When you run the following command,

```
anvi-pan-genome -g TEST-GENOMES.db -n TEST --mode structure --force-overwrite
```

Without setting up PROSTT5 first, you get the following error:

```
File/Path Error: No such file: '/Users/meren/github/anvio/anvio/data/misc/PROSTT5/weights' :/
```

Obviously this is not the most useful error. What do we expect the user to do with this information? There should be an earlier check in anvi'o style of exception-driven programming and things shouldn't even come all the way down here before we tell the user "you seem to be interested in doing this, but this other stuff is not in place yet -- please run this program first blah blah".

meren:
I haven't looked at the code yet, but this is not how we generally download things in anvi'o:

image

If you look at KEGG or COG downloads as examples you will see that we strive to report a percentage and ETA so the users are not forced to looking at a stale screen without knowing whether something is actually happening or not.

metehaansever:
> When you run the following command,
> 
> ```
> anvi-pan-genome -g TEST-GENOMES.db -n TEST --mode structure --force-overwrite
> ```
> 
> Without setting up PROSTT5 first, you get the following error:
> 
> ```
> File/Path Error: No such file: '/Users/meren/github/anvio/anvio/data/misc/PROSTT5/weights' :/
> ```
> 
> Obviously this is not the most useful error. What do we expect the user to do with this information? There should be an earlier check in anvi'o style of exception-driven programming and things shouldn't even come all the way down here before we tell the user "you seem to be interested in doing this, but this other stuff is not in place yet -- please run this program first blah blah".

There is a problem with error messages in general, yes I will overhaul all of them. :thinking:

anvio/__init__.py
metehaansever: Frankly it seems more correct to use _prostt5-weight-dir_ here. Because what we downloaded is a model's weight file and Foldseek uses it with the same name, for example:` foldseek createdb db.fasta db --prostt5-model weights`

meren: It doesn't matter though. As far as our users are concerned, it is yet another data they are downloading. Regardless what individual data items mean to individual processes, we can have a common vocabulary we bring in from the outside world. What we download from COGs is FASTA files. What we download from KEGG is models. We call all of them data on anvi'o side.

FlorianTrigodet:
We will need 2 additional layers in the structure based pan, with information taken from the AA pangenome. 
One layer will be "AA-GC Count" for the number of amino acid GCs that ended up in a structure GC.
The second one would be a barplot (stacked bars) called "AA-GC Type" with three information: "core", "singleton", "else". You would count how many genes in the structure GC came from a AA core or singleton GC. If non of them, then you count it as "else".

meren:
There is something annoying going on here and we need to fix it soon. As you can see here, SCGs (the red layer) are not coming together nicely anymore.

As the flat branches in the dendrogram like the one I marked below has an alphabetical order, when we named our gene clusters based on their frequency across genomes in the regular pangenomics workflow, this would enforce SCGs to be next to each other nicely in the core. But it seems it is not happening anymore because the way we name them does not care about their frequencies across genomes :/ I'm putting it here as a note so we take care of it at some point.

image

metehaansever:
We should add pan-db-type variant as 'structure' to get structural specific data table in interactive class when pan-mode == 'structure'.

meren:
@metehaansever, https://github.com/merenlab/anvio/pull/2364/commits/3084b8f41d2eb69b3c6d7911d1f21a921173b9ed should fix the bug above. Please test :)

PR #2376, opened 2024-11-26
A global argument to keep all functional annotations per gene (fixes #2375)

### LINKED ISSUE #2375: [BUG] anvi-summarize doesn't report all KOfam annotations when there are multiple for the same gene call

Sometimes gene calls get annotated with multiple KOfams, but when you run `anvi-summarize` only one of these KOfams will be included in the output summary.

## anvi'o version

anvi'o-dev 

@tucker4 first noticed this issue in the context of pangenome summaries, and I've just confirmed it for regular profile db summaries. Basically, when a gene has multiple KOfam annotations, each one is recorded in a separate row of the `gene_functions` table of the contigs database:

```
$ for gid in 612 17396; do sqlite3 CONTIGS.db "select * from gene_functions where gene_callers_id='$gid' and source='KOfam'"; done
612|KOfam|K00111|glycerol-3-phosphate dehydrogenase [EC:1.1.5.3]|1.5e-147
612|KOfam|K00105|alpha-glycerophosphate oxidase [EC:1.1.3.21]|0
17396|KOfam|K02275|cytochrome c oxidase subunit II [EC:7.1.1.9]|4.8e-62
17396|KOfam|K02261|cytochrome c oxidase subunit 2|1.6e-107
```

When we run `anvi-summarize`, only one of the functions is incorporated into the summary `*gene_calls.txt` file for the corresponding bin. For instance, for gene call 612 we only see K00105 in the output:
```
$ for gid in 612 17396; do grep -e "^$gid" SUMMARY/bin_by_bin/*/*gene_calls.txt; done
SUMMARY/bin_by_bin/E_facealis/E_facealis-gene_calls.txt:612	Day17a_QCcontig1	652697	654527	f	Glycerol-3-phosphate dehydrogenase (GlpA) (PDB:2RGH)	COG0578	Glycerol-3-phosphate dehydrogenase	COG0578	C	C	KEGG Orthology (KO)>>>09100 Metabolism>>>09103 Lipid metabolism>>>00564 Glycerophospholipid metabolism [PATH:ko00564]!!!Enzymes>>>1. Oxidoreductases>>>1.1  Acting on the CH-OH group of donors>>>1.1.3  With oxygen as acceptor>>>1.1.3.21  glycerol-3-phosphate oxidase	ko00001!!!ko01000					Isoprenoid biosynthesis	COG0578	Energy production and conversion	C			alpha-glycerophosphate oxidase [EC:1.1.3.21]	K00105
```

(in this case, gene call 17396 is not within one of the bins of the collection I was testing)

This is happening because the method that loads functions from the contigs database (`init_functions`) is keeping only the hit with lowest e-value when there are multiple per gene. This is happening [here in the codebase](https://github.com/merenlab/anvio/blob/621288cda3afdc3506e82c85f577d248f8f13278/anvio/dbops.py#L646C1-L653C75), but I copy the relevant lines of code below for convenience:

```
for hit in hits:
            gene_callers_id = hit['gene_callers_id']
            source = hit['source']
            accession = hit['accession']
            function = hit['function']
            e_value = hit['e_value']

            if gene_callers_id not in self.gene_function_calls_dict:
                self.gene_function_calls_dict[gene_callers_id] = dict([(s, None) for s in self.gene_function_call_sources])

            if self.gene_function_calls_dict[gene_callers_id][source] and e_value:
                if self.gene_function_calls_dict[gene_callers_id][source][2] < e_value:
                    # 'what we have:', self.gene_function_calls_dict[gene_callers_id][source]
                    # 'rejected    :', ('%s :: %s' % (function if function else 'unknown', accession), e_value)
                    continue

            entry = (accession, '%s' % (function if function else 'unknown'), e_value)
            self.gene_function_calls_dict[gene_callers_id][source] = entry
```
basically: for each annotation (row from the contigs db functions table), if we haven't seen the gene caller ID before we initialize an empty dictionary of annotation sources for it. If we HAVE seen the gene caller id before and if the current annotation has a higher (worse) e-value than the previous entry for this gene call, we skip the current annotation. Otherwise, we add the current annotation into the dictionary, which effectively replaces any annotations that we saw before. Therefore, **only the last KOfam annotation of a given gene makes it into the `anvi-summarize` output.**

This only affects KOfam, as far as I can tell, because other annotation sources handle the multiple annotation case by putting them all in the same row of the gene_functions table, separated by the `!!!` delimiter. And clearly it is a bit misleading and confusing for users since they aren't seeing all relevant KOfams in the output.

## Files / commands to reproduce the issue

I was using the Infant Gut Tutorial datapack for the examples above.

## Do we need to fix it? And how to fix it

One consideration is that we are effectively keeping the 'best KOfam hit' per gene using the current strategy. Some people might like that. However, this behavior is different from how we handle all other annotation sources and is not documented anywhere, which can lead to confusion in some cases. So one opinion I would like to gather (from experts like @tucker4 and @meren ), is: _is this an issue and should we therefore fix it?_  I am leaning towards 'yes' at the moment but would like to hear what others think.

If we fix it, I see two strategies to make sure that multiple KOfam annotations all make it into the `anvi-summarize` output:

1. update `init_functions()` in `dbops.py` to include all annotations for a given gene call (delimited by !!!) regardless of e-value
2. update `anvi-run-kegg-kofams` to store multiple KOfam annotations for a given gene in a single row (delimited by !!!) so that we don't ever get to this issue of replacement in `init_functions()`

A potential disadvantage of (1) is that I believe some downstream functions might rely on only getting 'one best hit' per gene call (perhaps the functional enrichment programs), and if we remove that, it might be problematic. Therefore, (2) might be a better solution, as it would likely only have a downstream impact on `anvi-estimate-metabolism` and `anvi-reaction-network` (tagging @semiller10 to see if that is the case).

ivagljiva:
This PR fixes #2375 with the addition of a global argument called `RETURN_ALL_FUNCTIONS_FROM_SOURCE_FOR_EACH_GENE`. When this argument is set to True (which currently only happens when running `anvi-summarize` on a profile database), the `dbops.py` function `init_functions` can return multiple functional annotations (from a given annotation source) per gene separated by the `!!!` delimiter.

For instance, BEFORE this fix, `anvi-summarize` output was showing only one KOfam for gene 612 in the Infant Gut Dataset:
```
612	Day17a_QCcontig1	652697	654527	f	K00105
```
But now it is showing both KOfams:
```
612	Day17a_QCcontig1	652697	654527	f	K00111!!!K00105
```

A couple of caveats:

1. I didn't make this global argument into a flag; that is, a user cannot add `--return-all-functions-from-source-for-each-gene` to their command line in order to set this argument to True. I didn't see the need for this, since it's used in just one case so far. But perhaps in the future, if we need it elsewhere, we could change this to enable users to set the flag as well.
2. e-values are also combined with the `!!!` delimiter when this flag is used in `init_functions` to set the contents of the `self.gene_function_calls_dict`. This means that any downstream code expecting the e-value to be a float will break. So far, it is not a big deal because we only use it in `anvi-summarize` (which works fine), but if we start using this flag elsewhere, it could matter.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -74,6 +74,9 @@ DEBUG_AUTO_FILL_ANVIO_DBS = '--debug-auto-fill-anvio-dbs' in sys.argv
 USER_KNOWS_IT_IS_NOT_A_GOOD_IDEA = '--I-know-this-is-not-a-good-idea' in sys.argv
 DOCS_PATH = os.path.join(os.path.dirname(__file__), 'docs')
 TMP_DIR = None
+# global args that we can set internally as needed
+RETURN_ALL_FUNCTIONS_FROM_SOURCE_FOR_EACH_GENE = False  # set to True if you want all functional annotations from a given annotation source
+                                                        # instead of the best hit per gene
 
 # if the user wants to use a non-default tmp directory, we set it here
 if '--tmp-dir' in sys.argv:
--- a/anvio/dbops.py
+++ b/anvio/dbops.py
@@ -602,6 +602,9 @@ class ContigsSuperclass(object):
         If self.split_names_of_interest has a value, the dictionary only includes gene calls from those splits.
 
         Afterwards, it sets self.gene_function_calls_initiated to True.
+
+        Note: the global argument RETURN_ALL_FUNCTIONS_FROM_SOURCE_FOR_EACH_GENE affects the behavior of this function. If False, we get 
+        the best hit per gene (lowest e-value) for a given annotation source. If True, we get all hits.
         """
         if not self.contigs_db_path:
             return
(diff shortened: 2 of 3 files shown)


meren:
Hey @ivagljiva, I see what you did here and it makes sense for the time being. As you said, someone can revisit this if it becomes necessary to have a flag for this.

Thank you very much for fixing this! :)

PR #2394, opened 2025-01-31
Report all completion and redundancy estimate

FlorianTrigodet:
I added a flag to `anvi-estimate-genome-completeness` to report completion/redundancy estimate from all domains. The function was already in the code but not available when using the classic anvi'o SCG collections.


meren:
Gorgeous :)

PR #2396, opened 2025-02-18
Update legend color sync

metehaansever:
Fix color synchronization between legends and visualization

- Initialize color pickers with actual colors from categorical_data_colors instead of default white
- Add proper color synchronization in colpick onChange handler to update data structures
- Add immediate redraw trigger after color changes to keep visualization in sync
- Store category information in color picker elements for accurate data mapping

This fixes the issue where colors in the legend panel didn't match the colors in the visualization.


meren:
❤️

PR #2414, opened 2025-03-20
Better locked database handling

meren:
*His palms are sweaty, knees weak, arms are heavy*

When we are running a large number of anvi'o programs on a large number of databases all in parallel, sometimes we get errors like this from our snakemake workflows:

```
Num relevant SCGs in contigs db ..............: 17,568
Taxonomy .....................................: /fs/dss/groups/agecodatasci/SOFTWARE/anvio/anvio/data/misc/SCG_TAXONOMY/GTDB/ACCESSION_TO_TAXONOMY.txt.gz
Database reference ...........................: /fs/dss/groups/agecodatasci/SOFTWARE/anvio/anvio/data/misc/SCG_TAXONOMY/GTDB/SCG_SEARCH_DATABASES
Number of SCGs ...............................: 22

Parameters for search
Max number of target sequences ...............: 20
Max e-value to report alignments .............: 1e-05
Min percent identity to report alignments ....: 90.0
Num aligment tasks running in parallel .......: 1
Num CPUs per aligment task ...................: 40
Log file path ................................: /fs/dss/scratch/6934191_tepp5511_mpcs045/tmp5xmvoev0

✖ anvi-run-scg-taxonomy encountered an error after 0:00:49.854952
  File "/fs/dss/groups/agecodatasci/SOFTWARE/anvio/bin/anvi-run-scg-taxonomy", line 73, in <module>
    main(args)
  File "/fs/dss/groups/agecodatasci/SOFTWARE/anvio/anvio/terminal.py", line 926, in wrapper
  File "/fs/dss/groups/agecodatasci/SOFTWARE/anvio/bin/anvi-run-scg-taxonomy", line 29, in main
    t.populate_contigs_database()
  File "/fs/dss/groups/agecodatasci/SOFTWARE/anvio/anvio/taxonomyops/__init__.py", line 1203, in populate_contigs_database
    self.tables_for_taxonomy.delete_contents_of_table(anvio_taxonomy_table_name, warning=False)
  File "/fs/dss/groups/agecodatasci/SOFTWARE/anvio/anvio/tables/tableops.py", line 119, in delete_contents_of_table
    database._exec('''DELETE FROM %s''' % (table_name))
  File "/fs/dss/groups/agecodatasci/SOFTWARE/anvio/anvio/db.py", line 525, in _exec
    self.commit()
  File "/fs/dss/groups/agecodatasci/SOFTWARE/anvio/anvio/db.py", line 470, in commit
    self.conn.commit()
sqlite3.OperationalError: database is locked
```

This is happening, because another program (like `anvi-run-kegg-kofams` is trying to write to the database while `anvi-run-scg-taxonomy` is also trying to the same database. Previously we had solved this problem with the `check_if_db_writable` function in `db.py`. But it is only checking the presence of a journal file when the DB is FIRST INITIALIZED, which is not a good solution.

This PR is introducing an entirely different way to solve this at the level of atomic operations.

I tested it using a Python program to lock anvi'o databases, `lock.py` (lol, creativity star here):

``` python
import sys
import sqlite3
import time

while True:
    conn = sqlite3.connect(sys.argv[1])
    cursor = conn.cursor()

    # create a dumb table
    cursor.execute("CREATE TABLE IF NOT EXISTS a_test_table_that_may_or_may_not_be_there (id INTEGER PRIMARY KEY, value TEXT)")

    # Start a transaction and lock the database
    cursor.execute("INSERT INTO a_test_table_that_may_or_may_not_be_there (value) VALUES ('Locked Entry')")
    print(f"Database is locked until you press CTRL+C")
    time.sleep(42 * 42)

    # lol
    conn.commit()
    conn.close()
```

See it in action.

locked_db

I tested it with mini-test, and it seems working, but more eyes are needed on this before putting it in `master` 😬

### DIFF
--- a/anvio/db.py
+++ b/anvio/db.py
@@ -459,7 +459,11 @@ class DB:
 
 
     def execute_safely(self, func, *args):
-        """Execute an SQLite instruction safely and by handlng database locked errors gracefully"""
+        """Execute an SQLite instruction safely and by handlng database locked errors gracefully
+
+        See https://github.com/merenlab/anvio/pull/2414 for details and example code to test this
+        functionality.
+        """
 
         retries = 0
         while retries <= self.max_retries:


semiller10 (commented):
This strategy looks good! I assume that the "database is locked" error is most likely to occur rather than an error like "unable to open database file" due to another node locking it while it is being opened.

I have one logical quibble, which is that `while retries < self.max_retries` should be changed to `while retries <= self.max_retries` in `execute_safely`. For example, if `max_retries` equals 1, then a single retry (two total tries) would be expected, but as it is currently written, no retries (one total try) is allowed.

meren:
Great point, Sam :) There should also be a sanity check over the self.max_retries variable since it shouldn't be ever set to something < 1.

PR #2419, opened 2025-04-07
Anvi'o can has COG24?

meren:
This PR adds the latest COGs distribution from the NCBI described in https://doi.org/10.1093/nar/gkae983.

As it kind of doubles the number of genes in the database (that resolve only to a few new COGs (so don't get your hopes up)), this update will increase the runtime of `anvi-run-ncbi-cogs` quite a bit, but not too bad.


meren:
Please note that when we merge this PR, every anvi'o installation that tracks the main branch (including those on our HPCs and elsewhere) will give the following error when someone runs `anvi-run-ncbi-cogs` as a part of their new or ongoing workflows:

image

One solution is to keep the default version as `COG20` the other is to go full anarchy. I lean towards the latter. LMK WHAT U THINK PLS.

meren:
There is one last thing I want to mention. I learned that the FASTA file (COGorg24.faa.gz) contains sequences of all 7,698,004 proteins, but only 6,027,885 of those sequences are currently assigned to a COG. Which means there are about 1.7 million sequences that we include in our search databases, but we do nothing with the resulting hits since they are not associated with a COG ID (yet).

A good practice during `anvi-setup-ncbi-cogs` would be to get all the protein IDs associated with a COG, and remove those 1.7 million sequences from the FASTA file after download. This would remove about 20% of the data that end up in the search database with no impact on the output of `anvi-run-ncbi-cogs` and avoid all that additional computation.

This is not very elegant and I don't want to tinker with data from upstream, but then the situation is not ideal, which makes it a good solution. What are your thoughts on this?

ivagljiva:
Yay! Nice work turning this around so quickly, Meren :) 

> One solution is to keep the default version as COG20 the other is to go full anarchy. I lean towards the latter.

I prefer full anarchy as well. The error message is clear, and we can deal with the consequences :p   I would suggest we put an announcement on the Discord once this is merged (and one of us can quickly go and setup COG24 on our HPC to avoid workflow issues for ourselves)

> that there are some functions COG20 could annotate but COG24 is missing

weird. but I guess they could have removed some sequences during the update, so maybe not too weird.

>  A good practice during anvi-setup-ncbi-cogs would be to get all the protein IDs associated with a COG, and remove those 1.7 million sequences from the FASTA file after download. This would remove about 20% of the data that end up in the search database with no impact on the output of anvi-run-ncbi-cogs and avoid all that additional computation.

I'm in favor of this approach. I think there is no point running the extra comparisons, and as long as we are clear in the terminal output and documentation about what we are doing, I think we can safely remove the sequences. There are two options I can see: keep the no-COG sequences around in the data dir just in case (with perhaps a flag to include them in the comparisons in case someone wants to take a look at the temp output files and wants to see where these sequences match?) similar to the `orphan_data` in the KEGG data dir, OR delete them entirely to save some space on everyone's computers. I guess I would lean towards the former, but if the storage space is significantly reduced by deleting those no-COG sequences, perhaps we could make yet another flag so that the people can decide for themselves if they want to keep everything or save space :)  Sorry to suggest things that require more work to implement :p (happy to help if needed!)

ivagljiva (approved):
looks good to me so far @meren  :) I am testing it at the moment on my computer but I don't anticipate any issues (will report back if I find something)

ivagljiva:
Well, I tested the annotation w/ COG24 on a genome (B adolescentis), and it was blindingly fast :) Only took 25 seconds (with 6 threads). So maybe the performance gains from removing those 1.7 million sequences won't be significant (at least on this scale, with large metagenomes it could still make a difference). 

The number of genes matching to no-COG protein IDs was rather large (210, which is like 1/6 of the number of gene calls that got annotations), and the resulting warning listing all the relevant protein IDs was quite extensive. Since there are a lot of these, I am even more convinced that it would be good for people to have the option to include these mysterious protein IDs in their analysis; hence, I am even more in favor of the following:

> keep the no-COG sequences around in the data dir just in case (with perhaps a flag to include them in the comparisons in case someone wants to take a look at the temp output files and wants to see where these sequences match?)

One funny thing I noticed: the citation statements are present multiple times when you run `anvi-run-ncbi-cogs`:
```
CITATION
Anvi'o will set up the COG24 version of NCBI COGs. We recommend citing the
following reference for this database when you publish your findings : Galperin
et al. 2025 (https://doi.org/10.1093/nar/gkae983)


CITATION
Anvi'o will annotate your database with the COG24 version of NCBI COGs. We
recommend citing the following reference for this database when you publish your
findings : Galperin et al. 2025 (https://doi.org/10.1093/nar/gkae983)

[...]
DIAMOND VIEW
Diamond  tabular output file .................: /var/folders/nc/7dlw5z2j16q3s14586qddwl8nhxpgh/T/tmp8dqqx5rh/diamond-search-results.txt
COG version ..................................: COG24
COG data source ..............................: The command line parameter.
COG base directory ...........................: /Users/ivvese001/Lab/test-kegg/COG24

CITATION
Anvi'o will set up the COG24 version of NCBI COGs. We recommend citing the
following reference for this database when you publish your findings : Galperin
et al. 2025 (https://doi.org/10.1093/nar/gkae983)
```

There is the 'annotate' citation statement happening 1x, and the 'set up' citation statement happening 2x because it is printed each time we initialize an instance of `COGsSetup`. Since we have different language depending on what program is run, I think we should keep both versions of the statement in the code base, but turn off the `set up` version when we initialize COGsSetup from the `COGs` class. We could do that by adding a `quiet` flag to the COGsSetup arguments. 

But the fact that the 'set up' statement occurs 2x means that we are initializing twice (once after Diamond is already done), and the second initialization might not be necessary. I think the second one is happening because we initialize `COGsData` which in turn initializes `COGsSetup` again. Not sure how best to resolve these dependencies, but I suppose an easy solution is to keep the classes as they are and just add a `quiet` flag to suppress the COGsSetup citation statement.

PR #2423, opened 2025-04-14
Load hits into memory for quicker processing

bcoltman:
Loading the hits file into memory and processing it all at once led to a speed boost (not comprehensively tested) for large hit files. Hit files are unlikely to be that large and therefore I decided I'd open a pull request

### DIFF
--- a/anvio/drivers/hmmer.py
+++ b/anvio/drivers/hmmer.py
@@ -312,7 +312,7 @@ class HMMer:
                         append_function = self.append_to_main_standard_file
                     elif file_type == 'domtable':
                         append_function = self.append_to_main_table_file
-
+                    
                     append_function(main_file_buffer, worker_file, main_file_lock)
 
             except KeyboardInterrupt:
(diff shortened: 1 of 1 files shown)


meren:
Hi again @bcoltman! Thank you for this PR! I have two minor questions:

* I understand that you haven't done a comprehensive benchmarking I am still wondering what kind of a speed gain we are talking about here. Would you care to comment on that?
* Since this is such a central function to almost every workflow in anvi'o, can you please tell us how did you test whether it is working seamlessly?

Thank you so much!

bcoltman:
Hi @meren
You're right, I'll collect some benchmark/test data and give you an update!

bcoltman:
Hi @meren, 
Before I start, I wanted to provide a little bit of background as to why I edited the function as it seems a little random. I was looking at using the existing anvio codebase and adapting it to a testing pipeline I had developed. My pipeline took raw short-reads, merged mates if possible and then called ORFs and annotated them. A functional profiler for reads if you will. Obviously this led to a large a number of hits being generated and the hits file being quite large. I profiled my prototype and was surprised that ```append_to_main_table_file``` appeared took a while, so decided to update it. That is not to say that it was slow (I can't find the result now, it was a short while back...), but I thought I could optimise it by processing it all at once. 

Anyway, I've managed to assess the robustness and performance pre and post-update.

Robustness: produces the same output, tested using 4 threads. 
Performance: small improvement

## Robustness
I twice ran a complete genome database annotation: 
```bash
anvi-gen-contigs-database
anvi-run-hmms
anvi-run-kegg-kofams
anvi-run-pfams
anvi-run-cazymes
anvi-run-ncbi-cogs
```

One time with the original ```append_to_main_table_file``` function, and a second time with the edit included in my pull request.

I compared the two databases with ```anvi-display-contigs-stats``` and found no differences in [their statistics](https://github.com/user-attachments/files/19759702/Contigs.DB.Stats.pdf)

I also ran ```diff db1 db2``` on the outputs of both ```anvi-get-sequences-for-gene-calls``` and ```anvi-export-functions``` for all annotation sources (CAZyme, COG20_CATEGORY, COG20_FUNCTION, COG20_PATHWAY, KEGG_BRITE, KEGG_Class, KOfam, Pfam, Transfer_RNAs) with the only difference being due to sort ordering of the CAZyme ```anvi-export-functions```
 
 ## Performance

I cProfiled the functions for differences. I captured ```output_dict``` from the function and thus, this performance comparison only considers a single worker. I used both the original output_dict (172 entries), but also created alternative dictionaries with additional entries (I repeated 3 lines 1e3, 1e4, 1e6 and 1e7 times). THus, my largest file was ~3e7 entries and around 4GB.

Results below:
| File Size (Rows) | Old version Total Time (s) | New version Total Time (s) |
|------------------|--------------------|--------------------|
| 172         | 0.015              | 0.013              |
| + 3*1e3            | 0.019              | 0.036              |
| + 3*1e4            | 0.261              | 0.102              |
| + 3*1e6            | 16.015             | 6.444              |
| + 3*1e7            | 155.000            | 59.000             |



Let me know if you want any additional comparisons. I agree that it is a rather central function so I understand the hesitancy.

meren:
Hi @bcoltman, thank you so much for these insights and comparisions :) My only fear is the following: we often run HMMs on contigs-db files that describe metagenomic assemblies rather than individual genomes where the number of hits sometimes reaches to millions. But I guess even in that instance reading the entire file into memory should not be a significant issue. If that becomes a problem we can revisit this, but apart form that we shall get any improvement we can happily :)

I will merge this now so it gets tested in every running `anvio-dev` instance (usually it is a great way to learn if there are edge cases we missed in PRs :)).

Thank you very much for looking into this and taking the time to respond to questions here. I look forward to interacting more.

bcoltman:
Thanks @meren! Sounds good to me and I'm happy I could help. Cheers!

PR #2424, opened 2025-04-14
Updating get-sequences-for-gene-calls for genes with multiple annotat…

bcoltman:
When exploring function calls for some databases I recognised that the output of anvi-export-functions:
e.g. For gene_id 1401
```
1401    KOfam   K24866  precorrin-2 dehydrogenase [EC:1.3.1.76] 1.5e-71
1401    KOfam   K02302  uroporphyrin-III C-methyltransferase / precorrin-2 dehydrogenase / sirohydrochlorin ferrochelatase [EC:2.1.1.107 1.3.1.76 4.99.1.4]     1.4e-197
1401    KOfam   K02303  uroporphyrin-III C-methyltransferase [EC:2.1.1.107]     5.1e-120
```

didn't match up with what I saw from anvi-get-sequences-for-gene-calls.

```md
GCF_019670005_000000000001      .       CDS     1557585 1559012 .       -       .       ID=GCF_019670005___1401;Name=K02302;db_xref=KOfam:K02302;product=uroporphyrin-III C-methyltransferase / precorrin-2 dehydrogenase / sirohydrochlorin ferrochelatase [EC 2.1.1.107 1.3.1.76 4.99.1.4]
```

This behaviour was intended and would pick the function hit with the lowest e-value. I thought it might be valuable for anvio users (but mainly for me) to export multiple annotations for each gene call in their gffs (Yay...) so looked at how I could implement it. 

I found that the behaviour was dependent on the global attribute ```RETURN_ALL_FUNCTIONS_FROM_SOURCE_FOR_EACH_GENE``` and that by setting it to TRUE, I could achieve the desired outcome. I made this dependent by including an additional flag ```--return-all-hits```. 

The other code is for formatting multiple strings. The script will now look at all gene calls and if multiple features are present then the strings are formatted, comma-separated and populate the relevant fields. According to [GFF3 specification](https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md)(the GFF3 specification), comma-separation is okay for db_xref and product, but not really for Name. If merged, you may choose to retain the !!! separation for Name.

anvi-get-sequences-for-gene-calls can now be called as:
```bash
anvi-get-sequences-for-gene-calls --annotation-source KOfam --export-gff3 -c results/bacdive/bacteria_anvio/database/GCF_019670005.db -o GCF_019670005_new.gff3 --return-all-hits
```

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -1269,6 +1269,14 @@ D = {
                      "the gene names that appear multiple times, and remove all but the one with the lowest e-value. Good "
                      "for whenever you really need to get only a single copy of single-copy core genes from a genome bin."}
                 ),
+    'return-all-function-hits-for-each-gene': (
+            ['--return-all-function-hits-for-each-gene'],
+            {'default': False,
+             'action': 'store_true',
+             'help': "Any given function annotation source may provide more than one annotation for a given gene. Using this flag "
+                     "will instruct anvi'o to report all hits, even if the default behavior is to report only the most "
+                     " statistically significant one."}
+                ),
     'unique-genes': (
             ['--unique-genes'],
             {'default': False,
--- a/anvio/dbops.py
+++ b/anvio/dbops.py
@@ -1448,16 +1448,26 @@ class ContigsSuperclass(object):
                 attributes = f"ID={entry_id}"
                 if gene_functions_found:
                     if gene_callers_id in self.gene_function_calls_dict:
-                        accession, function, evalue = self.gene_function_calls_dict[gene_callers_id][gene_annotation_source]
-                        accession, function = accession.split('!!!')[0], function.split('!!!')[0]
+                        accession_str, function_str, evalue = self.gene_function_calls_dict[gene_callers_id][gene_annotation_source]
+                        accessions = accession_str.split('!!!')
+                        functions = function_str.split('!!!')
 
-                        # now we have the function text, but it may contain characters that are offensive to the GFF3
-                        # specifications, so we will simply replace them with space character. for more information
-                        # please see https://github.com/merenlab/anvio/issues/2070
-                        for character in [':', ';', ',', '&', '\t']:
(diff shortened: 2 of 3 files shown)


anvio/__init__.py
```diff
@@ -1269,6 +1269,14 @@ def TABULATE(table, header, numalign="right", max_width=0):
                      "the gene names that appear multiple times, and remove all but the one with the lowest e-value. Good "
                      "for whenever you really need to get only a single copy of single-copy core genes from a genome bin."}
                 ),
+    'return-all-hits': (
+            ['--return-all-hits'],
+            {'default': False,
+             'action': 'store_true',
+             'help': "A bin (or genome) may contain more than one hit for a gene name in a given HMM source. For instance, there may "
```
meren, line 1276: Hi @bcoltman, thank you very much for these changes. They generally look good to me, but I think the help menu could be a little more generalized since function annotation sources are not specific to HMMs and flags in `__init__.py` can be used in multiple different programs :)

> Any given function annotation source may provide more than one annotation for a given gene. Using this flag will instruct anvi'o to report all hits, even if the default behavior is to report only the most statistically significant one.

My second suggestion is the following:

Perhaps we can use `--return-all-function-hits-for-each-gene` instead of `--return-all-hits`. I know it is longer an clunkier, but it would better connect with the global flag `RETURN_ALL_FUNCTIONS_FROM_SOURCE_FOR_EACH_GENE` and it would be easier for others to recognize in the future that there is this community flag that they can use for similar tasks. Return all hits is a bit cryptic in that sense.

meren:
Hi @bcoltman! Thanks for sending this in! It is almost ready to go in, but I just have one suggestion as you will see above :)

bcoltman:
Thanks for the solid suggestions @meren 

I've incorporated them into the fork and squashed the commits

meren:
Thank you, @bcoltman! I'm merging this :)

PR #2437, opened 2025-05-11
Update anvi-gen-variability-profile.md

mankeldy:
Updated info for the --output-file, which is can be necessary when scripting or working with multiple engine parameters (otherwise, I imagine the file would get overwritten or anvio wouldn't like that it already exists. I ran a script without this and I have no idea where the file went, so hopefully someone doesn't make the same mistake :) )


meren:
Thank you for this update!

PR #2451, opened 2025-07-06
Adding the necessary functionality to EcoPhylo to work with multiple HMMs

meren:
@FlorianTrigodet is running the show, but I wanted to open a PR so we have a way to tag issues.


FlorianTrigodet:
Needs:
- [ ] Documentation
- [ ] Update ecophylo tutorial and/or reproducible workflow by Matt

Can use:
- [ ] better implementation of the metagenomics workflow
- [ ] at the moment, it is unable to combine hmm and run scg-taxonomy. Should be solvable though.
- [ ] explicit output directories, regardless of tree/profile mode (e.g. final contigs.db and profile.db in a single final directory, regardless of the mode)

PR #2454, opened 2025-07-14
Multithreading anvi-script-find-misassemblies

FlorianTrigodet:
I tested it on a test case (single contig = single threads) and one real world dataset (metaMDBG assembly of Zymo mock community) and it is working with no difference in the output, except for the conitg order but we can live with that. 

Currently testing on ALL samples from ["Assemblies of long-read metagenomes suffer from diverse errors"](https://doi.org/10.1101/2025.04.22.649783).

Will merge when I confirm that ALL outputs are identical to the previous version of the script. 

Note: I don't have a special case for when the number of threads is one. I can see that in the rest of the codebase, we treat single thread processes differently. Notably the progress bar is quite different. But are there any other reasons I should be aware or is that fine for this code?

The PR include an update to the documentation as well.

### DIFF
--- a/anvio/docs/programs/anvi-script-find-misassemblies.md
+++ b/anvio/docs/programs/anvi-script-find-misassemblies.md
@@ -56,3 +56,9 @@ Another default behavior of the script is to skip the first and last 100bp of a 
 {{ codestart }}
 anvi-script-find-misassemblies -b sample01.bam -o result --min-dist-to-end 0
 {{ codestop }}
+
+You can also speed up the process by using multiple threads with the flag `-T`:
+
+{{ codestart }}
+anvi-script-find-misassemblies -b sample01.bam -o result -T 8
+{{ codestart }}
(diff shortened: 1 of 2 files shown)


FlorianTrigodet:
The results are strictly identical. So I am merging. 🎉

meren:
Excellent, job, @FlorianTrigodet. Can you say about the extent of reduction in runtime as a function of increasing number of threads?

FlorianTrigodet:
I tested it on a few assemblies:


Threads | HumanO1 | SheepA | AD2W1 | HADS_013 | Zymo_HiFi
-- | -- | -- | -- | -- | --
1 | 56m41.634s | 126m10.316s | 88m49.528s | 109m23.836s | 18m2.091s
2 | 29m43.119s | 67m58.245s | 50m8.724s | 75m43.273s | 9m32.865s
4 | 25m16.220s | 49m34.889s | 31m17.395s | 61m57.717s | 5m54.634s
8 | 23m45.001s | 47m7.056s | 29m32.674s | 60m10.297s | 4m28.061s
16 | 23m34.741s | 46m59.205s | 29m12.306s | 60m24.931s | 3m54.246s
32 | 23m34.729s | 45m14.510s | 28m27.484s | 59m57.612s | 3m45.396s


After profiling the code, it is I/O limited with pysam reading the bam file.

PR #2458, opened 2025-07-29
Modernized pypackaging for anvi'o

meren:
This PR migrates all 160+ anvi'o programs to use Python entry points and modernizes the anvi'o packaging system to catch up with the current Python best practices.

We currently keep anvi'o programs under `bin/` and `sandbox/` directories in the codebase. That means, when the user sets up an anvi'o conda environment to track the active codebase (i.e., `anvio-dev`), the conda activation script must set PATH variable to point to these directories (and a PYTHONPATH variable to point to `anvio/anvio`) so they can execute anvi'o programs directly from the copy of the anvi'o code from GitHub.

This strategy worked perfectly fine for development, but it conflicted with modern Python packaging standards and created issues when anvi'o needed to be installed in non-development environments or distributed through standard Python packaging channels.

So this week I decided that it was time to bite the bullet, and have MANUALLY migrated all 160+ anvi'o programs to use so-called 'entry points' -- a modern Python mechanism where programs are installed as proper commands. It is true that EVERY single change you see (**364 changed files with 18,191 additions and 17,389 deletions**) happened in my trusty vi and took about 16 hours in total :) See the new [pyproject.toml](https://github.com/merenlab/anvio/blob/2fa6fe59cf360037a0612a87affff6963d55542a/pyproject.toml) file that replaces our `setup.py`.

As far as I am concerned, THE PR IS READY TO MERGE (along with the branch `modernized-pypackaging` that is waiting in our web repo at https://github.com/merenlab/anvio.org).

Yours truly,
Meren


FlorianTrigodet:
Tested and working. Let's merge and see 🤞

Issue #2148, opened 2023-10-15
[BUG] Most annotations from anvi-run-cazymes have undefined ('-') accession numbers


ivagljiva:
## Short description of the problem

When we annotate a contigs database with `anvi-run-cazymes`, most of the resulting hits in the gene functions table have `-` for an accession value:

<img width="694" alt="image" src="https://github.com/merenlab/anvio/assets/10360586/5456b739-bc7a-40e0-a36f-c904404b30ef">

It is not an issue coming from our code, but from the structure of the CAZyme HMM profiles, because we can see these undefined accessions in the `hmmscan` output when we run the program with `--debug` and check the temp output files:

```
$ head /var/folders/1n/2s6d_kq53pv9js812zwcljq80000gn/T/tmpiuhuh68p/hmm.table.fixed
438	-	CBM32.hmm	-	2.2e-34	114.8	1.7	1.6e-21	73.2	0.6	2.4	2	0	0	2	2	2	2	-
847	-	CBM32.hmm	-	1.7e-33	111.9	0.0	5e-21	71.6	0.0	2.6	2	0	0	2	2	2	2	-
871	-	CBM32.hmm	-	4.6e-27	91.1	0.3	1.5e-26	89.4	0.3	1.9	1	0	0	1	1	1	1	-
346	-	CBM32.hmm	-	3.4e-26	88.3	7.5	1.8e-25	86.0	2.8	2.6	2	0	0	2	2	2	1	-
892	-	CBM32.hmm	-	2.1e-25	85.7	0.0	4.5e-25	84.7	0.0	1.5	1	0	0	1	1	1	1	-
1069	-	CBM32.hmm	-	1.3e-14	50.9	5.2	2.4e-12	43.6	0.1	3.9	2	1	1	3	3	3	1	-
```

A related issue is that the function definition column contains the enzyme class names rather than the actual annotations (ie, `CBM32.hmm` when it would be more useful to see `Carbohydrate-Binding Module Family 32`, or `GH73.hmm` when it would be more useful to see `Glycoside Hydrolase Family 73`).  With the '.hmm' extension after the class ID number, these also look like filenames rather than annotations. 

The lack of unique accessions is a problem for anyone who wants to run `anvi-estimate-metabolism` with user-defined pathways using the CAZy database as a functional annotation source, because that requires a unique accession number to match each enzyme annotation to its pathway. It may also affect other downstream programs that rely on accession numbers like `anvi-display-functions`.

[...]

Issue #2099, opened 2023-08-01
[FEATURE REQUEST] update dbCAN2 → dbCAN3 for anvi-setup-cazymes


mschecht:
## The need

[dbCAN3](https://bcb.unl.edu/dbCAN2/) contains [dbCAN_sub](https://bcb.unl.edu/dbCAN_sub/) which enables substrate annotation for CAZymes. This upgrade will help catalyze CAZyme-related studies in anvi'o.

## The solution

Update [anvi-setup-cazymes](https://anvio.org/help/main/programs/anvi-setup-cazymes/) to download [dbCAN3](https://bcb.unl.edu/dbCAN2/).

## Beneficiaries

Biochemists and microbial ecologists interested in carbohydrate heterotrophy!

iwilkie:
@mschecht Thanks for the quick and detailed reply! Yes, exporting the AA sequence annotating it and then bringing that back into anvi'o works well, that's what I have been doing. I just stumbled upon this issue and was curious as to how much this function would be incorporated into anvi'o.

Thanks again! :)

mschecht:
@iwilkie I started working on this feature request [here](https://github.com/merenlab/anvio/tree/upgrade-to-dbCAN3) and got confused about where I can find dbCAN3 files, in particular, the dbCAN-sub HMMs.

Is this the dbCAN3 [dbCAN_sub](https://bcb.unl.edu/dbCAN2/download/Databases/dbCAN_sub.hmm) file? It's under the dir `/dbCAN2/`, but that's what's hyperlinked from the [dbCAN3 downloads page](https://bcb.unl.edu/dbCAN2/index.php).

xvazquezc:
@mschecht as far as I know the HMM dbs for dbCAN and dbCAN-sub are different. Not all prots with a CAZy domain (i.e. matching dbCAN HMM profile), match dbCAN-sub profiles. Some dbCAN-sub profiles are not necessarily made from a subset of a dbCAN family domain, e.g. the subfamily `PL6_e6` profile include sequences matching PL6 and CBM16

Coming back to the files, the one you indicate is the dbCAN-sub HMM, [this one is the current dbCAN HMM](https://bcb.unl.edu/dbCAN2/download/dbCAN-HMMdb-V12.txt) (barely one month old).

[...]

PR #2460, opened 2025-07-31
Update anvi-run-cazymes to V13

mschecht:
## Summary

This PR enhances [`anvi-run-cazymes`](https://anvio.org/help/main/programs/anvi-run-cazymes/) and updates [`anvi-setup-cazymes`](https://anvio.org/help/main/programs/anvi-setup-cazymes/) to support downloading **dbCAN version V13**.

## Related Issues

- Closes [#2148](https://github.com/merenlab/anvio/issues/2148)
  - Moves the CAZyme family name to the accession field in contigs-dbs - tagging @ivagljiva because this will have implications for user-defined metabolic pathways.
  - Added sanity check to not overwrite existing cazyme annotation without `--just-do-it`

- Closes [#2099](https://github.com/merenlab/anvio/issues/2099)
  - Expands support for dbCAN version V13
  - Implementing dbCAN3 which includes other HMM databases of CAZyme subfamilies and subtrate annotation was beyond the scope of this PR. I recommend in the documentation to import dbCAN3 annotations with [anvi-import-functions](https://anvio.org/help/7/programs/anvi-import-functions/).

## Testing

This PR can be tested with the component test: `run_component_tests_for_cazymes.sh`


meren:
(but apart from that discussion, I've made changes to the PR to make it compatible with the new pypackaging infrastructure via #2458).

yinlabniu:
> I have a question regarding this.
> 
> When I annotate a random [contigs-db](https://anvio.org/help/main/artifacts/contigs-db/) with `anvi-run-cazymes`, many of the hits have no description apart from the accession ID:
> 
> image
> This is not surprising since the collection of HMM models `anvi-setup-cazymes` downloads ([dbCAN-HMMdb-V13.txt](https://bcb.unl.edu/dbCAN2/download/Databases/V13/dbCAN-HMMdb-V13.txt)) does not have any additional information about each hit beyond the unique ID assigned to each model.
That is true. dbCAN fam HMMs are built for protein annotation at the CAZyme family level. CAZy (https://www.cazy.org/) defines these families with some "seed" proteins with EC numbers, and dbCAN just provides the HMMs for each family. For ~30 families, CAZy also classifies them into subfams, so dbCAN also includes HMMs for these CAZy defined subfams. However, connecting CAZyme family to EC and glycan substrate is not easy, because the well-known polyspecificity issue (one family may have multi-ECs and multi-substrates).
> 
> But when I was browsing the directory one level up from where the v13 collection of CAZyme HMMs are kept,
> 
> https://bcb.unl.edu/dbCAN2/download/Databases/
> 
> I found a file called [subfam_EC_mapping.tsv]
This is NOT the right file to use. This file is created from eCAMI result (https://github.com/yinlabniu/eCAMI). eCAMI was developed to classify ALL CAZyme families into subfams (unlike CAZy which only classifies ~30 families). 
(https://bcb.unl.edu/dbCAN2/download/Databases/subfam_EC_mapping.tsv). When I look for the accession ID I highlighted in the first screenshot, `GH73`, this is what I see:
> 
> ```
> curl https://bcb.unl.edu/dbCAN2/download/Databases/subfam_EC_mapping.tsv -o subfam_EC_mapping.tsv
> grep GH73 subfam_EC_mapping.tsv | awk 'BEGIN{FS="\t"} {print $1 "\t" $2 "\t" $5}' | anvi-script-tabulate
> ```
> 
> image
> So I get not only the EC numbers, but also some sort of an insight into what it might be.
> 
> When the accession ID is not like `GH73` but like `GH13_46` (with an underscore), then you don't get any direct hits in this file:
> 
> ```
> grep GH13_46 subfam_EC_mapping.tsv | wc -l
> 0
> ```
> 
> But if you search for these kinds of accession IDs like this: `GH13_46` -> `GH13_e46`, many times, there is a hit:
GH13_46 is from CAZy's subfam. Out of the ~30 CAZy classified families, GH13 happens to be one of them. GH13_e46 is from eCAMI. We put "e" in the subfam name to distinguish from CAZy's subfam. See some explanation in https://academic.oup.com/nar/article/51/D1/D557/6833251. 

image

So GH13_46 is NOT GH13_e46. Do NOT use the EC/substrate of GH13_e46 for GH13_46. I believe Anvio only integrates dbCAN HMMs not eCAMI, so "XX_eXX" won't appear in anvi-run-cazymes result.

> 
> image
> Another example -- the hit is `GH13_9` -> `GH13_e9`:
Again, GH13_9 and GH13_e9 are not the same thing.

> 
> image
> There are also cases like this. Here is the hit `PL8_4`, but there is no hit for `PL8_e4`:
They are not the same thing. Do not use subfam_EC_mapping.tsv for finding glycan substrates. https://bcb.unl.edu/dbCAN2/download/run_dbCAN_database_total/fam-substrate-mapping.tsv is more useful, but it does not include CAZy subfams. The substrate columns are what we manually curated from CAZy website (www.cazy.org).

> 
> ```
> grep PL8_4 subfam_EC_mapping.tsv | wc -l
> 0
> ```
> 
> But if you go back to `PL8`, then there are many hits:
> 
> image
> One can't be sure how to annotate this (as in should one go with host glycan, xanthan, or alginate? Or simply say "may be host glycan, or xanthan, or alginate" as an annotation), one could say that that the enzyme class is [4.2.2.-](https://enzyme.expasy.org/EC/4.2.2.-).
> 
> This is all to say, I think we can do much more that what we are doing with the CAZyme database to populate contigs-db tables with potentially useful information (and there may be even better files to work with in that directory, I haven't exhaustively explored it).
> 
> I just wanted to ask if you think this is worthy of another look (I'm happy to do it myself). What do you think y'all? (pinging @ivagljiva and @semiller10 here).

In summary, if you use anvi-run-cazymes to perform CAZyme annotation, you will only get CAZyme family info (and for ~30 families you also get CAZy subfams). CAZy subfams do not have "e" in the name, and are not equal to eCAMI (which has "e" in the names). If you are not satisfied with only CAZyme family annotation and want to go one step further to get glycan substrate info, use https://bcb.unl.edu/dbCAN2/download/run_dbCAN_database_total/fam-substrate-mapping.tsv. We update this file each year.

meren:
Dear @yinlabniu, thank you very much for this detailed response and for your patience. I knew I was proposing something rather stupid, but I should have better captured the differences between these resources before wasting your and others' time. I apologize for that.

Please let us know if you have any thoughts on what would be the best way to provide users with some additional insights here beyond leaving them with `GH25` or `GT4` as their final functional annotation for their genes. We would be very happy to follow your lead here if you have any suggestions, otherwise I think we will leave `anvi-run-cazymes` as is :)

Thanks again,
Meren

yinlabniu:
> @meren the thing is that for the default HMM collection `GH25` would be the actual annotation. If you look into the CAZy website, they have removed the list of activities that used to be included because it was more misleading than helpful. 
It is not removed. Instead it's moved to a specific page, e.g., : https://www.cazy.org/GH5_activity.html. GH25 is simple: if a protein is assigned to GH25, it's most likely for peptidoglycan degradation. In this sense, eCAMI subfams for GH25 under-annotate protein queries for substrates. However, GH5 is more complex as it has almost 30 ECs, so having eCAMI subfams does help, even if some subfam still has multiple ECs (as @xvazquezc pointed out below).
The families are more defined based on structure and mechanism of action (like protein superfamilies) than on the substrate. image
> 
> Even if you match the subfam/eCAMI, you can have proteins with different activities under those (also, most of them have no known EC associated anyway). For example, for `GH5_e55`, has members with 4 different EC numbers with 3 different substrates (which covers only 27 out of 32 clustered seq domains, i.e. 5 of them are completely unkwown): image
> 
It is very true that most eCAMI subfams have no ECs so subfam assignment does not help with substrate annotation. However, this is at least one step further toward glycan substrate annotation. If no subfams, one has to stop at just fam level assignment, or they have take all possible substrates from a fam. For the GH5 example, it will be 12 possible substrates (according to https://bcb.unl.edu/dbCAN2/download/run_dbCAN_database_total/fam-substrate-mapping.tsv).
image

Unfortunately, this is what most people do currently ...
> One way to make it more reliable in terms of annotations could be to run DIAMOND on the seqs with HMM matches to eCAMI that have known EC/substrate (most eCAMI have no EC for any seq at all). If the DIAMOND search matches the sequence (given certain parameters), it'd be annotated, otherwise the substrate would be left empty. This could prob be a separated function more associated with metabolic modelling than annotation itself.

What you suggested might be okay but a direct diamond search against CAZy seqs (https://bcb.unl.edu/dbCAN2/download/CAZyDB.07142024.fa or a subset of it with only CAZymes having ECs) is easier and can also do the work. A lot of general functional annotation tools like Humann4 just uses KO but it's for sure not complete for CAZyme annotation. CAZy/dbCAN is certainly more useful in this regard as even just assigning a protein to a CAZyme fam is a useful info. Knowing the substrates from the fam is not a simple task ...

Issue #2468, opened 2025-09-01
[BUG] Missing annotations when combining HMMER output from multiple threads


ivagljiva:
## Short description of the problem

When we run HMMER in multiple threads, not all of the per-thread output is combined into the final `hmm.table` file that anvi'o processes, leading to missing annotations. This is a bug that 
- was introduced recently in #2423 
- hopefully affects only large contigs databases (I found it while testing on a database containing ~3 million genes) given that we had tested #2423 on smaller cases and found no issues
- likely impacts any anvi'o program running our HMMER driver in the backend like `anvi-run-pfams`, although I found the bug while testing `anvi-run-kegg-kofams`
- might be a race condition and likely has some degree of stochasticity in terms of which HMMER hits are missing

## anvi'o version

anvio-v8-dev as of commit 5eb9eea88ac92b07012aaabe0e296e6006d487b6 
(See below for details on how I tested to determine that this change introduced the bug)

## Detailed description of the issue

Shoutout to Dylan Mankel (@mankeldy) who [identified this issue on Discord](https://discord.com/channels/1002537821212512296/1398766743148695702) and provided the test data. They noticed that some of their expected KOfam annotations were missing after they recently re-ran `anvi-run-kegg-kofams`, and they gave me one of their huge metagenome samples to test with.

### Identifying the problem
I ran the following command on their metagenome:
```
anvi-run-kegg-kofams -c [CONTIGS.db] --kegg-data-dir KEGG_4039c79d9c82/ -T 40 --debug --tmp-dir no_prefilter_tmp/ --just-do-it --no-hmmer-prefiltering
```

When looking at the debug output for a gene that was erroneously missing an annotation, I realized that it was not present in the `hmm.table` file, which is supposed to contain the combined HMMER table output from all threads, but it _was_ present in the thread-specific HMMER table, ie `AA_gene_sequences.fa.25_table`:

```
$ grep K00399 no_prefilter_tmp/*/*hmm.table | grep 2148300
# no output
$
$ grep K00399 no_prefilter_tmp/*/AA_gene_sequences.fa.*_table | grep 2148300
no_prefilter_tmp/tmpvo8a37hm/AA_gene_sequences.fa.25_table:2148300              -          K00399               -                  0 1044.3   0.0         0 1044.1   0.0   1.0   1   0   0   1   1   1   1 -
```

When I counted the total number of HMM hits across all thread-specific outputs, I got a high value of around 90 million hits:
```
[...]

PR #2469, opened 2025-09-01
Fixing batch processing of HMMer outputs being appended to main table

bcoltman:
This pull request addresses the bug report #2468  identified by @mankeldy and reported by @ivagljiva.

This fixes errors introduced in #2423. The intent of pull request #2423 was to process up to 1,000,000 hits per thread. When a full ```batch``` was created (i.e. 1 million hits) then the variable ```output_files``` was generated. This was never processed, filtered or output in the final HMM hits output table. Subsequently, if a file contained e.g. 2.5 million hits, then the first two batches (of 1 million hits each) would not be output, and only the final 0.5 million would be.  This is a very silly error.

This update fixes that mistake - each batch, regardless of whether it is full (1 million) or not, is processed and written to the output file. This fix keeps the the intended time and memory savings of #2423, but actually works now.

To test it, I reannotated an anvio database as so:
```
anvi-run-kegg-kofams -c GCA_000690595.db -T 4 --just-do-it --kegg-data-dir resources/anvio_ref_dbs/kegg/ --include-stray-KOs --debug
```

and confirmed whether all the hits (normal and stray) were output in the final hmm..table:
```
(anvio-dev) $ grep -v -e "^#"  /tmp/tmpzng8q_ib/AA_gene_sequences.fa.*_table | wc -l
93478
(anvio-dev) $ wc -l /tmp/tmpzng8q_ib/hmm.table
93478 /tmp/tmpzng8q_ib/hmm.table
(anvio-dev) $ grep -v -e "^#" /tmp/tmptzqmr6w7/AA_gene_sequences.fa*_table | wc -l
3143
(anvio-dev) $ wc -l /tmp/tmptzqmr6w7/hmm.table
3143 /tmp/tmptzqmr6w7/hmm.table
```

I also changed the number of lines per chunk (to 100), which would previously cause only incomplete batches (<100 lines) to be output to the final hmm.table.

```
(anvio-dev) $ grep -v -e "^#" /tmp/tmpureyqili/AA_gene_sequences.fa.*_table | wc -l
93478
(anvio-dev) $ wc -l /tmp/tmpureyqili/hmm.table
93478 /tmp/tmpureyqili/hmm.table
(anvio-dev) $ grep -v -e "^#" /tmp/tmpbxdio9op/AA_gene_sequences.fa.*_table | wc -l
3143
(anvio-dev) $ wc -l /tmp/tmpbxdio9op/hmm.table
3143 /tmp/tmpbxdio9op/hmm.table
```

Many thanks to @ivagljiva and @mankeldy for finding this issue. A very annoying mistake and I apologise for introducing it in the first case.

### DIFF
--- a/anvio/drivers/hmmer.py
+++ b/anvio/drivers/hmmer.py
@@ -450,7 +450,6 @@ class HMMer:
 
         detected_non_ascii = False
         lines_with_non_ascii = []
-        output_lines = []
 
         def process_batch(line_batch, base_line_number):
             nonlocal detected_non_ascii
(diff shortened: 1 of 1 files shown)


bcoltman:
Great, thanks a lot @ivagljiva!

PR #2471, opened 2025-09-02
Fixing issues with stray KOs breaking estimate-metabolism

### LINKED ISSUE #2466: [BUG] stray KOs breaking anvi-estimate-metabolism

ConfigError when using anvi-estimate-metabolism due to stray KOfams being incorrectly processed. 

## anvi'o version

```
anvio-dev - most recent (8df86a6)
```

## kegg data version

Whilst it is not the most recent version, I don't believe this to be the issue.
```
v2024-03-09:
url: https://figshare.com/ndownloader/files/44953354
archive_name: KEGG_build_2024-03-09_23910d68b4f2.tar.gz
hash: 23910d68b4f2
modules_db_version: 4
no_modeling_data: True
no_binary_relations: True
no_maps: True
```

I generate a contigs DB, annotate it, map reads to my metagenomes and profile it:

```
    anvi-gen-contigs-database -f $DEST_FASTA \
                              --num-threads $THREADS \
                              --project-name ${SRA_ACCESSION} \
                              -o $META_BASE/anvio/database/${SRA_ACCESSION}.db

    # annotate DBs
    anvi-run-kegg-kofams -T $THREADS \
                         -c $META_BASE/anvio/database/${SRA_ACCESSION}.db \
                         --kegg-data-dir $DBCACHE/kegg/ \
                         --include-stray-KOs

    anvi-profile -i ${TMPDIR}/${SRA_ACCESSION}.bam \
                 -c $META_BASE/anvio/database/${SRA_ACCESSION}.db \
                 -o $META_BASE/anvio/profile/${SRA_ACCESSION} \
                 -T $THREADS \
                 --min-contig-length 500 \
                 --skip-hierarchical-clustering
```

I then have an issue running:

```
    anvi-estimate-metabolism -c $META_BASE/anvio/database/${SRA_ACCESSION}.db \
                             -p $META_BASE/anvio/profile/${SRA_ACCESSION}/PROFILE.db \
                             --kegg-data-dir $DBCACHE/kegg/ \
                             --include-stray-KOs \
                             --add-coverage \
                             --output-modes modules,hits \
                             --output-file-prefix $META_BASE/anvio/annotations/${SRA_ACCESSION}/${SRA_ACCESSION}_estimate
```
where I receive the following error:

```
Config Error: We cannot find the KEGG enzyme K15971_anvio_version in the dictionary of enzyme
              hits, even though this enzyme is annotated in your data. There are 3 main ways
              this can happen: (1) you are using --enzymes-txt input that includes KOs that
              are different from the set used for annotation with `anvi-run-kegg-kofams`, (2)
              your contigs database was annotated with `anvi-run-kegg-kofams --include-stray-
              KOs` but you didn't use the `--include-stray-KOs` flag for `anvi-estimate-
              metabolism`, or (3) you imported external annotations with the source name
              `KOfam` that include KOs not in the KEGG data directory currently being used.
              You have a few options to get around this error depending on which case applies
              to your situation. If this is case (2) and you want to include these enzymes in
              the analysis, then re-run `anvi-estimate-metabolism` with the flag `--include-
              stray-KOs`. If this is case (1) or (3) and you want to include these enzymes in
              the analysis, then re-run `anvi-estimate-metabolism` with the flag `--include-
              kos-not-in-kofam`. And no matter what the situation is, if you want to IGNORE
              these unknown annotations for the purposes of estimating metabolism, you can re-
              run `anvi-estimate-metablism` with the flag `--ignore-unknown-KOs`. If this
              message made you worry, you could also re-do your annotations or remove these
              unknown enzymes from your input --enzymes-txt file to be on the safe side.
```

I modified the kegg.py script to print the kos at [this point](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4708) in the ```mark_kos_present_for_list_of_splits``` function and found that the ```knum``` in [all_kos](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4689) include the ```knum``` as well as the anvio version suffix -e.g. both K15971 and K15971_anvio_version are printed. Currently, entries are only checked for presence in ```self.ko_dict```. However, stray KOs are removed from ```self.ko_dict``` and will therefore fail [this check](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4709C37-L4709C61). Updating the code to search for ```f"{knum}{STRAY_KO_ANVIO_SUFFIX}"``` in ```self.stray_ko_dict``` would fix this, but only ```knum``` without the suffix would pass the if statement and subsequently there would only be an entry for the ```knum``` without suffix in the ```bin_level_ko_dict``` e.g. only for K15971. 

I then confirmed that the ```ko``` derived from [for ko, gene_call_id, split, contig in kofam_hits_in_splits:](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4739)
does include the ```_anvio_version``` suffix. The presence of this ```ko``` is then confirmed in ```bin_level_ko_dict``` [here](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4751). However, as the ```bin_level_ko_dict``` is populated exclusively with KOs without suffixes, then this currently fails and leads to the original ConfigError. Later in the code, the KO is [updated](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4780). Moving this block of code to [here](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4751) fixed this issue with the lookup in ```bin_level_ko_dict```. This would also lead to the entry in ```bin_level_ko_dict``` without the anvio version suffix being [updated later](https://github.com/merenlab/anvio/blob/8df86a6994fa0724219aa83e9a72b3925bc7e9fb/anvio/kegg.py#L4806) as well as in ```bin_level_module_dict[m]["kofam_hits"][ko]```. Subsequently, this allows for the ```anvi-estimate-metabolism``` function to run without error - when inspecting the DEBUG output, I see an entry for ```Generating output for KO .....................: K15971```, without the suffix. 

Without fully understanding the code and the intended flow, I didn't want to implement my edits and open a pull request before discussing it further.  I am unsure if there is any connection with a previous issue #2336 with ```estimate-metabolism``` that we resolved. I believe it might do as that issue removed the addition of stray KOs to the main ko_dict. 

Some general questions are: 
- What should the keys in ```bin_level_ko_dict``` be? Should they include the ```_anvio_version``` suffix?
- How should the KO in ```bin_level_module_dict[m]["unique_to_this_module"].add(ko)``` and ```bin_level_module_dict[m]["kofam_hits"][ko] = [gene_call_id]``` be formatted? Should they include the ```_anvio_version``` suffix?

bcoltman:
### Summary

This PR fixes a `ConfigError` raised by `anvi-estimate-metabolism` when a database is annotated with `--include-stray-KOs`. The error was caused by inconsistent handling of stray KO accessions (e.g. `K15971_anvio_version`) during lookups and output assembly. After this change, `anvi-estimate-metabolism` runs to completion and reports results consistently.

**Closes #2466  .** (Related context: #1953   #2336  )

---

### Background & root cause

* With `--include-stray-KOs`, `self.all_kos_in_db` contains both base KO accessions (e.g. `K15971`), their associated stray variants (if relevant, e.g. `K15971_anvio_version`) and some module numbers.

* The “sanity check” and subsequent lookups assumed all relevant accessions lived in `self.ko_dict`. After #2336 removed stray KOs from `self.ko_dict`, those checks no longer matched, leaving some accessions out of `bin_level_ko_dict` and ultimately triggering:

  ```
  Config Error: We cannot find the KEGG enzyme K15971_anvio_version in the dictionary of enzyme hits ...
  ```
* The core mismatch: we mixed accession forms. We sometimes indexed with the `_anvio_version` suffix, while downstream structures and outputs expect the base KO accession.

---

### What this PR changes

1. **Early normalisation of stray KO accessions**

   * In `kegg.py` (`mark_kos_present_for_list_of_splits`), after extracting the `ko` from `kofam_hits_in_splits` we detect the `_anvio_version` suffix up front, record a flag (`is_anvio_version`), and **replace the working accession with the base KO** before any dictionary lookups or insertions.
   * Effect: all bookkeeping and lookups use base KO accessions consistently.

2. **Robust sanity check for available profiles**

   * The previous check assumed all profiles were in `self.ko_dict`.
   * We now simply verify that a usable profile exists **in either** `self.ko_dict` (KEGG HMM) **or** `self.stray_ko_dict` (anvi’o HMM for stray KO). 
   * This check is okay as `self.stray_ko_dict` is only populated if `--include-stray-KOs` is specified.
   * It is important to note that there are two types of stray KOs; 
        * `knum` whose profiles are trained on multiple sequences have a `knum` WITH a `STRAY_KO_ANVIO_SUFFIX` added to the `self.stray_ko_dict`
        * Those based on a single sequence remain as `knum` without the suffix. 
   * Importantly, we check for presence in `self.stray_ko_dict` for the `knum` with and without the `STRAY_KO_ANVIO_SUFFIX`. This ensures that all stray `knum` that have a profile pass the sanity test (i.e. are FALSE). 
   * Only KOs with no available profile are excluded when `self.exclude_kos_no_threshold` applies.

3. **Consistent dictionaries**

   * `bin_level_ko_dict` now **stores only base KO accessions** (no `_anvio_version` keys).
   * `bin_level_module_dict[m]["kofam_hits"]` and `["unique_to_this_module"]` remain base KO accessions (unchanged in format).

4. **Transparent warnings when a stray HMM was used**

   * When a KO was annotated via an stray model, we:

     * keep the base KO in all metrics and IDs; and
     * attach a human-readable warning.
   * Warnings are:

     * preserved at module level (existing behaviour), and
     * **added to the “hits” output** as a new `warnings` column so users can see when a `_anvio_version` HMM contributed to a hit.

5. **Problems still present**

   * Suffix-appending logic is still used in the sanity check (e.g. there are still `self.ko_dict` lookups for  `..._anvio_version_anvio_version`).

---

### User-facing impact

* The crash reported in #2466 no longer occurs when running with `--include-stray-KOs`.
* Output accession formatting is stable and consistent (base KO accessions everywhere).
* **New:** the *hits* mode now includes a `warnings` column when a KO was annotated using an anvi’o stray HMM (e.g. `used 'K15971_anvio_version' model to annotate K15971`).

---

### Reproduction

A pipeline for reproduction was included in #2466  but the issue could more simply be reproduced with the following

```bash
anvi-estimate-metabolism \
  -c [CONTIGS_DB].db \
  --include-stray-KOs \
  --output-modes modules,hits \
  --output-file-prefix [OUTPUT]_estimate
```

I have shared a small database with @ivagljiva that led to a `ConfigError` when running the above code. After this PR, the code completes and the appropriate warnings are included in the output files.

### Remaining points
1. `self.all_kos_in_db` yields `knum` with and without `_anvio_version` suffixes, in addition to some module accessions. I introduced a lookup whether `knum.startswith("M")` to prevent modules being added `bin_level_ko_dict[knum]` - please confirm whether this is okay or whether they should be present in `bin_level_ko_dict[knum]`.
2. In the bug report #2466 @ivagljiva mentioned "output-related constants/functions". I inspected the outputs and the appropriate warnings were printed in `_modules.txt` (e.g. used 'K15971_anvio_version' model to annotate K15971). I have not yet tested the other outputs and would ask for you to confirm whether the current output is as expected.

### DIFF
--- a/anvio/kegg.py
+++ b/anvio/kegg.py
@@ -4499,7 +4499,7 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
                                                                where_clause=hits_where_clause)
 
         contigs_db.disconnect()
-
+        
         # combine the information for each gene call into neat tuples for returning
         # each gene call is only on one split of one contig, so we can convert these lists of tuples into dictionaries for easy access
         # but some gene calls have multiple kofam hits (and some kofams have multiple gene calls), so we must keep the tuple structure for those
@@ -4622,8 +4622,7 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
         # we update the header list for the affected modes
         self.available_modes["hits"]["headers"].extend(kofam_hits_coverage_headers + kofam_hits_detection_headers)
         self.available_modes["modules"]["headers"].extend(modules_coverage_headers + modules_detection_headers)
-
-
+    
     def mark_kos_present_for_list_of_splits(self, kofam_hits_in_splits, split_list=None, bin_name=None):
         """This function generates two bin-level dictionaries of dictionaries to store metabolism data.
 
@@ -4695,6 +4694,7 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
                                            "unique_to_this_module": set(),
                                            "warnings" : set()
                                           }
+        
         for knum in all_kos:
             """
             We can only add warnings about missing KOfam profiles because for other annotation sources, we don't
@@ -4705,8 +4705,17 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
             Furthermore, this can only be done when we are using both KEGG data and user data (ie, not --only-user-modules)
             because we need access to the self.ko_dict
             """
-            if not self.only_user_modules and self.all_kos_in_db[knum]['annotation_source'] == 'KOfam' and knum not in self.ko_dict \
-                        and (f"{knum}{STRAY_KO_ANVIO_SUFFIX}" not in self.ko_dict) and self.exclude_kos_no_threshold:
+            
+            if knum.startswith("M"):
+                continue
+            
+            if (not self.only_user_modules 
+                and self.all_kos_in_db[knum]['annotation_source'] == 'KOfam' 
+                and knum not in self.ko_dict 
+                and knum not in self.stray_ko_dict 
+                and f"{knum}{STRAY_KO_ANVIO_SUFFIX}" not in self.stray_ko_dict
+                and self.exclude_kos_no_threshold):
+                
                 mods_it_is_in = self.all_kos_in_db[knum]['modules']
                 if mods_it_is_in:
                     if anvio.DEBUG:
@@ -4728,25 +4737,37 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
                                          f"re-running this program with the --include-kos-not-in-kofam flag.")
 
                 continue
-
+            
             bin_level_ko_dict[knum] = {"gene_caller_ids" : set(),
                                      "modules" : None,
                                      "genes_to_contigs" : {},
-                                     "contigs_to_genes" : {}
+                                     "contigs_to_genes" : {},
+                                     "warnings" : set()
                                      }
 
         kos_not_in_modules = []
+        
         for ko, gene_call_id, split, contig in kofam_hits_in_splits:
+            # make sure we can count annotations to anvi'o versions of stray KO models by using KEGG's original accession
+            is_anvio_version = False
+            if ko.endswith(STRAY_KO_ANVIO_SUFFIX):
+                is_anvio_version = True
+                ko = ko.replace(STRAY_KO_ANVIO_SUFFIX, "")
+                
             if ko not in self.all_kos_in_db:
+                
                 kos_not_in_modules.append(ko)
+                
                 # KOs that are not in modules will not be initialized above in the ko hit dictionary, so we add them here if we haven't already
                 if ko not in bin_level_ko_dict:
                     bin_level_ko_dict[ko] = {"gene_caller_ids" : set(),
                                              "modules" : None,
                                              "genes_to_contigs" : {},
-                                             "contigs_to_genes" : {}
+                                             "contigs_to_genes" : {},
+                                             "warnings" : set()
                                              }
             else:
+                
                 # if we are missing the KO from the dictionary at this point, we should fail nicely instead of with a KeyError
                 if ko not in bin_level_ko_dict:
                     if self.ignore_unknown_kos:
@@ -4765,6 +4786,7 @@ class KeggMetabolismEstimator(KeggContext, KeggEstimatorArgs):
                                       f"metabolism, you can re-run `anvi-estimate-metablism` with the flag `--ignore-unknown-KOs`. If this message "
                                      f"made you worry, you could also re-do your annotations or remove these unknown enzymes from your input "
                                      f"--enzymes-txt file to be on the safe side.")
+                
                 present_in_mods = self.all_kos_in_db[ko]['modules']
                 bin_level_ko_dict[ko]["modules"] = present_in_mods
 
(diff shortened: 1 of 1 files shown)


bcoltman:
Great work, thanks @ivagljiva ! 

Additional thanks for the very helpful explanations!

PR #2478, opened 2025-09-12
New SamplesTxt class

FlorianTrigodet:
New class to handle the samples-txt artifact. Will replace the utils function what was solely used by the primer search class. I will also replace the current functions in the metagenomics workflow that handles samples-txt with this new class.

### DIFF
--- a/anvio/artifacts/samples_txt.py
+++ b/anvio/artifacts/samples_txt.py
@@ -107,25 +107,47 @@ class SamplesTxt:
         """Return the raw TSV text of the samples-txt file."""
         return self._raw_text
 
-    def as_dict(self):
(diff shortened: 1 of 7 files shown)


FlorianTrigodet:
Changes made in:
- [x] metagenomics workflow
- [x] ecophylo workflow
- [x] trna workflow
- [x] samples-txt documentation

For all workflows, the samples-txt is expected to be paired-end reads for compatibility with current code behavior. I will later change the metagenomics workflow to allow for LR, but that's another story. 

I ran the tests for the metagenomics and ecophylo workflow and everything works as expected. 
There is no test for the trnaseq workflow, so I don't know if I made a mistake there. The changes should be minimal and could better use the new class, but I don't want to mess with the code without a test case.

PR #2481, opened 2025-09-26
Changing the terminology of 'stray KOs' to 'no-threshold KOs'

ivagljiva:
About 1.5 years ago, we added the capacity for annotating KEGG Orthologs that come without bit score thresholds: https://github.com/merenlab/anvio/pull/2237

They were called 'stray KOs' in the codebase since then. However, earlier this year we were part of [a paper benchmarking annotation of KOs](https://doi.org/10.1093/bioadv/vbaf039) that used the `--include-stray-KOs` functionality, but while iterating over drafts of that paper we ultimately settled on a better, more descriptive name for these things: 'no threshold KOs' or 'nt-KOs' for short.

I thought it would be best if the user-facing parts of the metabolism codebase were to use the terminology consistent with that publication. Hence, in this PR, all instances of 'stray KO' in the program flags, terminal outputs, and documentation pages have been changed to 'nt-KOs'.

The old flag `--include-stray-KOs` has been changed to the new flag `--include-nt-KOs` for all programs in the metabolism suite. That said, you _can_ still use the old flag, thanks to @meren's addition of a synonym flag. Any old workflows using `--include-stray-KOs` will still function properly :)  
(pinging @Kekananen here just so you are aware of this change 👋)

Note that none of the variable names or the file names in the KEGG data directory related to nt-KOs have been renamed, so there are still instances of 'stray KO' terminology where developers can see them. It is only a superficial change to avoid breaking everything in the backend and destroying backwards-compatibility with older KEGG data snapshots. :)

I am currently running `anvi-self-test --suite metabolism` to make 100% sure that everything still works, and once that is done I will merge this.

### DIFF
--- a/anvio/metabolism/algorithms.py
+++ b/anvio/metabolism/algorithms.py
@@ -896,11 +896,11 @@ class KeggEstimationAlgorithms:
         return now_complete
 
 
-    def add_module_coverage(self, mod, meta_dict_for_bin, profile_db=None, enzymes_of_interest_df=None, coverage_sample_list=None):
+    def add_module_coverage(self, mod, meta_dict_for_bin, profile_db=None, enzymes_of_interest_df=None):
         """This function updates the metabolism dictionary with coverage values for the given module.
 
-        It must be called after init_gene_coverage() or add_gene_coverage_to_headers_list() so that
-        the self.profile_db attribute is established.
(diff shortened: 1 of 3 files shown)


ivagljiva:
All tests passed :)

PR #2477, opened 2025-09-09
Ability to infer the involvement of a list of genes/gene clusters in metabolism through improvements to programmatic access to KEGG estimator class

meren:
The purpose of this PR is to do what title says. Which required some fundamental changes in the codebase, which included,

* Refactoring `anvio/kegg.py` and split it into multiple modules under `anvio/metabolism/`,
* Enable programmatic access to completion estimations of metabolism without an enzymes-txt file,
* Adding new functions to `ContigsSuperclass` and `PanSuperclass` so they can estimate metabolic module involvement of a list of genes and gene clusters,
* Updating relevant bottleroutes functions to recover metabolic insights,
* Updating the JavaScript code to process and visualize such data.

Since it is now possible to do the following,

```python
from types import SimpleNamespace
from anvio.dbops import PanSuperclass

gcs = ["GC_00002419", "GC_00002456", "GC_00002454", "GC_00002452", "GC_00002451", "GC_00002450",
       "GC_00002447", "GC_00002446", "GC_00002445", "GC_00002444", "GC_00002443", "GC_00002442",
       "GC_00002439", "GC_00002438", "GC_00002437", "GC_00002436", "GC_00002435", "GC_00002434"]

pan = PanSuperclass(SimpleNamespace(pan_db="PAN.db", genomes_storage="GENOMES.db"))

metabolism_d, kofams_d, pan.get_metabolism_estimates_for_a_list_of_gene_clusters(gene_clusters_of_interest=gcs)
```

We literally can do this:

https://www.youtube.com/watch?v=BBiVQnuv29M

PR #2490, opened 2025-11-05
Show gene functions and contributions to metabolism in gene-mode

meren:
This PR introduces the new features implemented in #2477 to interactive displays in `gene-mode`.

Using the INFANT-GUT-DATA as a test case, when we run the following command,

```
anvi-interactive -p PROFILE.db \
                 -c CONTIGS.db \
                 -C default \
                 -b E_facealis \
                 --gene-mode
```

# Before this PR

This is what we see in the bins panel after a few selections:

image

And this is what happens when we click the 'Splits' button (where even the name makes no sense, let alone the irrelevant completion and redundancy columns):

image

Misérable, en effet.

# After this PR

After this PR, running the same command and action shows this in the bins panel:

image

And this is what happens when we click the 'Genes selected' button:

image

# What will this do?

It will improve our ability to make sense of the functions of genes and their involvement in metabolism directly from the interface as we look at changes in transcription or peptide counts (like the stuff @semiller10 has been doing lately).

And this happens with only 22 lines-of-code change. Talk about efficiency :p

### DIFF
--- a/anvio/data/interactive/js/main.js
+++ b/anvio/data/interactive/js/main.js
@@ -2214,9 +2214,9 @@ function buildContigAndSplitNamesTable(response, config) {
     const splitNames = response && response.split_names;
 
     let contigAndSplitNamesContent = `
-        <p style="font-size: large; border-bottom: 1px solid black; background: #f5f5dc9c;">Contig Names</p>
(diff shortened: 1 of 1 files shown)


meren:
Example bin from the Infant Gut Dataset:

image

PR #2492, opened 2025-11-10
tRNA taxonomy v226

FlorianTrigodet:
New version of the tRNA taxonomy using GTDB v226.
I ran a quick test and it worked fine. I can also see that the database version's check are working.
I also noticed that in the original commit for the first and only version of the tRNA taxonomy database in anvi'o that we would make a README file showing how we generated the data. But, alas, no such README was ever committed.

I will write a short one explaining how I got the data and how I transformed it.


meren:
This PR has been waiting for the  `anvi-setup-trna-taxonomy` help docs to be updated with what is now in the README.txt file.

PR #2505, opened 2025-12-07
Taxonomy inspired bin labels for interactive binning

meren:
This PR addresses one of the most annoying things in the anvi'o interactive interface when it is used for binning efforts.

Every time the user clicks on a bunch of contigs to add them into a bin, the automatically generated bin names go Bin_1, Bin_2, etc. Here is an example from the IGD:

image

A few years ago, we added real-time taxonomy estimation for bins, which gives us really nice initial names for bins if everything is in place:

image

Which is great, but it doesn't solve the problem of dumb bin names :/ This causes a downstream headaches since our default bin names doesn't mean anything even when our own interface knows much more about them. So the user either has to manually change all the names by looking at the taxonomic information before storing their bins in a collection, or they have to work with these dumb bin names for much longer.

This PR adds a new button in the interface:

image

When clicked, this optional button makes the following magic happen, and replaces initial bin names with something much more useful:

image

The option 'Best' uses the best taxonomy available. But the user can specifically ask for a particular taxonomic level to be used for labeling bins (which was very useful when working with very novel populations):

image

When stored as a collection, these bin names are used for everything downstream, making life much easier.

I tested it with IGD, as well as some novel data I've been playing with.

### DIFF
--- a/anvio/data/interactive/js/bin.js
+++ b/anvio/data/interactive/js/bin.js
@@ -79,6 +79,9 @@ function Bins(prefix, container) {
         'taxonomy': {}
     };
 
+    this.manualBinNamesBackup = null;
+    this.taxonomyLabelingEnabled = false;
+
     this.keepHistory = false;
     this.allowRedraw = true;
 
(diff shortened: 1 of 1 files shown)


ivagljiva (approved):
I tested it out @meren and I think it is great :)

meren:
Thank you, Iva!

PR #2513, opened 2025-12-26
Make pangenome graph code work with identical genomes

meren:
I was running `anvi-pan-genome-graph` with many many genome sets, and I run into a situation where the genomes were so similar, there was no variation in the resulting graph, and things started crashing down.

No one cares about working with graph structures with no variation, but the code should handle that gracefully so we can run things on many genome sets automatically without exploding the code.

This PR solves this issue by adding a bunch of defensive checks that only activate in such edge cases. Everything else runs identically.

### DIFF
--- a/anvio/panops.py
+++ b/anvio/panops.py
@@ -1581,6 +1581,15 @@ class PangenomeGraph():
         # generate pan-graph-db and populate it with information
         self.generate_pan_graph_db()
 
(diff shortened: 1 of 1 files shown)


meren:
Just merging this, but wanted to have a PR so when @ahenoch is back in the new year he can take a look at the commits to see what is changed :)

PR #2536, opened 2026-02-02
remove prodigal driver and multithread class

FlorianTrigodet:
We still have a prodigal driver and a class in threadingops.py and the only place where prodigal is imported is in genecalling.py but no user can request the use of Prodigal from the command line interface.

The only reason we could keep it is if (1) we plan to make it available to the user, or (2) when someone uses anvi'o classes/function outside of the command lines.

@meren, let me know your thoughts. If we should keep it, just delete the branch.


EricDeveaud:
I woul remove the use of prodigal as there is a long standing bug that still unpatched in “plain” prodigal 
Pyprodigal fix it
Nb I can provide a patch vs “plain” prodigal if you which it
Regards 
Eric

PR #2538, opened 2026-02-09
Profiler prefilter and memory optimization

FlorianTrigodet:
A lot of changes in a critical part of anvi'o. WHAT COULD GO WRONG?

The original issue is that the profiler would run out of memory with large dataset, particularly when the contigs.db is very large. In a test case, the contigs.db has >7M contigs (large collection of MAGs) and the mapping was very sparse with only ~4% of contigs having at least one read mapping to them. Yet the current profiler will profile ALL CONTIGS and load terribly large data structure like nt_positions for SNVs profiling.

All the code changes were done by Claude Code, and most of the commit message are from Claude as well. I also the LLM to sumarize the key changes:

# Key changes

1) --skip-contigs-without-coverage: BAM index pre-filtering

  When profiling a metagenome against a large reference (e.g., 7.5M contigs from a MAG database), most contigs have zero coverage. This flag scans the BAM index to identify contigs with at least one mapped read, and only loads data for those contigs. In our test case, this reduced the working set from 7.5M contigs to 292K (3.86%).

2) Shared memory for large data structures

  contig_sequences and nt_positions_info (totaling ~24GB for 292K contigs) are packed into OS-level shared memory (/dev/shm) instead of being held as Python objects in each worker process. Workers attach to these segments read-only through lightweight proxies (SharedContigSequencesProxy, SharedNtPositionsProxy), so the data exists exactly once in physical RAM regardless of the number of workers.

3) Subprocess data loading to avoid heap fragmentation

  The shared memory segments are loaded and packed in a short-lived subprocess rather than the main process. This prevents glibc heap fragmentation: when the subprocess exits, the OS reclaims its entire heap. If the data were loaded in the main process, freed memory would remain trapped in the heap (glibc's malloc_trim can only shrink from the top), and each forked worker would inherit those unreclaimable pages.

4) WorkerContext: lightweight proxy for worker processes

  Instead of passing the entire BAMProfiler object to each worker, a minimal WorkerContext is built carrying only the ~22 attributes and 4 methods workers actually need. This reduces the surface area for copy-on-write memory duplication, since workers never touch the hundreds of other attributes on BAMProfiler.

5) GC freeze before fork

  gc.freeze() is called before forking worker processes. This moves all existing Python objects into a permanent GC generation that child processes will never traverse, preventing the garbage collector from incrementing reference counts on inherited objects and triggering copy-on-write page duplication.

6) Safety improvements:
  - SQL injection fix: parameterized queries for split name lookups from BAM-derived contig names
  - Shared memory cleanup wrapped in try/finally to prevent leaks on errors
  - gc.freeze()/gc.unfreeze() wrapped in try/finally
  - Timeout on subprocess communication to prevent infinite hangs
  - Resource tracker unregistration to prevent premature shared memory cleanup
  
# Key changes - explained simply
If, like me, you are not a python expert and the above points were not clear, here is the simpler version that helped me understand what happened.

The core problem: when anvi-profile runs with multiple threads, it actually creates separate processes using fork(). Each process gets its own copy of the parent's memory. Linux is supposed to be smart about this, it shares the same physical memory pages between parent and children, and only makes a real copy when someone writes to a page. This is called "copy-on-write" (COW).

But Python ruins this. Every time Python looks at an object (even just to read it) it updates a counter on that object ("reference counting"). That tiny write triggers a real copy of the entire memory  page. So with 128 workers, each one touching the same inherited data, you end up with 128 real copies of data that nobody is actually modifying. That's how 24GB becomes 3+ TB.

1) Pre-filtering is straightforward: don't load data for contigs with zero coverage. The BAM index already knows which contigs have reads, so we ask it first.

2) Shared memory is a way to share data between processes that Python's reference counting can't touch. Instead of each worker having a Python dict with sequences, we pack the raw bytes into a block of memory managed by the OS (/dev/shm on Linux). Workers read from it through a thin wrapper. Since it's not a Python object, there's no reference count to update, so no copy-on-write.

3) Subprocess loading solves a subtle problem with how the C memory allocator works. When you allocate memory, then free it, the freed space doesn't always go back to the OS — it stays reserved in your process's "heap." It's like a parking garage where you can only demolish floors from the top: if floor 3 is empty but floor 4 is occupied, you can't reclaim floor 3. By loading the 24GB of data in a separate subprocess, it gets its own parking garage. When that subprocess exits, the OS tears down the entire garage. The main process's heap never grew in the first place.

4) WorkerContext is about giving workers a small lunchbox instead of the whole kitchen. The BAMProfiler object has hundreds of attributes (database connections, progress bars, configuration, etc.), but workers only need about 22 of them. By copying just those 22 into a small object, workers don't accidentally touch (and therefore copy) pages belonging to things they don't need.

5) GC freeze tackles another source of accidental page copying. Python's garbage collector periodically walks through all objects to find cycles. In child processes, this walk touches objects inherited from the parent, triggering reference count updates and copy-on-write. gc.freeze() tells Python: "these objects are permanent, don't walk them." After the workers have started, we unfreeze.

@meren, it is working as far as my test cases (profiles in self test + couple of example on the HPC) but I would really appreciate if another "real" human could review the code.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2973,7 +2973,7 @@ D = {
             {'default': 5000,
              'metavar': 'INT',
              'required': False,
-             'help': "How many contigs should be kept in memory before they are written to the disk. The default is "
+             'help': "How many splits should be kept in memory before they are written to the disk. The default is "
                      "%(default)d. The larger the buffer size, the less frequently the program will access the disk, yet the more memory "
                      "will be consumed since the processed items will be cleared off the memory only after they are written "
                      "to the disk. The default buffer size will likely work for most cases. Please keep an eye on the memory "
--- a/anvio/profiler.py
+++ b/anvio/profiler.py
@@ -806,6 +806,7 @@ class BAMProfiler(dbops.ContigsSuperclass):
         self.num_threads = int(A('num_threads') or 1)
         self.queue_size = int(A('queue_size') if A('queue_size') is not None else 0)
         self.write_buffer_size = int(A('write_buffer_size_per_thread') if A('write_buffer_size_per_thread') is not None else 5000)
+        self.buffer_splits = 0
         self.total_length_of_all_contigs = 0
         self.total_coverage_values_for_all_contigs = 0
         self.total_reads_kept = 0
@@ -1854,7 +1855,9 @@ class BAMProfiler(dbops.ContigsSuperclass):
             # memory. So here we are accessing to the atomic data structures in our split
             # objects to try to relieve the memory by encouraging the garbage collector to
             # realize what's up. Afterwards, we explicitly call the garbage collector
-            if self.write_buffer_size > 0 and len(self.contigs) % self.write_buffer_size == 0:
+            self.buffer_splits += len(contig.splits) if contig else 0
+
+            if self.write_buffer_size > 0 and self.buffer_splits >= self.write_buffer_size:
                 self.progress.update(f"{received_contigs}/{self.num_contigs} contigs ⚙ | WRITING TO DB 💾 ...")
                 self.store_contigs_buffer()
                 for c in self.contigs:
@@ -1866,6 +1869,7 @@ class BAMProfiler(dbops.ContigsSuperclass):
                     del c.coverage
                     del c
                 del self.contigs[:]
+                self.buffer_splits = 0
                 gc.collect()
 
         self.progress.update(f"{received_contigs}/{self.num_contigs} contigs ⚙ | WRITING TO DB 💾 ...")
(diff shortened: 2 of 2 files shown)


meren:
Students are working on an exercise, so I read through the changes and skimmed through the code :)

It generally looks good to me. This level of granularity was certainly required at this point, and seeing it make me very happy. It also makes me upset that it wasn't there at the beginning, but I remind myself that at the very beginning it wasn't clear where the profiler was going, and without that long-term vision and understanding, it is impossible to identify the boundaries around independent processes. Thank you for working on this :)

This is such a big change, I feel like we need to test it on more things. But I will let you decide how much testing is enough testing :) That said, it would be nice to go through all other anvi'o programs and libraries behind them to see if there are instances where they use the profile module, and whether we have an issue with API backwards compatibility in those cases. Based on what I see, there shouldn't be, but I am not 100% sure. Maybe you can take a look?

One question I have is why `--skip-contigs-without-coverage` is not default behavior? Why do we even need a flag for that?

PR #2543, opened 2026-02-17
Profiler shared memory

FlorianTrigodet:
This PR add on #2542
After the filtering for contigs with coverage, we tackle the large memory improvement by using shared memory for workers.

# Summary
When anvi-profile runs with -T >1, each worker process previously received a copy of the entire BAMProfiler object via fork(). Python's reference counting triggers copy-on-write page duplication on every attribute access, so even read-only data like contig_sequences and nt_positions_info gets physically duplicated per worker, multiplying resident memory by the thread count.

This PR eliminates that duplication by:

  - Packing large read-only data into OS-level shared memory (multiprocessing.shared_memory.SharedMemory), which exists once in physical RAM regardless of worker count. Three are: contig sequences, nucleotide position info, and per-contig metadata (splits, genes).
  - Replacing the full BAMProfiler with a lightweight WorkerContext in workers. WorkerContext binds only the methods workers need (process_contig, populate_gene_info_for_splits, etc.) and carries ~22 scalar attributes instead of the full object graph.
  - Loading data in a subprocess before fork, so the parent's heap stays clean. The subprocess creates a fresh ContigsSuperclass, loads everything, packs it into shared memory, and exits. Its allocations never touch the parent's address space.
  - Clearing LazyProperties from the parent before fork (splits_basic_info, contig_name_to_genes, genes_in_contigs_dict, contigs_basic_info), reducing the set of Python objects that would trigger COW in workers. These are saved and restored after the workers finish.

Also:

  - contig_sequences loading is deferred from __init__ to the profiling methods. Single-thread loads lazily at the start of profile_single_thread(); multi-thread loads via the shared memory subprocess.
  - Write buffer flushing is changed from a contig count (write_buffer_size_per_thread * num_threads) to a flat split count (5000 splits), which gives more predictable memory behavior since contigs vary widely in how many splits they produce.

### DIFF
--- a/anvio/cli/compute_functional_enrichment_across_genomes.py
+++ b/anvio/cli/compute_functional_enrichment_across_genomes.py
@@ -84,6 +84,7 @@ def get_args():
 
     groupE = parser.add_argument_group('OPTIONAL THINGIES', "If you want it, here it is, come and get it.")
     groupE.add_argument(*anvio.A('qlambda'), **anvio.K('qlambda'))
+    groupE.add_argument(*anvio.A('skip-checking-genome-hashes'), **anvio.K('skip-checking-genome-hashes'))
     groupE.add_argument(*anvio.A('just-do-it'), **anvio.K('just-do-it'))
 
     return parser.get_args(parser)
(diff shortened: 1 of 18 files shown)


meren:
Big change. Are the SNVs tables and coverage data identical between master and this branch? Based on my reading of the code there is absolutely no reason for them to be different, but nothing beats an actual comparison.

PR #2544, opened 2026-02-17
Update variability-profile-txt.md

mankeldy:
`--profile-SCVs` in anvi-merge is not a valid parameter so this was just a little confusing :) Now it's similar to the description for SCVs


meren:
I'm sorry -- I just realized this change has to be done in the anvi'o repository :) This is the file that needs to be changed since this is the source file from outputs are rendered:

https://github.com/merenlab/anvio/blob/master/anvio/docs/artifacts/variability-profile-txt.md

PR #2553, opened 2026-03-09
Revamp anvi-run-workflow for 'contigs'

meren:
This PR revamps the anvi'o snakemake workflow for contigs (the workflow that generates anvi'o contigs-db files from input FASTA files at scale).

This PR does two critical things:

1. It removes unnecessary decompression of FASTA files
2. Removes -contigs.db postfix from contigs-db files, and moves .done files into a separate directory to remove the clutter from `02_CONTIGS` directory in the default workflow output

Quick notes on the implications of these changes:

* The first change requires a version bump for the config.json files, and **the PR includes a migration script for `anvi-migrate`. So no problem there**.
* The second change, however, creates a **backwards compatibility issue** since the new directory structure and file naming patterns will make older directories look incomplete as far as snakemake workflows are concerned. This is not something we can fix with anvi'o code, but we can with a one-time additional migration script outside of anvi'o.  **The PR includes a Python script below that will make the existing `02_CONTIGS` directories compatible with the code. Please see the Migration section at the end of this document for this**.

The details for each item is below.

## Remove unnecessary gunzip step from contigs workflow for gzipped FASTA input

The purpose of this change is to take advantage of 97aec3ac266748253c1a6e54936c4d56267bce63, which made anvi'o great again with compressed FASTA files, and make sure our snakemake workflows are not trying to decompress FASTA files first.

Indeed, the contigs workflow has been decompressing gzipped FASTA files into temporary uncompressed files before passing them to downstream tools, even though those tools already support gzipped input natively.

Prior to this PR, when a .gz FASTA file was provided in a `fasta-txt` file, `get_raw_fasta()` in the contigs workflow redirected to a temp .fa path, which triggered the Snakemake rule called `gunzip_fasta`, through which every gzipped FASTA had to be fully decompressed to disk before any processing could begin. 

This effectively influenced two parts of the code:

  - When `anvi_script_reformat_fasta` was enabled the .gz file was unzipped before being passed to `anvi-script-reformat-fasta`
  - When `anvi_script_reformat_fasta` was disabled the .gz file was unzipped before being passed to `anvi-gen-contigs-database`

This PR changes that wrong attitude and,

  - Simplifies get_raw_fasta() and renames it to `get_input_fasta_path()` (since this is what it does now). The `remove_gz_suffix` parameter is still in there for API compatibility with `MetagenomicsWorkflow`, which overrides and calls this method, but the parameter has no impact anymore.
  - Removes the `gunzip_fasta` rule from `anvio/workflows/contigs/Snakefile `and from the rules list in `__init__.py`.
  - Bumps `workflow_config_version `from 4 to 5 in `anvio/version.py`, since existing config files with a `gunzip_fasta` section would otherwise fail validation.
  - Ads `anvio/migrations/config/v4_to_v5.py` to automatically remove the `gunzip_fasta` key from old config files when users run `anvi-migrate`.
  - Updates all test workflow config files so they have the right rule set and versoin numbers.

## Remove `-contigs` addition to contigs-db files

I always thought -contigs.db was an unnecessary addition to our contigs-db files in anvi'o workflows. They later propagate names we have to manually correct in pangenomes and everywhere else.

This PR includes changes with c1ae56885f10b7548895ee037f959ead6755f630 that remove `-contigs.db` postfix from contigs-db files. The biggest implications of this change is that the exiting workflow output directories will be incompatible with the new code if/once this PR is merged to `master` since snakemake will not be able to find the existing contigs-db as the code will no longer be looking for SAMPLE-contigs.db, but for SAMPLE.db.

**The migration script will take care of this.**

## Move `.done` files into a separate directory

Anvi'o snakemake workflows generate all these files to mark that certain steps are done:

```
  02_CONTIGS/
  ├── anvi_run_hmms-apple.done
  ├── anvi_run_hmms-banana.done
  ├── anvi_run_hmms-orange.done
  ├── anvi_run_kegg_kofams-apple.done
  ├── anvi_run_kegg_kofams-banana.done
  ├── anvi_run_kegg_kofams-orange.done
  ├── anvi_run_ncbi_cogs-apple.done
  ├── anvi_run_ncbi_cogs-banana.done
  ├── anvi_run_ncbi_cogs-orange.done
  ├── anvi_run_scg_taxonomy-apple.done
  ├── anvi_run_scg_taxonomy-banana.done
  ├── anvi_run_scg_taxonomy-orange.done
  ├── apple-annotate_contigs_database.done
  ├── apple-contigs.db
  ├── banana-annotate_contigs_database.done
  ├── banana-contigs.db
  ├── orange-annotate_contigs_database.done
  └── orange-contigs.db
```

Which is really annoying since the clutter is real.

The PR introduces changes to the code with fea805937d1573d9a61bdbfd325b2d7ce52d659e that results in a directory structure like this:

```
  02_CONTIGS/
  ├── apple.db
  ├── apple-steps
  │   ├── anvi_run_hmms.done
  │   ├── anvi_run_kegg_kofams.done
  │   ├── anvi_run_ncbi_cogs.done
  │   ├── anvi_run_scg_taxonomy.done
  │   └── annotate_contigs_database.done
  ├── banana.db
  ├── banana-steps
  │   ├── anvi_run_hmms.done
  │   ├── anvi_run_kegg_kofams.done
  │   ├── anvi_run_ncbi_cogs.done
  │   ├── anvi_run_scg_taxonomy.done
  │   └── annotate_contigs_database.done
  ├── orange.db
  └── orange-steps
      ├── anvi_run_hmms.done
      ├── anvi_run_kegg_kofams.done
      ├── anvi_run_ncbi_cogs.done
      ├── anvi_run_scg_taxonomy.done
      └── annotate_contigs_database.done
```

**The migration script will take care of this.**

## Migration

Here is a script written in Python that will make existing `02_CONTIGS` directories compatible with the new code so the anvi'o snakemake workflow for contigs will still recognize them as 'done':

*  [migrate_contigs_workflow_dir.py](https://github.com/user-attachments/files/25865635/migrate_contigs_workflow_dir.py)

After downloading this script, you can run it the following way:

```
python migrate_contigs_workflow_dir.py /PATH/TO/02_CONTIG
```

It will perform a dry run, and if everything looks OK, will do the moving of files.

---

Here is an older directory that looked like this:

```
tree 02_CONTIGS/
02_CONTIGS/
├── anvi_run_hmms-test.done
├── anvi_run_kegg_kofams-test.done
├── anvi_run_ncbi_cogs-test.done
├── anvi_run_scg_taxonomy-test.done
├── test-annotate_contigs_database.done
└── test-contigs.db
```

On which I run the migration script:

```
python migrate_contigs_workflow_dir.py 02_CONTIG
```

And this is how it looked after:

```
02_CONTIGS/
├── test-steps
│   ├── annotate_contigs_database.done
│   ├── anvi_run_hmms.done
│   ├── anvi_run_kegg_kofams.done
│   ├── anvi_run_ncbi_cogs.done
│   └── anvi_run_scg_taxonomy.done
└── test.db
```

Then, when I ran `anvi-run-workflow` again on this directory, it was able to recognize that everything is already in place:

```
anvi-run-workflow -w contigs -c test.json

(...)

Building DAG of jobs...
Nothing to be done (all requested files are present and up to date).
```

---

I think this is good to go, but who can ever know what the future holds eh?

### DIFF
--- a/anvio/docs/workflows/ecophylo.md
+++ b/anvio/docs/workflows/ecophylo.md
@@ -373,7 +373,7 @@ To visualize the output of [profile-mode](#profile-mode-insights-into-the-ecolog
 ```bash
 PROTEIN="" # Replace with name of protein from hmm_list.txt
 anvi-interactive -p METAGENOMICS_WORKFLOW/06_MERGED/"${PROTEIN}"/PROFILE.db \
-                 -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}"-contigs.db \
+                 -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}".db \
                  --manual
 ```
 
@@ -441,12 +441,12 @@ grep -v -f SUBSET_TREE/bad_branches_headers.txt SUBSET_TREE/collection-DEFAULT.t
 
 anvi-import-collection SUBSET_TREE/my_bins.txt -C curated \
                                                -p METAGENOMICS_WORKFLOW/06_MERGED/"${PROTEIN}"/PROFILE.db \
-                                               -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}"-contigs.db
+                                               -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}".db
                         
 anvi-split -C curated \
            --bin-id EVERYTHING_curated \
            -p METAGENOMICS_WORKFLOW/06_MERGED/"${PROTEIN}"/PROFILE.db \
-           -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}"-contigs.db \
+           -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}".db \
            --output-dir SUBSET_TREE/"${PROTEIN}"_curated
 ```
 
@@ -575,7 +575,7 @@ To visualize the results of the different ecophylo runs, just change the paths t
 
 ```bash
 PROTEIN="" # Replace with name of protein from hmm_list.txt
-anvi-interactive -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}"-contigs.db -p METAGENOMICS_WORKFLOW/06_MERGED/"${PROTEIN}"/PROFILE.db
+anvi-interactive -c METAGENOMICS_WORKFLOW/03_CONTIGS/"${PROTEIN}".db -p METAGENOMICS_WORKFLOW/06_MERGED/"${PROTEIN}"/PROFILE.db
 ```
 
 However, if you are interested in comparing the outputs of different parameters on the same protein, make a new %(workflow-config)s file, otherwise ecophylo will try and fail to overwrite your original run.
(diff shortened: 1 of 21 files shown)


ivagljiva:
One thing I thought about (and am sure you will test it going forward) -- will all rules in the metagenomics workflow be compatible with these changes? For instance, I remember from last week that the rule for removing reads that map to a reference required unzipped files

PR #2563, opened 2026-04-12
Adding a Distribution of Coverage (DisCov) statistic to anvi'o profiling

ivagljiva:
This PR implements a new profiling statistic in anvi'o: the distribution of coverage score, or DisCov for short. DisCov is designed to capture the presence-absence of a contig or genome in a metagenomic sample and is an alternative to using arbitrary detection thresholds for this purpose.

DisCov incorporates two independent metrics:
- we measure the _spread (S)_ of coverage across an input sequence by dividing that sequence into non-overlapping windows and computing the proportion of those windows that include bases with nonzero coverage. In other words, `S = # of windows with coverage / # of windows`
- we measure the _evenness (E) of coverage depth_, at all bases that have nonzero coverage, by computing the proportion of these bases at which the coverage falls within some fold-range of the median nonzero coverage. In other words, `E = # covered bases that fall within some range of the median / # of covered bases`

These two metrics are combined linearly with some weight _α_ to compute the DisCov score, using the formula `DisCov = αS + (1-α)E`.

DisCov is currently implemented as part of `anvi-profile-blitz`, and users have the option to modify pretty much all aspects of the calculation using the following parameters:

```
DISCOV SCORE:
  Parameters related to computing the distribution of coverage score: DisCov = αS + (1-α)E, where S =
  proportion of windows with coverage and E = proportion of covered bases within a fold-range of the median
  nonzero coverage. Does not apply to --gene-mode.

  --window-length INTEGER
                        How long to make the windows for computing the spread metric: S = # windows with
                        coverage / # windows. This is the default when no window length parameters are provided.
                        For contig-level stats, the default window size is 500, and for genome/bin-level stats
                        the default window size is 10000.
  --window-length-as-percentage INTEGER
                        With this option you can set the window length for the spread metric (S) dynamically as
                        a percentage of a given input sequence length. This works well when your input sequences
                        (contigs or genomes) have a wide size distribution and you don't have a
                        one-size-fits-all window length to use. For instance, if you want the window size to be
                        5% of the contig length, you would use `--window-length-as-percentage 5`. You may want
                        to consider also specifying a reasonable minimum window length with the
                        --min-window-length parameter.
  --min-window-length INTEGER
                        Use with --window-length-as-percentage to ensure that percentage-based window lengths
                        never fall below this value.
  --foldrange-lower FLOAT
                        When computing evenness of coverage depth (E), count any bases with coverage OVER this
                        value * the median nonzero coverage. In typical use-cases, this value will be the
                        inverse of --foldrange-upper. The default is '0.25'.
  --foldrange-upper FLOAT
                        When computing evenness of coverage depth (E), count any bases with coverage UNDER this
                        value * the median nonzero coverage. In typical use-cases, this value will be the
                        inverse of --foldrange-lower. The default is '4.0'.
  --alpha FLOAT         How much to weight S over E in the DisCov score. Should be a value in the range [0,1].
                        The default is '0.5'.
```

Note that if the user provides none of the DisCov parameters, we currently have the following set of defaults:
- for computing _S_: using `--window-length` with a default value of 500bp for contigs or 10000bp for genomes (collection mode)
- for computing _E_: using the fold-range with a factor of 4, ie counting bases with 0.25x to 4x of the median nonzero coverage
- for computing _DisCov_: using an alpha value of 0.5 (equal weight on _S_ and _E_)

I set these defaults based on what worked best after testing a few different sets of parameter values using 24 marine metagenomes each with 4-5 manually-confirmed 'present' genomes and 4-5 manually-confirmed 'absent' genomes. However, I do plan to do some more thorough hyperparameter tuning to identify which parameters are the best. I suspect that the fold-range for _E_ should be narrower when computing DisCov at the contig level.

`anvi-profile-blitz` reports not only the _DisCov_ score for each contig/bin, but also _E_, _S_, and the number of windows used for computing _S_:

|**`bin`**|**`sample`**|**`(...)`**|**`num_windows`**|**`prop_windows_covered`**|**`prop_cov_within_foldrange`**|**`dis_cov`**|
|:--|:--|:--|:--|:--|:--|:--|
|TARA_SAMEA2620403_MAG_00000042|SAMEA2620959_subsample_0.1_sorted|(...)|20|1.0|1.0|1.0|
|GIFF20_1_SAMN15311063_MAG_00000016|SAMEA2620959_subsample_0.1_sorted|(...)|20|1.0|1.0|1.0|
|TARA_SAMEA2621272_MAG_00000112|SAMEA2620959_subsample_0.1_sorted|(...)|20|1.0|1.0|1.0|
|RSGB23_1_GCF_014280075_V1_GENO_10000001|SAMEA2620959_subsample_0.1_sorted|(...)|20|1.0|1.0|1.0|

DisCov is not computed when using `--report-minimal` or `--gene-mode`.

This PR is not ready to be merged as some tasks remain:
- [x] hyperparameter tuning to set default parameter values (potentially distinct defaults for contig-level vs bin-level stats)
- [x] add a flag to generate a separate output file describing per-window information: window start, window stop, window length, whether or not window has any coverage (ie, is counted in numerator of _S_), how many of its bases are within the fold-range of the median nonzero coverage (ie, are counted in numerator of _E_), etc. This file will allow us to make nice plots showing the contribution of different sections of a given sequence to the DisCov score.
- [x] adding genome-level DisCov calculation and output to `anvi-summarize`
- [x] documentation, ideally with some nice schematic describing the calculation
- ❌ ~adding contig-level DisCov scores (and S, E metrics) to the profile database, to be filled in during `anvi-profile` and exported later with downstream programs.~ The calculation needs to be done at the contig level and the resulting values propagated to every split of a given contig to create a new view table for the interface.
- [x] it would be nice to have a sanity check that a specific `--window-length` value doesn't exceed the minimum contig/genome length, so that we can stop the show early instead of failing spontaneously when attempting to computing the window regions for very short sequences


ivagljiva:
## Window-length edge cases for contig-level metrics

Just wanted to document an issue that I ran into while testing DisCov for contig-level metrics. Sometimes contigs are short. This is not a problem when using percentage-based windows. But when using explicit window lengths, what do we do when a contig is shorter than the requested window length (`--window-length`)? 

The solution I came up with was to avoid splitting and instead have the entire contig be a single window. The implication of this is that the spread metric _S_ collapses to a binary value (does the contig have any coverage, or not), and then:
 - If the contig has no coverage (_S_ = 0), DisCov is 0 (as the evenness metric _E_ will also be 0). 
 - If the contig does have coverage (_S_ = 1), then the DisCov value has a minimum value of α and any increase from α is entirely dictated by _E_: `DisCov = α + (1-α)E`

I think this is a reasonable outcome for the edge case. No coverage would yield a DisCov of 0 regardless of input contig length. The _S_ = 1 case can yield artificially inflated DisCov values if there is like 1 read mapping to the contig yielding perfect evenness over an even shorter section of the already short contig, so it breaks down when the requested window length is quite long relative to the short end of the contig length distribution. But I hope that most users would see the logic in using percentage-based windows, or very short explicit lengths, when working with contigs.

A potential alternative to this would be to simply give up on these short contigs and report 'NA' values instead, but I would strongly prefer if DisCov, like other coverage metrics, was a real number for all input sequences. So I implemented the solution above, and I let the user know how many contigs are impacted with a warning like this one:
```
WARNING
483 out of 1009 contigs (0.48%) are smaller than the requested window size for
the DisCov metric. These contigs will not be split into multiple windows when
computing DisCov. Just so you know. One of the culprits (in case you want to
check): YOSH21_1_SAMD00070342_MAG_00000001_000000000043
```

I might also add a little error that stops the show if the percentage of affected contigs is over 50%, requiring users to use `--just-do-it` if they _really_ want it. But perhaps it is just better to have sensible defaults: very short (currently ~500bp) window sizes, OR making percentage-based windows the default window sizing strategy for contig-level metrics.

However. There is yet another edge case that appears. Rarely are the input sequence lengths perfectly divisible by the window length (even when using percentage based windows, since there is rounding via integer division), and this means that sometimes the final window (whose length will be the remainder of the integer division step) is very short compared to the rest. In normal circumstances (longer contigs split into more than 1 window), I don't want very small 'remainder windows' to have the same influence on the calculation as the other windows. So I have this little check that simply throws the last window out when it is less than 10% of the requested window length:
```python
final_window_len = windows[-1][1] - windows[-1][0]
if final_window_len < 0.1*discov_window_length: # don't count the last window if it is too small
                windows = windows[:-1]
```
This breaks down if the input sequence is so small compared to the requested window length that a) the contig is a single window and b) the contig length is <10% of the requested window length. In these situations, we don't throw anything out so that we can still have 1 window for our calculations. It is not ideal, but that is as far as I am willing to take this for an edge case of the edge case that should not dictate what scientists want to do with their data :p

ivagljiva:
## Insights from Hyperparameter Tuning 

I tested many combinations of the DisCov input parameters to see how we should set the default values. This is detailed more extensively elsewhere (Meren Lab members can view it [here](https://github.com/merenlab/lab-fiesta/issues/2392)) but for full transparency I also wanted to document the outcomes here where they are publicly viewable.

### Test data
I used a set of 24 deeply sequenced marine metagenomes, in each metagenome I identified 4 to 5 'present' genomes and 4 to 5 'absent' genomes (with manual verification in `anvi-interactive`), and then I randomly subsampled the reads to simulate a wide range of lower sequencing depths. Mapping back to these genomes from the original metagenome and its subsampled versions yielded BAM files that I could then run `anvi-profile-blitz` on to test how well DisCov works for identifying present vs absent across different levels of sequencing depth.

### Parameter ranges tested

For genome-level metrics (using `anvi-profile-blitz` with a collection `-C`):
- window length:
  - either `--window-length` with options 500, 1000, 5000, 10000, 25000, 50000, 100000
  - OR `--window-length-as-percentage` with options 1, 5, 10 and `--min-window-length` with options 300 500 1000
- foldrange upper bound `--foldrange-upper` with options u = 2, 3, 4, 5 and corresponding `--foldrange-lower` set to 1 / u
- `--alpha` value with options 0.2, 0.4, 0.5, 0.6, 0.8

This was a total of 320 genome-level combinations. 

For contig-level metrics (normal use of `anvi-profile-blitz`):
- window length:
  - either `--window-length` with options 500, 1000, 5000, 10000, 25000, 50000
  - OR `--window-length-as-percentage` with options 1, 5, 10, 15, 20 and `--min-window-length` with options 300 500 1000
- foldrange upper bound: `--foldrange-upper` with options u = 2, 3, 4, 5 and corresponding `--foldrange-lower` set to 1 / u
- `--alpha` value with options 0.2, 0.4, 0.5, 0.6, 0.8

This was a total of 420 contig-level combinations (more percentage-based window options and less absolute length options than for genome-level). 

### Tuning workflow
For each parameter combo, I ran `anvi-profile-blitz` on each of the 24 samples (BAM files from all subsampling levels + the original). I combined the sample-specific output into one big table, and I did the following evaluations:
- Spearman's correlation of DisCov value with the binary presence-absence label. report the rho value. Note that correlation is not necessarily the best method here for Discov (a continuous value in range [0,1]) vs the binary label, but I used it to evaluate whether (and how consistently) 'present' genomes or contigs had higher DisCov scores than 'absent' ones.
- test multiple thresholds of the DisCov value to separate 'present' from 'absent'. for each threshold, compute the accuracy of classification. find the best threshold (highest accuracy)
- for the best threshold, report the accuracy and F1 score.

This was done separately for testing genome-level metrics and contig-level metrics.

For contigs, I couldn't manually assign presence-absence labels so I had to simply propagate the genome label to all contigs in the genome. Not perfect (there will be some mislabeled contigs due to nonspecific read recruitment, or genome contamination, or natural small-scale variation of environmental populations, etc), but for the most part the contig labels should match that of the overall genome, and that will hopefully be good enough for assessing performance. 

I used the final report table to evaluate the impact of changing parameters, the relationships that different parameters have with each other, and to get a rough idea of how to set the parameters for overall good performance. Below is a summary of my findings at the genome level, and at the contig level:

### Genome-level

Just to help people understand a bit more the test data and the whole 'threshold' business I described above, here are some example plots of subsampling level (x-axis) vs DisCov value (y-axis) for genomes in 6 of the test samples. This is an example of a very good parameter combo:

image

And this is an example of a not-so-good parameter combo:
image

The 'best threshold' would be the horizontal line that most accurately splits present genomes from absent genomes  at all subsampling levels in my tests. Note that the 'maybe' genomes (purple lines) weren't considered in the performance evaluation.

For my test genomes (which are admittedly an ideal case), performance is pretty good regardless of parameter choices. Spearman's rho fluctuated between 0.78 and 0.87 (always strongly positive, so 'present' genomes tend to have higher DisCov scores) and the reported 'best' accuracy was >= 87% for all parameter combinations. That said, a few guidelines emerged based on the patterns I saw in parameter combinations that had great performance:

- for whole genomes, use an explicit window size of <=50kb rather than a percentage-based strategy. More specifically, setting  `--window-length` to <=10kb has worked the best so far.
- set the fold-range "factor" to be within 2 to 4. Most of the parameter combinations with >98% accuracy used a factor of 2 (so 0.5x-2x, or `--foldrange-lower 0.5` and `--foldrange-upper 2`).
- set alpha between 0.4 and 0.6 (ie, roughly equal weight on spread and evenness in the DisCov calculation).
- If you want to use percentage-based windows for genome-level metrics, you should use a very small percentage value to ensure small enough windows. 
   - The smallest percentage I tested was 1% and it worked much better than 5 or 10%, but smaller might be even better (especially with longer genomes) -- consider that 1% of a 3Mb genome is 30kb, which is already much longer than a lot of the best-performing `--window-length` values I tried.
  - when using very long windows (only observed in my percentage-based tests), fold-range factor and alpha both need to be higher for good performance


Based on these results, I set the following default parameter values for genome-level DisCov metrics:
- `--window-length 1000`
- fold-range factor of 2 (`--foldrange-lower 0.5` and `--foldrange-upper 2`)
- `--alpha 0.5`

### Contig-level

For contigs, performance was once again generally good, but not as good as when we look at full genomes: accuracy fluctuated in the narrow range of 0.893 - 0.906. Spearman's rho was still strongly positive (0.75 - 0.80).

I think the lower accuracy is at least partially due to the labeling scheme -- contigs were labeled as 'present' if they came from a 'present' genome, or absent if they came from an 'absent' genome, ignoring the actual read recruitment evidence. That means that a small portion of labels are likely incorrect.

Regardless, a few patterns still emerged from the higher accuracy parameter combinations:
- using very small windows works best and this can be achieved using either window sizing strategy -- the most accurate parameter combos tended to use `--window-length 500` (500bp) or `--window-length-as-percentage 1` for a 1% window. For the latter option, the minimum window lengths didn't have much influence on performance, but 300 bp appeared more often in the most accurate combos.
- like with the genome-level metrics, using a fold-range factor of 2 (so 0.5x-2x) tended to work the best most consistently
- an alpha value of 0.4 or 0.5 also works the best most consistently (regardless of window sizing strategy)

Based on these results, I think we can safely use the same default parameter values for alpha and fold-range as the genome-level metrics. The window length needs to be shorter, though -- and I think the safest option is to go with the percentage-based strategy by default to avoid the edge case of "short contig = only 1 window" that I mentioned in the previous comment. These contig-level defaults should therefore be:

- `--window-length-as-percentage 1`
- `--min-window-length 300`

ivagljiva:
## Genome-level DisCov scores when using default parameters 

I thought I'd show some data from the hyperparameter tuning that reveals a few guidelines on how DisCov scores for genomes relate to their presence or absence, at least when using the default parameters for genome-level stats (window length of 1000bp, fold-range coverage depth between 0.5x and 2x of the median nonzero coverage, and alpha = 0.5). I plotted the DisCov scores from the `anvi-profile-blitz` output for this particular parameter combo in two ways.

First, here are the per-sample line plots like I showed in the previous comment. Each genome has its own line of scores across the different subsampling levels (from 0.1% of reads to 100% of reads) in a given sample, so that you can see how the score changes with sequencing depth.
dis_cov_part1
dis_cov_part2
dis_cov_part3
dis_cov_part4

Second, here is the distribution of all scores across all samples, split by genome label. Each genome-sample pair is represented by multiple points in this plot (once per subsampling level), with lighter points indicating lower subsampling levels (<1% of reads kept).
discov_violins

Based on the above plots, a few common patterns emerge:
- for present genomes, DisCov scores are always > 0.5 and increase as sequencing depth increases (here simulated by larger number of subsampled reads)
- for absent genomes, DisCov scores are mostly < 0.5 and decrease or stay flat as sequencing depth increases
- there is a bit of overlap in the present and absent score distributions around 0.5, particularly in low sequencing depth samples.
- for this particular set of test data, a DisCov score >= 0.6 indicates that a genome is present without any false positives, but with some false negatives particularly in low sequencing depth samples.

Note that the 'maybe' genomes (purple) were originally in the 'absent' group, but I changed their label after manually visualizing the coverage plots because I wasn't sure that they were truly absent. There are only 8 'maybe' genomes. However, most of these 'maybe' genomes clearly show the same DisCov score pattern as other absent genomes. Only one in sample `SAMEA4397091` and one in sample `SAMEA2622677` actually have an increasing DisCov score as the number of sequencing reads increases, starting from ~2% of reads or 20% of reads, respectively -- I suspect that only these genomes are potentially present at low abundance in their respective samples, while the other 'maybe' genomes are truly absent. Based on how clear the signal is here, DisCov seems to have better discriminating power than my own eyes.

## What contributes to higher DisCov scores for absent genomes at lower sequencing depths?

It seems to be the coverage evenness score (E in the DisCov equation). I plotted the distributions of S and E separately with violin plots, and while S is always low for absent genomes, E can be quite high particularly when we keep less than 1% of sequencing reads (the fainter points are quite prominent at higher E values):

image

image

This suggests that increasing the alpha value (to put more weight on S) might help discriminate between present and absent genomes for samples when sequencing depth is low.

How low is low? The 100% read count for the test samples used here ranges from ~882,000 reads to ~195,000,000 reads, so a 1% subsample is typically on the order of tens of thousands of reads or less. Note that those numbers are not the original size of the sample, but rather count only the reads that map to the ~10 test genomes per sample -- if you want to see the original number of reads you can check their accessions on NCBI SRA.

## How to use DisCov to decide on presence or absence of genomes?

These test genomes were cherry picked because they mostly had unambigious coverage patterns allowing us to label them as truly present or absent. "Real" data will include ambiguous cases and will not be as clean. That said, I think we can still identify a few guidelines on how to use the DisCov score from the insights above:
- if DisCov <= 0.5, the genome is likely absent. Using the default alpha of 0.5, a DisCov value of 0.5 means that either coverage spread (S) or evenness (E), or both, is low. 
- DisCov scores just above 0.5 seem to be ambiguous, since genomes of all labels in the test data above could have DisCov scores around this value -- particularly when the sequencing depth is low.
- higher DisCov scores (let's say >= ~0.7) indicate that both spread (S) and evenness (E) are at least somewhat high with the default alpha of 0.5. To identify present genomes, people can choose a threshold that they are comfortable with: making the threshold higher if they want to be conservative to avoid false positives, or lower (closer to 0.5) if they want to identify more 'present' genomes at the risk of including false positives.
  - it is also possible to use a more complex strategy by using the individual S and E values; for instance, if you trust the spread metric more, you could identify presence by focusing on genomes with both middle-to-high DisCov values and high spread (ie, ignoring genomes for which evenness is high but spread is not)

ivagljiva:
I recently added a geometric version of DisCov and thought I should explain why here:

## When few reads map, Evenness value is often high and linear DisCov can be > 0.5

In my test data, I recently ran into some cases where a very small number of reads (< 7) mapped to a genome. The genomes' spread (S) values were very small but the evenness (E) values were 1.0 (probably each read was isolated, yielding 1x coverage at all bases with nonzero coverage). Because linear DisCov weights these values evenly (with the default alpha of 0.5), these genomes received DisCov scores of around or slightly greater than 0.5 despite the limited signal.

I felt that was rather unintuitive, especially given that the S scores were really quite tiny (often < 0.01) and it seemed unreasonable to have a mid-level DisCov score in this scenario (especially given that you could achieve a similar DisCov score from having middle-low values of both S and E). Of course, one option to resolve this is to simply increase the alpha parameter to weight S more than E -- however, when I tried this with alpha values of 0.6 and 0.8, it only worked well for the 0.6 case because there are many present genomes for which E is high while S is low-to-mid value, so E is actually important for the signal in some cases. Increasing alpha for the linear DisCov helps identify some types of present genomes but makes it harder to identify others.

After considering various solutions (and trying to avoid those that require arbitrary thresholds or filters on sequencing depth, or incorporating info in the DisCov formula that isn't normalized by reference length), I settled on an optional secondary formula for Discov utilizing a weighted geometric mean, which essentially requires both S and E to be high in order to get a high score.

## Geometric DisCov

The geometric version of DisCov has the formula:  DisCov = S^α * E^(1-α)

Here is the distribution of scores for my labeled test genomes (using default params of linear DisCov, including an alpha of 0.5):
image

And with just the lower subsampling levels (my proxy for low sequencing depth):
image

Considered in comparison to the linear DisCov versions of these plots, we can see that the range of scores is lengthened for the present genomes (the lower scores largely occur in the low subsampling levels where much fewer reads are available to map), while the range is somewhat compressed for absent genomes. There is much more overlap between the genome classes than with linear DisCov (we can no longer perfectly separate absent vs present), but the scores scale more sensibly relative to S, E, and sequencing depth.

### Parameters for Geometric DisCov

I tested multiple parameter combinations through hyperparameter tuning like I did before with the linear combination, only for genome-level metrics. The results were overall rather similar to what we had before, except that geometric Discov yields somewhat lower classification accuracies due to the increased overlap between present and absent genome scores at lower subsampling levels. The outcome for each mutable parameter was:
- using absolute length rather than a percentage-based window is better
- using a window length of >= 5000 bp and <= 100000 bp is best (note that this is longer than the default for linear DisCov, which is 1000 bp)
- changing the fold-range factor doesn't affect the outcome much, but 0.5x - 2x (the default for linear) is very slightly better than the rest
- changing the alpha value also doesn't affect the outcome much

Because of this, I elected to maintain the current set of defaults for linear DisCov (which is also the default formula). If people want to use geometric DisCov, they could try setting the window length higher (5000 bp worked the best most consistently across all parameter combos I tried), keeping in mind that once again, the impact of changing parameter values on overall accuracy (when using a global threshold value to separate present vs absent genomes) is small (most combos yield >90% accuracy already).

ivagljiva:
I've decided to skip the implementation of DisCov in `anvi-profile`:

> adding contig-level DisCov scores (and S, E metrics) to the profile database, to be filled in during `anvi-profile` and exported later with downstream programs.~ The calculation needs to be done at the contig level and the resulting values propagated to every split of a given contig to create a new view table for the interface.

DisCov requires some parameterization, and I don't want people to have to make those decisions at the time of profiling, nor do I want to enforce some defaults. I also don't think keeping the DisCov scores in the profile DB is worth it (particularly as they would be generated given a certain set of parameters).

However, it is still possible to include calculation of DisCov in `anvi-summarize`, as long as the user holds onto their auxiliary database (which contains binary blobs of the per-nucleotide coverage values for each split). I am working on this now.

ivagljiva:
## Added DisCov option to `anvi-summarize`

You can now request DisCov to be computed as part of your summaries with the use of the `--report-discov` flag. If the auxiliary database is available, DisCov will be computed both at the contig-level and at the bin-level (although we only compute it for contigs within the collection the user is summarizing, so unbinned contigs will not be considered). If the aux DB is not available, the user gets a warning and a normal summary without the DisCov output.

Note that we have the same default parameter setup as in `anvi-profile-blitz`: if the user doesn't specify any window sizing strategy, we use different strategies for bin-level stats (fixed length, 1000bp) than for contig-level stats (1% with minimum 300bp), _even though_ `anvi-summarize` is computing both (as opposed to `anvi-profile-blitz`, for which the user chooses either a collection focus or a contig focus). Doing this required some not very elegant shenanigans in the code for decision-making based on what flags the user did or did not provide. However, I think it is worth it because (a) I didn't want to enforce the use of fixed-length windows on contigs that could be smaller than 1000bp (the consequence would be that their Spread score turns binary -- not the end of the world but also not that informative), and (b) I didn't want the user to get different contig-level DisCov scores on the same dataset when using `anvi-profile-blitz` vs `anvi-summarize`. Of course, if the user explicitly sets some parameters for window length, `anvi-summarize` will apply those parameters for both bin- and contig-level stats.

The DisCov output files are written to the `bins_across_samples` directory of the summary folder, as `discov_bins.txt` and `discov_contigs.txt`.

|**`bin_name`**|**`sample_name`**|**`length`**|**`num_windows`**|**`prop_windows_covered`**|**`fold_range_coverage_depth`**|**`dis_cov`**|
|:--|:--|:--|:--|:--|:--|:--|
|YOSH21_1_SAMD00070342_MAG_00000001|SAMEA2621487_subsample_50_sorted|772390|773|1.0|0.42771023012028064|0.7138551150601403|
|YOSH21_1_SAMD00070342_MAG_00000001|SAMEA2621487_subsample_90_sorted|772390|773|1.0|0.4250037430753107|0.7125018715376553|
|SCHR19_1_SAMN09405988_MAG_00000009|SAMEA2621487_subsample_50_sorted|781499|782|0.9872122762148338|0.29338695939660875|0.6402996178057212|
|SCHR19_1_SAMN09405988_MAG_00000009|SAMEA2621487_subsample_90_sorted|781499|782|0.9884910485933504|0.2855191351178135|0.637005091855582|

|**`bin_name`**|**`contig_name`**|**`sample_name`**|**`length`**|**`num_windows`**|**`prop_windows_covered`**|**`fold_range_coverage_depth`**|**`dis_cov`**|
|:--|:--|:--|:--|:--|:--|:--|:--|
|YOSH21_1_SAMD00070342_MAG_00000001|YOSH21_1_SAMD00070342_MAG_00000001_000000000001|SAMEA2621487_subsample_50_sorted|30788|101|1.0|0.7185423993747557|0.8592711996873779|
|YOSH21_1_SAMD00070342_MAG_00000001|YOSH21_1_SAMD00070342_MAG_00000001_000000000002|SAMEA2621487_subsample_50_sorted|27639|93|1.0|0.5155065472088215|0.7577532736044108|
|YOSH21_1_SAMD00070342_MAG_00000001|YOSH21_1_SAMD00070342_MAG_00000001_000000000003|SAMEA2621487_subsample_50_sorted|22509|75|1.0|0.242036518725843|0.6210182593629215|
|YOSH21_1_SAMD00070342_MAG_00000001|YOSH21_1_SAMD00070342_MAG_00000001_000000000004|SAMEA2621487_subsample_50_sorted|19705|66|1.0|0.29552967502415706|0.6477648375120786|


I tested this by taking two of my test BAM files, computing DisCov with `anvi-profile-blitz` (default params) for both bin- and contig-level. Then I computed it with `anvi-summarize` and compared the scores to ensure that we got the same values (we did, except that blitz reports only up to the 4th decimal place while anvi-summarize reports the whole number). I also verified that the other anvi-summarize outputs were the same after these changes (they are, though there is also some minimal differences between runs due to rounding error).

ivagljiva:
Documentation is now in the codebase :) I think this is now ready to be merged.

The only remaining task is to add a new program called `anvi-compute-discov` that computes DisCov based on user-provided parameters and adds it to a profile database as a new view table, so it can be visualized in the interface. But that is more appropriate for another PR as it will require a version bump to the profile database and a migration script (to add empty DisCov tables).

PR #2566, opened 2026-04-20
new display subsection in main tab

FlorianTrigodet:
Been changing a couple of things in the interface lately and I though this one deserve a PR and maybe a discussion.
It bothered me for a while that the display type was in the 'Items' subsection of the Main tab. Does not make a lot of sense. 
Also, we often need/want to change basic settings for the display like the radius/angle of the circle phylogram, and the height/width of the phylogram, but they are buried in the options tab.

In this PR, I created a new subsection of the Main tab, called "Display":

image

I left the option for edge normalisation in the options tab as it is more more advanced tunning.

Note: it now automatically displays the values used by default. What I mean is that when switching to phylogram, the height and width was always '0' in the settings, making it unnecessarily difficult to guess what value would give you. NOT ANYMORE.

I also don't find that section particularly beautiful and I am also questioning if it should be close by default, so if anyone reads this PR, let me know how you feel and I'll see what I (lol, more likely Claude) can do.

### DIFF
--- a/anvio/data/interactive/images/display-chart.svg
+++ b/anvio/data/interactive/images/display-chart.svg
@@ -1,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="40"
+   height="40"
+   viewBox="0 0 128 128"
+   fill="none"
+   version="1.1"
+   id="svg13"
+   sodipodi:docname="display-chart.svg"
+   inkscape:version="1.4.3 (0d15f75, 2025-12-25)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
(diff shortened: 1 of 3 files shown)


meren:
This has been bothering me for so long too! :)

srisvs33:
Thank you for this PR @FlorianTrigodet . I am also wondering whether something can be done to keep an option to increase the font size of the names in phylogram, like bin name or species name . for example 
image

The current options were very limited, and doing it manually for larger trees was very difficult. 


Thanks and regards

Venkat

FlorianTrigodet:
Great idea, I'll try to add that feature when I find the time. For now, I will merge this PR if we don't have big complains about the new "Display" subsection.

Issue #2211, opened 2024-02-02
[REPORT] Updating anvi'o SCG taxonomy databases with new genes


meren:
So we recently decided to change the collection of SCGs in anvi'o that are used to infer taxonomy of genomes rapidly. The SCG data are coming from the GTDB, and the purpose and the utility of this functionality is explained [here](https://merenlab.org/2019/10/08/anvio-scg-taxonomy/). The bottom-line is that it helps a lot during binning to have that real time taxonomy estimation: 

![](https://merenlab.org/images/anvio/2019-10-08-anvio-scg-taxonomy/igd-selection.gif)

But then we found other uses for the SCG taxonomy framework, including its use in @mschecht's EcoPhylo workflow. But every good thing leads to more pain, so of course Matt needed a DIFFERENT kinds of SCGs to be used in anvi'o SCG framework, because he found out that not all ribosomal proteins were created equal and their rate of assembly in metagenomes and distributions across MAGs were not identical (as also eloquently covered by Matt Olm and others). So he asked me to change the list of ribosomal proteins we have been using in anvi'o to infer taxonomy from the existing list of things that included these genes:

* Ribosomal_S2
* Ribosomal_S3_C
* Ribosomal_S6
* Ribosomal_S7
* Ribosomal_S8
* Ribosomal_S9
* Ribosomal_S11
* Ribosomal_S20p
* Ribosomal_L1
* Ribosomal_L2
* Ribosomal_L3
* Ribosomal_L4
* Ribosomal_L6
* Ribosomal_L9_C
* Ribosomal_L13
* Ribosomal_L16
* Ribosomal_L17
* Ribosomal_L20
* Ribosomal_L21p
* Ribosomal_L22
* ribosomal_L24
* Ribosomal_L27A

But Matt asked for the removal of the following ones from this list:

* Ribosomal_L6: domain duplication, detected twice in genomic datasets
* Ribosomal_S3_C: C terminus domain only, not full length
* Ribosomal_L9_C: C terminus domain only, not full length
* Ribosomal_S20p: low detection of GTDB bacteria
* ribosomal_L24: low detection of GTDB bacteria

And the inclusion of these:

* Ribosomal_L5: Covers Bacteria GTDB very well
* Ribosomal_L19: well detected in human gut, human oral, GTDB Bacteria
* Ribosomal_S15: well detected in human gut, human oral
* Ribosomal_S16: well detected in human gut
* Ribosomal_S2: well detected in human oral
[...]

PR #2570, opened 2026-04-24
Fix issue link for GH interface

ulgens:
GHA interface renders this link with the dot at the end, which redirects to ".../issues/created_by/2211."

Screenshot 2026-04-24 at 13 45 30 UTC@2x


ulgens:
To test the broken rendering, this is the URL in GHA screen:

`https://github.com/merenlab/anvio/issues/2211.`

The comments interface automatically cleans the URL and renders an issue link so the raw URL is needed for the demonstration of the issue.

A live example can be found under https://github.com/merenlab/anvio/actions/runs/24882731902/job/72854978080, under "Run component tests" section.

FlorianTrigodet:
thanks for catching this!

PR #2571, opened 2026-04-24
Don't ignore "anything starts with a dot"

ulgens:
Originally implemented in https://github.com/merenlab/anvio/commit/a26b1ea3be606f78a1bf8d39747610c10ce01204 but I believe it was forgotten there after adding the extra rules.

Having it in place hides the changes for makes-sense-to-version files/folders like .github/ and .pre-commit-config.yaml


meren:
Absolutely :) We will have to be a bit more careful to expand the list of rules going forward, but this certainly is better practice. Thank you, @ulgens!

PR #2572, opened 2026-04-24
Merge "daily" and "on demand" test workflows

ulgens:
Cleans code duplication.

Originally added in https://github.com/merenlab/anvio/commit/06be73813fdc69fef2fe0934e313acc500c8d1a4

I wasn't able to find the reasoning/explanation for having two workflows instead of one with two triggers. Please let me know if I'm missing anything.


meren:
Haha -- it is because SOMEONE WHO SHALL NOT BE NAMED clearly did not first try to figure out how to include multiple triggers for the same workflow, so they just copy-pasta'd their way out of a frustrating hour of need. THANKS for catching this!

PR #2575, opened 2026-04-25
Remove duplicate checkout step

ulgens:
This was handled in dcd3111f4032dbdb1e1403b2405583a7cdc01273 and reverted in 5dc7d699b10b0f0671584d474cb2d023b85b7cd4. I believe the problem called for the revert was irrelevant to the checkout step but it got reverted because the changes were not separated into multiple commits.


meren:
Precisely.

PR #2576, opened 2026-04-25
General git hooks

ulgens:
Using https://github.com/ulgens/python-anything/blob/162a15b286ea85d7baf9195eeb0fc37f595d589d/.pre-commit-config.yaml as base.

PR #2578, opened 2026-04-25
No more trailing whitespaces

meren:
This PR adds a new rule to `ruff.toml` to prevent trailing whitespaces in future commits, and removes the existing ones from the repository.


ulgens:
@meren Looks like this PR got merged with multiple remaining trailing whitespaces and the CI was failing. Do you have a plan for the rest? If not so, I can push another PR to handle them.

meren:
Oh. I thought I was able to push them in. I didn't realize CI was failing (this was an interesting less than if you are adding a new rule, push the rule *after* fixing the codebase as per the rule).

Any help would be most welcome since I am a bit swamped today and don't want to screw things up any further :p

ulgens:
@meren Got it. I guess some of the fixes were not auto-applied

>   Found 23 errors.
>  No fixes available (23 hidden fixes can be enabled with the `--unsafe-fixes` option).

If #2576 is looking good, I can follow it up with white space stuff.

PR #2580, opened 2026-04-26
Styling whitespaces

ulgens:
Enables & applies whitespace-related lint rules.

I excluded `mock_data_for_structure/` from  `trailing-whitespace` and FASTA files `end-of-file-fixer`. I'm not familiar with those file formats and I didn't want to break anything in case the file formats need those extra or missing whitespaces. Please let me know if this is not needed.


meren:
Hi @ulgens, would you consider excluding the entirety of `anvio/data` and `anvio/tests` directories from whitespace considerations? But it would still be good to keep checks for big files or secrets and private keys and so on. I am not sure if we can be that granular, but I think it would serve us better in the long run to keep these directories as is as far as whitespaces go.

meren:
Thank you, @ulgens!

PR #2581, opened 2026-04-26
Apply "unsafe" W291 fixes

ulgens:
These fixes are not auto applied (missed in #2578) because they not considered "safe": https://docs.astral.sh/ruff/rules/trailing-whitespace/#fix-safety

We can auto apply unsafe fixes by `--unsafe-fixes` argument in hook config but I think the current approach is overall safer.

Relates to:
- #2578

### DIFF
--- a/anvio/data/misc/MODELLER/scripts/binarize_database.py
+++ b/anvio/data/misc/MODELLER/scripts/binarize_database.py
@@ -1,6 +1,6 @@
 """
 Positional arguments:
-1. INPUT - PIR database 
+1. INPUT - PIR database
 2. OUTPUT - BINARY database
 """
 import sys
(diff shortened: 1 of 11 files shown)


ulgens:
Using https://docs.astral.sh/ruff/settings/#lint_extend-safe-fixes for W291 is another option but I don't think this will be needed often enough to be included in the config.

meren:
Excellent. I think this is done now :) Thank you for your help, @ulgens.

By the way, given your response rate to things on GitHub I started to feel like perhaps you do not receive emails I send you? I sent you two emails in the past few day to which I have not received a reply :p IT IS OK IF YOU ARE IGNORING ME BUT I JUST WANTED TO MAKE SURE IT IS NOT UNINTENTIONAL hehe

ulgens:
@meren Ahah sorry, it was just some light weekend procrastination on my end.

PR #2582, opened 2026-04-28
Don't override default ruff rules selection

ulgens:
This change allows the default config to be effective. By the date of the PR, the default lint.select is `["E4", "E7", "E9", "F"]`

https://docs.astral.sh/ruff/configuration/

### DIFF
--- a/anvio/agglomeration.py
+++ b/anvio/agglomeration.py
@@ -204,7 +204,7 @@ class Agglomerator:
 
             while remapping_stack:
                 remapping_item = remapping_stack.pop()
-                current_ref_name = remapping_item[0]
+                remapping_item[0]
                 current_aligned_ref = remapping_item[1]
                 # Record mismatches between query sequences and the agglomerated reference sequence,
                 # with the coordinate system being nucleotide positions in the agglomerated reference.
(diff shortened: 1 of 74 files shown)


meren:
I go through the changes and I see a lot of useful and holy shit how did this survive this long kind of stuff. But I find it interesting that one of these rules replace our anonymous functions that deal with the args variables classes receive with proper inline functions. Is there a technical reason for this? Like new versions of Python will no longer permit it kind of thing? If not, I think we should preserve it :)

ulgens:
I was trying to handle the majority of default styling at once but the result is trying to cover so many things at once, I'll try to separate this into multiple PRs.

> I go through the changes and I see a lot of useful and holy shit how did this survive this long kind of stuff.

Is there anything specific that you would like to prioritize?

> But I find it interesting that one of these rules replace our anonymous functions that deal with the args variables classes receive with proper inline functions. Is there a technical reason for this? Like new versions of Python will no longer permit it kind of thing? If not, I think we should preserve it :)

It's coming from https://docs.astral.sh/ruff/rules/lambda-assignment/. My understanding is, lambdas are expected to be anonymous so if you need your functions named, `def func_name()` syntax is preferred. I think it doesn't make much difference, I prefer to comply with the whole E ruleset. That being said, I added the rule to the ignore list already, we can revisit and revert if needed.

meren:
I think E713, E711, and F541, along with F401 in #2588 are no brainers for starters :)

I understand why PEP8 wants E731, but PEP8 can also want to go fly a kite 🪁

PR #2583, opened 2026-04-28
Organize imports

ulgens:
Enables import organization-related ruff rules and applies the fixes.

- https://docs.astral.sh/ruff/rules/#isort-i
- https://docs.astral.sh/ruff/rules/#flake8-tidy-imports-tid

Triggered the workflow manually, it is passing: https://github.com/merenlab/anvio/actions/runs/25054108473/job/73389515218

This PR comes with a small risk of breaking untested things, but I'm not sure how to minimize that risk. Please let me know if there is anything we would like to check explicitly.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -2,11 +2,17 @@
 
 """Lots of under-the-rug, operational garbage in here. Run. Run away."""
 
-from anvio.version import anvio_version, anvio_codename, major_python_version_required, minor_python_version_required, get_versions
-
-import sys
 import inspect
 import platform
+import sys
+
+from anvio.version import (
+    anvio_codename,
+    anvio_version,
+    get_versions,
+    major_python_version_required,
+    minor_python_version_required,
+)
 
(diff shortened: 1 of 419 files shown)


meren:
Hey @ulgens, thank you for this suggestion. But we are very used to the import structure in the codebase (both sorting, and the logical blocks). It helps me immensely for instance when I open a library to see immediately what is going on. It has been such a long tradition in anvi'o codebase, I'm sure it is serving similar purposes for others. If anything, I would have loved a rule that preserves this :p

ulgens:
Oh got it. Is it okay if I update the checks to just find & clean unused imports and don't touch the organization?

meren:
That would be splendid! My biggest contribution to the codebase was to remove unused imports. I GUESS I CAN RETIRE NOW.

Thank you very much for your patience and help, @ulgens :)

meren:
I am closing this since unused import check is already in the codebase through another PR by @ulgens :)

PR #2584, opened 2026-04-28
Optimize Test CI

ulgens:
Currently, a single run of "Component Tests and Migrations" workflow takes more than 13 minutes. Looking at one of the latest successful runs, https://github.com/merenlab/anvio/actions/runs/25046167874/job/73361693210, I believe there is room for couple of improvements:

* `conda-incubator/setup-miniconda@v4` - 1m30s
  * It looks like this step installing a small but rich envrironment, which I believe includes some unused packages too.
  * It's not cached and packages are being downloaded and installed everytime.
* `Install anvi'o from the codebase` - 1m16s
  * Installing packages with uv instead of vanilla pip shaves of a minute, even with a cold cache.
  * With a warm cache, the step can be skipped.
* `Run component tests` - 9m 35s
  * This step includes 3 test suites without parallelization.
  * None of test suites make use of `num_threads`.

### DIFF
--- a/.github/workflows/component-tests-and-migrations.yaml
+++ b/.github/workflows/component-tests-and-migrations.yaml
@@ -16,10 +16,12 @@ jobs:
       matrix:
         python-version: ["3.10"]
       fail-fast: false
+
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         persist-credentials: false
+
     - uses: conda-incubator/setup-miniconda@8ee1f361103df19b6f8c8655fd3967a8ecb162d5 # v4.0.1
       with:
         # FIXME:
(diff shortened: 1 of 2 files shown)


ulgens:
Please let me know if this gets too noisy for the other contributors, I'm not sure how failed workflow notifications are sent.

meren:
I think it is mostly I and @ivagljiva receive these notifications. She has the patience of a prophet, and I am happy to read such stuff :) If there is someone who is feeling that the notifications feel too frequent, I hope they would either change their notification preferences while you're working on these or let us know :)

ulgens:
Clicked on ready as mistake on mobile and can't revert it now 😢

meren:
Good opportunity for everyone else to see that even pros can make mistakes :p

ulgens:
Eheh, teveccühünüz. (İngilizcem yetmedi.)

meren:
I guess this PR can go away, right, @ulgens? Our CI workflows seem to be working well at the moment, and Component Tests now takes about 3 minutes rather than 13 minutes (again, thanks to you, of course).

PR #2591, opened 2026-04-29
Fix CI - Tests workflow

### LINKED ISSUE #2590: [BUG] CI tries to install Python deps with Python 3.12 and fails

"Component Tests and Migrations" workflow fails with

```python
AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?
```

on the latest master.

https://github.com/merenlab/anvio/actions/runs/25109281683/job/73578621777

## anvi'o version

The issue lives on CI and not relates to a specific anvi'o version.

## System info

-

"Component Tests and Migrations" is failing on the latest master branch. I am not sure when this started or what has changed (and I feel like the investigation won't be worth the time it will take). From what I understand from the logs:

* pip fails to install the packages to the default location: https://github.com/merenlab/anvio/actions/runs/25109281683/job/73578621777#step:6:8 This doesn't happen in a past successful run of the same workflow: https://github.com/merenlab/anvio/actions/runs/25046167874/job/73361693210#step:6:8
* Then it defaults to a different Python setup (3.12) and tries to install dependencies there and fails.

## Files / commands to reproduce the issue

Trigger [this](https://github.com/merenlab/anvio/actions/workflows/component-tests-and-migrations.yaml) workflow using `workflow_dispatch` trigger with brach master.

ulgens:
Fixes #2590

Turned out miniconda [released a new version yesterday](https://www.anaconda.com/docs/getting-started/miniconda/release-notes#miniconda-26-3-2-2), and the workflow started to use it automatically, thanks to `miniconda-version: 'latest'`

On top of that, 85735407452fa5d16614052d12e6a2c744808e6a had an issue with bash syntax.

Successful workflow run after the fixes: https://github.com/merenlab/anvio/actions/runs/25125976761/job/73638995385

### DIFF
--- a/.github/workflows/component-tests-and-migrations.yaml
+++ b/.github/workflows/component-tests-and-migrations.yaml
@@ -22,7 +22,11 @@ jobs:
         persist-credentials: false
     - uses: conda-incubator/setup-miniconda@8ee1f361103df19b6f8c8655fd3967a8ecb162d5 # v4.0.1
       with:
-        miniconda-version: 'latest'
+        # FIXME:
+        #   The next version (26.3.2-2) causes https://github.com/merenlab/anvio/issues/2590
+        #   Release notes: https://www.anaconda.com/docs/getting-started/miniconda/release-notes
+        #   Available versions: https://repo.continuum.io/miniconda/
+        miniconda-version: "py310_26.1.1-1"
         auto-update-conda: true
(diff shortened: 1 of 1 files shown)


meren:
Ah. Great find. Thank you, @ulgens!

PR #2592, opened 2026-04-30
Enrich ruff rules

ulgens:
Coming from https://github.com/merenlab/anvio/pull/2582#issuecomment-4343760088, enables E711, E713 and F541

- E711: https://docs.astral.sh/ruff/rules/none-comparison/
- E713: https://docs.astral.sh/ruff/rules/not-in-test/
- F541: https://docs.astral.sh/ruff/rules/f-string-missing-placeholders/


meren:
🎉

PR #2593, opened 2026-04-30
Install Python packages with uv

ulgens:
Shaves of ~1 minute from the "Install anvi'o from the codebase" step without changing anything in the env

https://docs.astral.sh/uv/#highlights

# Before
* https://github.com/merenlab/anvio/actions/runs/25185510792/job/73841644493
* 1m 13s

# After
* https://github.com/merenlab/anvio/actions/runs/25185719993/job/73842375300
* 25s


meren:
Wow! (Or Wuv?) :)

Thank you, @ulgens!

PR #2594, opened 2026-05-01
Remove actions/setup-node step

ulgens:
After realizing `.conda/environment.yaml` already has nodejs listed as a dependency, I thought the node setup in the workflow may be redundant. The workflow works just fine without that step: https://github.com/merenlab/anvio/actions/runs/25219232420/job/73947027280


meren:
Interesting. Thanks for catching that :)

Issue #2605, opened 2026-05-18
[FEATURE REQUEST] Adding contig-level information to the contigs database


ivagljiva:
## The need

When analyzing metagenomes, we would like to be able to identify which contigs are eukaryotic vs non-eukaryotic, or which are plasmids/viruses, or import contig-level taxonomy information. This has numerous downstream applications, such as the ability to run domain-specific gene callers on subsets of contigs.

## The solution

We need a new table in the contigs database, keyed by contig, to store this information.

We also need an artifact (ie a TSV table), that allows people to import these data into a contigs database.

We need to do some research into tools that classify contigs (Whokaryote, Genomad, etc) so that we know what these outputs look like and envision how we might eventually connect this into the codebase.

## Beneficiaries

Anybody working on multiple domains of life, for instance :)
We would eventually like to incorporate the ability to run eukaryotic gene callers on only the eukaryotic contigs, or viral functional annotation programs on only genes from the viral contigs, etc.

guille0387:
## Hi There!

ivagljiva:
Branch for this project: `contig_level_info`

u-xixi:
# Designing the format of the artefact

We need to review tools that provide contigs level info, understand what they do, and what kind of output they produce. So we know what information we want the new artefact to have.

## Output of tools that provide contigs level info
geNOMAD: https://github.com/apcamargo/genomad#understanding-the-outputs
whokaryote: (from https://github.com/LottePronk/whokaryote)

> A file with all contig headers that were classified as eukaryotic (eukaryote_contig_headers.txt), and a similar file for prokaryotic contig headers (prokaryote_contig_headers.txt)
A fasta file with only contigs that were longer than 5000 bp, called contigs5000.fasta
A gene coordinates file generated by prodigal called contigs_genes.genes
A .csv file with all the calculated features called featuretable.csv
A file with the tiara predictions called tiara_pred.txt
A .csv file with the calculated features and the predictions called featuretable_predictions_[model].tsv
A .tsv file with only the contig ID and the predictions called whokaryote_predictions_[model].tsv

and the key file is: `whokaryote_predictions_T.tsv`, which looks like this

```
contig	predicted
s_106058wDxxx304mPxxx_000000000001	prokaryote
s_106058wDxxx304mPxxx_000000000002	prokaryote
s_106058wDxxx304mPxxx_000000000005	prokaryote
s_106058wDxxx304mPxxx_000000000007	prokaryote
```

[...]

PR #2609, opened 2026-05-18
Add anvi-import-contig-classification

tucker4:
Add anvi-import-contig-classification


ivagljiva:
This is related to https://github.com/merenlab/anvio/issues/2605. We are just going to merge it since it will update the `contig_level_info` branch only

PR #2615, opened 2026-05-19
Programs to setup and annotate contigs-db files with function annotations derived from GlobDB

meren:
[GlobDB](https://globdb.org/home) is a curated database of microbial genomes, and thanks to @dspeth's efforts, it also offers a set of gene families, which are also curated and most critically accompanied by gene-family-level Local Alignment Score Ratio (LASR) cutoffs to increase specificity of de novo gene families while minimizing false positives.

This PR adds two new anvi'o programs, (1) `anvi-setup-globdb-functions` and (2) `anvi-run-globdb-functions`, to (1) download and processes GlobDB Functions database and (2) annotate `contigs-db` files.

The PR includes documentation for `anvi-setup-globdb-functions`, `anvi-run-globdb-functions`, and the `globdb-data` anvi'o artifact.

This PR and the tools it brings into the anvi'o ecosystem represent an initial attempt both on GlobDB and anvi'o side, but we expect that things will grow from here.

Daan and I tested it on various contigs-db files, and things seem to be working as expected. There are no component tests at this stage, but when we reach a level of maturity we will add them to the codebase.


meren:
Ping @dspeth.

dspeth (changes requested):
@meren 

general:
At the moment the codebase mixes the use of globdb and globaa, using globdb mostly in commands/options/filenames and globaa in variables and the output tables. I think we should standardize this, and use either one throughout. I think using globaa throughout the code is the best option, and the documentation can describe how the GlobAA and GlobDB are related.


line 544:
the GlobDB identifiers have "___" to separate genome ID and locustag. If we use ___ to concatenate gene family ids and sequence IDs splitting on ___ afterward should take into account that the sequence id is then [1]___[2]


line 866:
The DIAMOND search might need an explicit sesitivity declaration. Sensitivity changes the algorithm used and could lead to different best hits being prioritized. Currently, the default sensitivity parameter is not equal to any of the ones that can be specified using "--sensitivity"
We might also have to specify the search algorithm used, since diamond adjusts this dynamically based on the search, making things not always reproducible. See https://github.com/bbuchfink/diamond/wiki/3.-Command-line-options (--algo option)

PR #2631, opened 2026-05-20
Adding Bastien Camillo informations

bastien-camillo:


meren:
@bastien-camillo, you forgot to put your 900x900px image under the AVATARS directory.

PR #2633, opened 2026-05-20
anvi-compute-functional-consensus: *FunctionalConsensus classes

ChiaraVanni:


ivagljiva:
hey @ChiaraVanni , there are three commits in here related to `anvi-compute-functional-consensus` that probably shouldn't be merged as part of your contributor info update :)

I think the cleanest way to fix this would be to do the following:
```
git checkout master
git pull
git checkout -b chiara-vanni-contributor-info-clean
git cherry-pick 51bd4805883dd2e7549b6272933816a79fd8d58d
git push
```
And then make a new PR from the `chiara-vanni-contributor-info-clean` branch :)

PR #2642, opened 2026-05-20
Exit with status code 1 in case of an error - fatal warning

ulgens:
The tests inside the component tests pipeline was failing but the workflow itself wasn't: https://github.com/merenlab/anvio/actions/runs/26166457784/job/76971855168

I updated the exit code for error - fatal warning cases so that any tooling wrapping anvio will be able to understand something is wrong. The workflow fails properly now: https://github.com/merenlab/anvio/actions/runs/26168320675/job/76978629756

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -87,7 +87,7 @@ if '--tmp-dir' in sys.argv:
         os.environ['TMPDIR'] = TMP_DIR
     except Exception as e:
         print("OSError: ", e)
-        sys.exit()
+        sys.exit(1)
 
 def P(d, dont_exit=False):
     """Poor man's debug output printer during debugging."""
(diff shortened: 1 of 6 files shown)


ulgens:
Btw,

```python
except Exception as e:
    print(f"error: {e}")
    sys.exit(1)
```

has pretty much the same effect as not catching the error (or, depending on the case, just `raise`.). Do we need to handle those cases like this?

meren:
These are very rare cases, and I'm not sure why they were left the way they are.

In the best case scenario, `sys.exit()` should not be used anywhere and there shouldn't be any pattern of except / print / exit outside of `anvio/cli/*` (since no tool should import anything from under cli, so all should be fine).

ulgens:
I'm a bit lost about the next step :) Atm CI is failing silently, and I was planning to follow this PR with the actual fix for the failure.

meren:
I'm very sorry I missed your previous comment, @ulgens. The next step is to merge this branch. I apologize for the delay :(

Issue #2617, opened 2026-05-19
[FEATURE REQUEST] general implementation for annotation with user-defined databases


iwilkie:
## The need

Currently each database that has been implemented for functional annotation has its own set of anvi'o tools (e.g. CAZymes, KEGG, COG and PFAM). Adding more annotation sources would streamline user analyses. 

## The solution

We want to generate a function that can run functional annotations based on user-defined databases, to avoid the need to add new functions for each new database. Something like "anvi-annotate".

## Beneficiaries

Any users who want to add annotations from other available sources or even custom databases, instead of annotating externally and then importing them into anvio, you could define the database name and path and run it in this function.

Issue #2300, opened 2024-07-09
[FEATURE REQUEST] Adding a program to annotate viral PHROGs via their HMMs


ivagljiva:
A small project to improve anvi'o, based upon feedback/ideas @FlorianTrigodet and I heard from our colleagues at the QIB in Norwich.

## The need

The [PHROGs database](https://phrogs.lmge.uca.fr/) contains orthologous groups of proteins from viruses. You can download their HMMs, but currently annotating them in anvi'o contigs databases would require either 1) running HMMER yourself and importing the resulting functions, or 2) making a custom HMM source out of the PHROGs and running them with `anvi-run-hmms`. 

## The solution

We could implement a program similar to `anvi-run-kegg-kofams` or `anvi-run-pfams` for the PHROGs database.

## Beneficiaries

Those who study viruses in anvi'o.

Ge0rges:
I may try to take a stab at this after ISME. I've written scripts to make custom HMM sources so I have those already, and wanted to look at viral stuff in my metagenome anyway. If someone else wants to work on this maybe we can help each other out let me know!

Ge0rges:
@ivagljiva would you rather see an implementation like what was [done](https://github.com/merenlab/anvio/pull/2244) for the MDM Defense Finder HMMs with a `anvi-script-gen-phrogs-hmm-directory` command, or something more integrated such as `anvi-run-phrogs`? I guess the precedent I established for these was to make directories easily accessible but I guess there's a conversation be had here about what makes the most sense.

meren:
Hi @Ge0rges, I just wanted to quickly come in and let you know that everyone is on vacation, and that's why we have been a bit quiet on Discord and GitHub. We will catch up with all these threads as people will start coming back in soon :)

PR #2644, opened 2026-05-20
General implementation for annotation with user-defined databases

l-gallucci:
As describe in the issue #2617, currently each database that has been implemented for functional annotation has its own set of anvi'o tools (e.g. CAZymes, KEGG, COG and PFAM). Adding more annotation sources would streamline user analyses.

We generated a function that can run functional annotations based on user-defined databases, to avoid the need to add new functions for each new database. 

Any users who want to add annotations from other available sources or even custom databases, instead of annotating externally and then importing them into anvio, you could define the database name and path and run it in this function.

This idea and the approaches that we used came to being through discussions with @dspeth , @igorspp , @ChiaraVanni , @elxerone , @iwilkie and @l-gallucci. 

@l-gallucci and @iwilkie wrote the code (cough cough with Claude...). 

@igorspp, @l-gallucci, and @iwilkie tested the implementation.

The new programs are: 1. anvi-setup-user-annotation-db and 2. anvi-run-user-annotation.

The general idea is that users provide a tab-delimited file with database names and paths that can be set up for HMM/Diamond searches in the next command. If an HMM model has an associated Fasta file (called a 'companion Fasta'), this can also be provided in the third column. In this case, anvi-run-user-annotation will run both HMMsearch and Diamond, providing double output for confirming the HMM annotation. The Fasta files are always used as input by anvi-setup-user-annotation-db to generate a DMND file with make-db. The annotation/hits are saved in the gene_functions table and can be exported or searched using the corresponding anvi programs: anvi-search-functions and anvi-export-functions. In the interactive interface, they appear as expected, alongside COG, CAZymes, Pfam, etc.

We tested this on our example genomes and the CONTIGS.db in the anvi/tests/sandbox/data/genomes/bacteria/P_marinus_CCMP1375.db directory. We compared the annotation from anvi-run-cazymes with the corresponding dbCAN HMM annotation and found that the hits matched.

All the relevant examples are included in the documentation files generated for both functions.


Ge0rges:
Does this make the following feature request https://github.com/merenlab/anvio/issues/2300 obsolete then? PHROGS has the particularity of providing `.hhm` files which need to be converted to `.hmm`. However that may not justify having its own command.

l-gallucci:
In principle, this functionality overlaps with existing programs like `anvi-scan-trnas`, `anvi-run-scg-taxonomy`, and `anvi-run-ncbi-cogs`. However, following @ivagljiva's suggestions, we decided to keep this as a separate function for now. If this implementation proves successful, it could serve as a unified framework to consolidate the others in the future. The core value of this new function is its ability to prepare both HMM and DIAMOND files simultaneously. Crucially, it accounts for companion FASTA files when they are present for a given HMM. This allows users to seamlessly cross-verify their HMM hits with a follow-up DIAMOND search.

Issue #2650, opened 2026-05-21
[BUG] Transfer_RNAs contains CCT_Arg, while hmm_hits contains Arg_CCT


Kekananen:
## Short description of the problem

While running on the master branch at commit `a4204c5697ed5aba79a17d2c3ad3d95666052e86`, there is an error thrown while running the contigs workflow component tests. The workflow itself completed; however, the failure happens immediately afterward when the self-test tries to summarize/display the generated contigs DBs via anvi-display-contigs-stats. 

The DB says the known model is CCT_Arg, but the hit row says Arg_CCT. When `get_gene_hit_counts_per_hmm_source() reaches the hit, gene_hit_counts['Transfer_RNAs']['Arg_CCT']` does not exist, causing a `KeyError: 'Arg_CCT'`

## anvi'o version

Keep the header of this section, but replace this text with the output of this command in your terminal:

```
(anvio-dev) [kananen.13@unity-login1 tests]$ anvi-self-test --version
Anvi'o .......................................: eunice (v9-dev)                                                                                                                                     
Python .......................................: 3.10.17                                                                                                                                             

Profile database .............................: 42                                                                                                                                                  
Contigs database .............................: 24                                                                                                                                                  
Pan database .................................: 21                                                                                                                                                  
Pangraph database ............................: 4                                                                                                                                                   
Genome data storage ..........................: 8                                                                                                                                                   
Structure database ...........................: 2                                                                                                                                                   
Metabolic modules database ...................: 4                                                                                                                                                   
tRNA-seq database ............................: 2                                                                                                                                                   
Genes database ...............................: 6                                                                                                                                                   
Auxiliary data storage .......................: 2                                                                                                                                                   
Workflow configurations ......................: 5 
```

## System info

OS Information:
```
(anvio-dev) [kananen.13@unity-login1 tests]$ lsb_release -a
LSB Version:    :core-4.1-amd64:core-4.1-noarch:cxx-4.1-amd64:cxx-4.1-noarch:desktop-4.1-amd64:desktop-4.1-noarch:languages-4.1-amd64:languages-4.1-noarch:printing-4.1-amd64:printing-4.1-noarch  
Distributor ID: RedHatEnterprise
Description:    Red Hat Enterprise Linux release 8.10 (Ootpa)
Release:        8.10
Codename:       Ootpa
```

## Detailed description of the issue

[...]

PR #2651, opened 2026-05-21
Trnahits bug: CCT_Arg vs Arg_CCT naming mismatch

Kekananen:
For issue #2650,   I verified this in the generated test DB: hmm_hits_info for Transfer_RNAs contains CCT_Arg, while hmm_hits contains Arg_CCT using the latest commit to the master branch `a4204c5697ed5aba79a17d2c3ad3d95666052e86`. 

I believe the right fix is to make the metadata use the same naming convention as the hit rows by switching `possible_genes.append(f'{acdn}_{decoded_aa_type}')` to `possible_genes.append(f'{decoded_aa_type}_{acdn}')` in the `trnahits.py` file.

This was accomplished by two small changes:
-  Transfer_RNAs metadata now registers possible genes as decoded_AA_type_anticodon, e.g. Arg_CCT, matching the actual `hmm_hits.gene_name values`.
- Added a local consistency check before writing tRNA hits into the HMM tables. If future naming logic drifts again, it raises a ConfigError at write time instead of producing a bad DB that crashes later in `anvi-display-contigs-stats`.

People would beed to regenerate their contig.db later on if they wanted this case to be fixed. 

The workflow with running the command `anvi-self-test --suite workflow-contigs -T 40 --output-dir test_output` in the `test` directory in anvi'o now runs to completion; however, due to running it on a remote server without x11 / port forwarding enabled **I was unable to see/test the anvi-display-contigs-stats to verify the visuals weren't odd**.


Kekananen:
I would appreciate if someone could verify this error is not something with my local install, as I am a little surprised this issue hasn't caused a larger volume of reports.

Kekananen:
This essentially closes https://github.com/merenlab/anvio/issues/2650

meren:
@semiller10, @Kekananen, do you think this needs a migration script?

semiller10:
It could use a migration script, but the issue seems minor to me (maybe not to other people), surfacing in `anvi-display-contigs-stats`, so perhaps it could be saved for a future contigs database migration alongside substantial changes.

meren:
OK. Let's go, and we have evidence here for how to solve it if something breaks.

PR #2659, opened 2026-05-21
Extending "Provides" and "Requires" with "Can use", and "Can provide"

meren:
This PR introduces a level of nuance that has been sorely missing in our program descriptions. For instance, the head matter for `anvi-interactive` looked like this:

```
__requires__ = ['profile-db', 'single-profile-db', 'contigs-db', 'genes-db', 'bin', 'view-data', 'dendrogram', 'phylogeny']
__provides__ = ['collection', 'bin', 'interactive', 'svg', 'contig-inspection']
__description__ = "Start an anvi'o server for the interactive interface"
```

Which generated a help output that looked like this:

image

With this update, the head matter for the same program looks like this:

```
__requires__ = ['profile-db']
__can_use__ = ['single-profile-db', 'contigs-db', 'genes-db', 'bin', 'view-data', 'dendrogram', 'phylogeny']
__provides__ = ['interactive']
__can_provide__ = ['svg', 'collection', 'bin', 'contig-inspection', 'dendrogram', 'phylogeny']
__description__ = "Start an anvi'o server for the interactive interface"

```

image

(which makes me realize it is entirely wrong since anvi-interactive does not require anything but a TAB-delimited text file and *can* use profile-db, too :p I will take care of it separately).

With this PR,

* all program head matters are revisited
* `anvi-help` (YES, the program no one uses) updated accordingly,
* Argparse modified so help menus look pretty (like shown above)
* and the `AnvioPrograms` class is updated to make sure our help output on the web will look pretty.


anvio/cli/compute_functional_enrichment_across_genomes.py
ivagljiva, line 1: this is looking great so far :)  but what is the best strategy for either-or input situations like with this program? `anvi-compute-functional-enrichment-across-genomes` can use either external genomes OR internal genomes OR a genomes-storage db. Obviously we cannot say that all of them are required, but we also cannot say that none of them are required. Based on this example, it seems like the current strategy is to pick one (maybe the most commonly used one?) and report only that one as required. But that is a bit misleading :( and it will break efforts like Bastien's to find a way to automatically connect things via their inputs and outputs.

At the risk of making things more complicated, what if we encode the actual artifact relationships in these statements using logical operators? Kind of like a KEGG Module definition :) It would require us to implement some more sophisticated parsing logic rather than simply reading from these lists.

Maybe like this?
```
__requires__ = "groups-txt & functions & (genomes-storage-db | external-genomes | internal-genomes)"
```

So instead of lists, we get strings with artifact names connected by AND/OR operators (and possibly parentheses). Everything currently in `__requires__` will most likely be connected by AND operators (except for these special cases where we have either-or relationships for at least a subset of inputs) while everything currently in `__can_use__` will most likely be connected by OR operators (except for special cases where we have some artifacts that must be used together -- not that I can think of any at the moment). Same for `__provides__` and `__can_provide__`

Thoughts?

anvio/cli/compute_metabolic_enrichment.py
ivagljiva, line 1: this one requires kegg-metabolism OR user-metabolism AND groups-txt, for instance :)

I also just realized that external-genomes, internal-genomes, and functions should not actually be in these lists. I think they are a carryover from when we split the original enrichment program into three with different input types

meren: This is GREAT idea. It is going to be a bit challenging, but I'll do it happily! :)

meren: Hey @ivagljiva, I really like this idea, but after a brief discussion with @bastien-camillo I realised that this level of specificity over these statements may render some of the user displays very difficult without a very carefully engineered translation system.

With a careful design, for some applications -such as an interactive workflow building interface- `AnvioPrograms` class can return simple requires / provides statements, and in other applications -such as building help pages- it can return nested data structures for interface rendering step to resolve carefully.

On the one hand I feel like we could merge this and keep this idea as a feature request. On the other I feel like we should do our due diligence to the task for excellence and go for it 100% before merging anything.

I am on the fence. What do you think?

meren: I see also @telatin reacted to this -- so I wanted to ping him also in case he has an opinion one way or another.

telatin: I cannot believe you edited all those files to split between require and can_consume. Sorry, I'm too slow :)

So, I think the pragmatic change of this PR might add a level of benefits without introducing drastic changes that would be already worth pursuing. This would reflect also to a better documentation website.

Iva's idea is lightyears ahead and yes, I would keep an open [DISCUSSION] thread more generally on introspection and metadata, so that we might come up with a redesign for citations, artifact input/output, ext tools usage etc for a future re-design?

meren: Just talked to Iva in person :) She is happy with this resolution: merge this now, keep the discussion going. I will do that, and start a discussion pinging everyone.

PR #2663, opened 2026-05-21
Add modifications parsing from MM and ML tags in BAMs during `anvi-profile`

### LINKED ISSUE #2152: [DISCUSSION] Methylation processing in anvio
Hello MerenLab,

While at EBAME I was encouraged to open this discussion by @FlorianTrigodet and @ivagljiva, and bring it to @semiller10's attention.

DNA methylation is on the verge of constituting a significant amount of sequencing data. Most methylation detection assays till now have relied on bisulfite sequencing. However this method is limited to certain types of methylation, and is not useful in the context of understanding community level DNA methylation. While PacBio offers very accurate methylation data, it is expensive. Nanopore promises to make epigenetic studies widely accessible thanks to it recent embrace of DNA modification detection. They already support 2/3 types of DNA methylation, and expect to support all 3 by next May. Moreover, Nanopore's method enable for methylation detection in any nucleotide context (sequence).

Thus I was wondering if there is any interest in adding DNA methylation support to anvio. This would consist at least in reading in methylation data from BAM files and match it to the corresponding read. More convenience functions could include things like:

- Getting a "methylated MAG" by mapping reads back onto the MAG into anvio (even if mapping is done externally)
- Getting differentially methylated regions (DMR) of a MAG (see [MethylKit](https://github.com/al2na/methylKit) and [modkit](https://github.com/nanoporetech/modkit)).
- Getting the annotation of a DMR

I would be very happy to work with the Anvio team towards making this happen, though my knowledge of the codebase is currently limited.

Ge0rges:
This pull request implements a first step towards implementing https://github.com/merenlab/anvio/issues/2152. 

The goal of this pull request is to add a new table called `modifications` to the `profile.db` artefact. This table keeps track of a pileup of nucleotide modifications for each position along a contig as parsed by pysam from `MM` and `ML` tags. For more information about these tags see: [the pysam docs](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.modified_bases) and the [SAM standard](https://samtools.github.io/hts-specs/SAMtags.pdf).

This is done by adding a new flag called `--include-modifications` to `anti-profile`. Along with this flag, are options to filter by minimum coverage and probability (defined in the ML tag). These default to 10 and 95% respectively. The filter flags are called `--modification-filter-default` and `--modification-filter` giving the user the option to filter per modification type or using the same filter for all. 

The information from this table is also meant to be piped to the UI to see a simple ratio, per base, of the modifications type in the inspect view. 

Currently, the table counts do not match the counts shown by IGV when inspecting a test BAM. Moreover, the UI does not seem to pipe this information at all. 

Testing instructions:
- Download the attached zip
- `anvi-gen-contigs -f hambi_1287_plas02_circ.fasta`
- `anvi-profile --include-modifications -c CONTIGS.db -i hambi_1287_plas02_circ_mapped_ft.bam`
- Load your BAM and FASTA into IGV
- In IGV, right click on the BAM track, select `Color alignments by ...` > `base modifications (all)`
- `sqlite3 hambi_1287_plas02_circ_mapped_ft/PROFILE.db`
- `SELECT * FROM modifications ORDER BY 3 LIMIT 10;`

Compare the counts at a position shown in the table and in IGV. Note that the default filter is 95% in anvi'o whereas IGV will show modifications with any probability.

[Archive.zip](https://github.com/user-attachments/files/28110353/Archive.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 # system / OS-specific helper files
 .DS_Store
 
+# IDE and editor files
+.idea/
+.vscode/
+
 # Python artifacts and compiled files
 *.pyc
 *.so
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "python-envs.defaultEnvManager": "ms-python.python:conda",
+    "python-envs.defaultPackageManager": "ms-python.python:conda"
+}
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -590,7 +590,7 @@ D = {
             {'default': [],
              'action': 'append',
              'metavar': 'TYPE:THRESHOLD',
-             'help': "Set a minimum modification probability per modification type (e.g., 'A:0.9'). This parameter can be "
+             'help': "Set a minimum modification probability per modification type (e.g., 'm:0.9'). This parameter can be "
                      "used multiple times to set thresholds for multiple modification types. Values must be between 0 and 1."}
                 ),
     'modification-filter-default': (
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -578,6 +578,13 @@ D = {
                      "having this information, you can skip storing this information by providing this flag. Note: If "
                      "--skip-SNV-profiling is provided, --skip-INDEL-profiling will automatically be enforced."}
                 ),
+    'include-modifications': (
+            ['--include-modifications'],
+            {'default': False,
+             'action': 'store_true',
+             'help': "If you use this flag, anvi'o will parse base modification tags (MM/ML) from BAM files during profiling "
+                     "and store them in the profile database for downstream analysis."}
+                ),
     'return-AA-frequencies-instead': (
             ['--return-AA-frequencies-instead'],
             {'default': False,
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -585,6 +585,30 @@ D = {
              'help': "If you use this flag, anvi'o will parse base modification tags (MM/ML) from BAM files during profiling "
                      "and store them in the profile database for downstream analysis."}
                 ),
+    'modification-filter': (
+            ['--modification-filter'],
+            {'default': [],
+             'action': 'append',
+             'metavar': 'TYPE:THRESHOLD',
+             'help': "Set a minimum modification probability per modification type (e.g., 'A:0.9'). This parameter can be "
+                     "used multiple times to set thresholds for multiple modification types. Values must be between 0 and 1."}
+                ),
+    'modification-filter-default': (
+            ['--modification-filter-default'],
+            {'default': 0.95,
+             'type': float,
+             'metavar': 'THRESHOLD',
+             'help': "Default minimum modification probability to use for modification types not explicitly listed in "
+                     "--modification-filter entries. Must be between 0 and 1."}
+                ),
+    'min-coverage-for-modifications': (
+            ['--min-coverage-for-modifications'],
+            {'default': 10,
+             'type': int,
+             'metavar': 'INT',
+             'help': "Minimum per-position coverage required before calculating modification ratios for contigs during "
+                     "interactive display or export."}
+                ),
     'return-AA-frequencies-instead': (
             ['--return-AA-frequencies-instead'],
             {'default': False,
--- a/anvio/bamops.py
+++ b/anvio/bamops.py
@@ -402,6 +402,11 @@ class Read:
         self.reference_start = read.reference_start
         self.reference_end = read.reference_end
 
+        self.modifications_mm = read.get_tag('MM') if read.has_tag('MM') else None
+        self.modifications_ml = read.get_tag('ML') if read.has_tag('ML') else None
+        self.modified_bases = read.modified_bases
+        self.query_pos_to_ref_pos = read.get_reference_positions(full_length=True)
+
         if read.has_tag('MD'):
             self.reference_sequence = np.frombuffer(read.get_reference_sequence().upper().encode('ascii'), np.uint8)
         else:
--- a/anvio/bamops.py
+++ b/anvio/bamops.py
@@ -403,18 +403,9 @@ class Read:
         self.reference_start = read.reference_start
         self.reference_end = read.reference_end
 
-        # Parse modification tags (MM/ML) and pysam's convenience attribute
-        # `modified_bases` can raise low-level parsing errors for malformed tags.
+        # Get pysam's convenience attribute `modified_bases`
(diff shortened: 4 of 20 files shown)


Ge0rges:
I have a suspicion may be misunderstanding the positions in splits/contigs versus how they are shown in IGV... ?

meren:
Oh, yes, it is very likely :) if positions are for splits, you may need to convert that relative position to contig position.

For your tests you can create your contigs db with `--split-length 0`, so you will have the same coordinates with IGV for every position.

Ge0rges:
Hmm doesn't seem to be the issue then.

Ge0rges:
After looking at things in the morning with Florian, seems like the table building is actually working perfectly. Needed to come back to this with some sleep. 

Florian and I have agreed that keeping `modifications` as a separate table inside the database is likely better than including it in `variable_nucleotides` as most rows are unlikely to align anyway, and modifications can have multiple rows per position unlike the rows in `variable_nucleotides` (since `modification` is a column). 

However, we also agreed to add `corresponding_gene_call numeric, in_noncoding_gene_call numeric, in_coding_gene_call numeric, base_pos_in_codon numeric, codon_order_in_gene numeric, coverage numeric` into the `modifications` table. So I'm working on that.

Ge0rges:
This is fixed and working. 

Only untested part is the migration script.

Ge0rges:
@copilot resolve the merge conflicts in this pull request

FlorianTrigodet:
@Ge0rges Sorry for the long time waiting. What is the status in your opinion? Does the code do what you needed and should I have a look at it? I think the migration script needs a little revision.

Ge0rges:
Hi Florian,

According to my last tests the modifications table gets built correctly.

Georges Kanaan
https://gkanaan.com

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

FlorianTrigodet:
Been learning a lot about DNA/RNA modification today. Here are some thoughts for discussions.

# How to compute relative coverage of modification.
> Currently, the table counts do not match the counts shown by IGV when inspecting a test BAM.

I've been learning about how modification coverage are usually computed, and I think I understand why the values here are different from what is reported in IGV, or standard tools like modkit (by ONT).

There are four different situations to consider, so let's use an example here. Let's say we have a reference position which is a C and here is was we get from the aligned reads at that position. In this context, we're looking at the 5mC modification and we apply the probablity threshold of 75% (for the example):

```
  REFERENCE
    (+) 5'-..........[C]..........-3'
    (-) 3'-..........[G]..........-5'
                      ^  modifiable cytosine is on the (+) strand only;
                         the (-) strand carries G -> nothing to modify here

  FORWARD reads   (base = C ->  5mC is scored)
    f1  >>>>>>>>>>>>>[C]>>>>>>>>>>>>>   P(5mC)=0.76   MODIFIED
    f2  >>>>>>>>>>>>>[C]>>>>>>>>>>>>>   P(5mC)=0.76   MODIFIED
    f3  >>>>>>>>>>>>>[C]>>>>>>>>>>>>>   P(5mC)=0.53   FAIL
    f4          >>>>>[C]>>>>>>>>>>>>>   P(5mC)=0.24   CANONICAL
    f5  >>>>>>>>>>>>>[C]>>>>>           P(5mC)=0.12   CANONICAL
    f6  >>>>>>>>>>>>>[C]>>>>>>>>>>>>>   P(5mC)=0.02   CANONICAL
    f7  >>>>>>>>>>>>>[C]>>>>>>>>>>>>>   P(5mC)=0.00   CANONICAL

  REVERSE reads   (base = G ->  nothing to score here)
    r1  <<<<<<<<<<<<<[G]<<<<<<<<<<<<<   --           off-strand (no call)
    r2  <<<<<<<<<<<<<[G]<<<<<<<<<<<<<   --           off-strand (no call)
    r3  <<<<<<<<<<<<<[G]<<<<<<<<<<<<<   --           off-strand (no call)
    r4  <<<<<<<<<<<<<[G]<<<<<<<<<<<<<   --           off-strand (no call)
    r5  <<<<<<<<<<<<<[G]<<<<<<<<<<<<<   (x5 reverse reads, none assessable)
```

In this context, we have at total coverage of 7+5=12x. Two modification passed the threshold. One could imagine that we could divide 2/12=16% of modification at that location. But that would be wrong for two reasons:

 1- Reverse reads CANNOT test for the modification because the G was sequenced there and the modification lives on the C. Opposite strand reads are NON informative: neither presence nor absence of modification.
2- What to do with ambiguous modification call? In this case, with a threshold of 0.75: everything above is a accepted modification and by mirroring the threshold, anything under 0.25 is an accepted NON-modification also called the [CANONICAL base in the field](https://github.com/nanoporetech/modkit#definitions). Any value in between `threshold` and `1-threshold` is ambiguous and is usually NOT counted as the denominator for the fraction of modification at a given site. 

Based on current practices in the field, the reported fraction here would be 2 / (2 MODIFIED + 4 CANONICAL) = 2/6 = 33%.

Current practices in the field does not mean they are right though. And I would not be upset if we compute 2 / (2 MODIFIED + 1 FAILED + 4 CANONICAL) = 2/7 = 28.5%. Basically the proportion of modification passing the threshold over the total number of read that CAN inform us about that modification.

# Platform specific modification

The 5mC modification does not live alone for is found in CpG sites. Look how beautiful it is:
image

ONT reads will carry the per strand information, but Pacbio is [strand-collapsed by construction](https://github.com/PacificBiosciences/jasmine/blob/main/README.md#5mc).

While the profiling in itself is not too complicated and we need to keep count of position that PASSED or FAILED a threshold, or HAVE NO modification information. And later a consumer program can choose how to compute the expected ratio. But we also have to think about analysis tools/program that could benefit from this profiling. For instance, something that identify and merge CpG site info (or not if it is coming from Pacbio sequence, because they already did it).

@Ge0rges let me know what you think and what you know about all this, and what you imagine would be the best application of DNA/RNA modification in anvi'o (binning, identify gene calls with more or less modification etc).

Ge0rges:
Hi Florian,

Yep the thresholding was the reason things didn't match but I had later identified that and was able to test things. I built in a filter threshold flag to be able to keep only high probability calls.

I think downstream applications will most likely involve looking at how methylation levels may change between samples, especially in areas of regulatory importance such as promoters (and therefore the gene called information is useful). It could also be used as a QC metric for bins manually.

I think more complex downstream analyses could include statistics to identify differentially methylated position and regions. That can be implemented once there is a need.

Georges Kanaan
https://gkanaan.com

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

PR #2666, opened 2026-05-21
Reaction network FormulaMatcher match by charge in addition to formula

semiller10:
It is important to be able to match a chemical formula to compounds in a reaction network taking into account the charge.

@ivagljiva, I now see you made an equivalent change to this in branch [`compound_matching`](https://github.com/merenlab/anvio/tree/compound_matching), so feel free to delete it :)


ivagljiva:
yay :) glad this is now in the main branch!

PR #2671, opened 2026-05-22
Add anvi-screw-you: It empathizes with the user once they are frustrated

sarilog:
This little program allows the user to vent and get a warm message that may even make them laugh :)


meren:
I think `anvi-screw-this` might be a more generic program name than `anvi-screw-you`.

The prefix `anvi-` in anvi'o programs often serves as a means to ask anvi'o to perform a specific task. Get sequences for my HMM hits (`anvi-get-sequences-for-hmm-hits`), display a pan-db (`anvi-display-pan`), etc.

Asking anvi'o to 'screw you' (i.e., `anvi-screw-you`) doesn't have the same ring, since 'you' in this context is anvi'o itself, but the action requested is vague, especially since based on the code, it often results in an apology.

In that sense I think 'screw this' would better match most situations/frustrations and would be more compatible with the rest of the codebase.

My 2 cents.

sarilog:
I changed the name to make it more compatible with the rest of the codebase, and I also fixed the display of the help so it will be more meaningful now.

PR #2673, opened 2026-05-22
anvi-report-indels: track mobile indels

FlorianTrigodet:
Adds a new program, **`anvi-report-indels`**, that aggregates large (or small) indels from one or more profile-dbs into a **loci × samples element-presence matrix** ready for `anvi-interactive --manual`. One possible use is to track mobile genetic elements (insertion sequences, ICEs, transposons, etc.) as they appear, disappear, or jump between sites across long-read metagenomes.

This PR also includes a small but consequential fix to `anvi-profile`'s CIGAR handling so that soft-clipped read regions are no longer recorded as insertions in the indels table.

## What it does
Gets indels sequence form the indels table, recover deletion sequence from the contig's (start-stop), cluster the sequences using mmseq2, in a length-tiered fashion. Group clusters member by location (contig + pos), compute presence (relative coverage of the element vs. full coverage). Deletion signal is inverted to reflect presence/absence similarly to insertions. 

In the final output files, every matrix cell means the same thing on a 0-1 scale: **what fraction of reads at this locus, in this sample, carry the element**. INS-dominant loci report that frequency directly; DEL-dominant loci flip the DEL allele frequency (and fall back to a 50%-detection auxiliary-coverage check for samples without any DEL event).

## CLI

```
anvi-report-indels -c CONTIGS.db \
                   -p PROFILE.db [-p PROFILE.db ...]   # OR --profile-dbs-file FILE
                   -o OUTPUT_DIR
                   [--min-length 500]
                   [--max-length INT]
                   [--event-type {INS,DEL,both}]
                   [--min-identity FLOAT]               # overrides length-tiered defaults
                   [--min-coverage 0.8]                 # mmseqs2 -c
                   [--position-tolerance 50]            # bp; merges nearby events into one locus
                   [--min-samples 2]                    # presence-based: ≥N samples with non-zero matrix cell
                   [--min-reads-per-element 5]          # filters low-occurrence clusters
                   [-T NUM_THREADS]
                   [--just-do-it]
```

## Some design decisions

- **INS + DEL clustered together** into one element family. Same MGE picked up as an insertion in some samples and a deletion in others should collapse into one cluster, since the underlying sequence is identical.
- **DEL coverage** comes straight from `anvi-profile`'s existing calculation (`mean(coverage_array[pos:pos+length]) + count_of_DEL_reads`), unchanged.
- **`--min-samples` is presence-based.** It now counts samples with a non-zero matrix cell (so INS-only and DEL-only loci are filtered symmetrically), not samples that contributed an indel event.
- **Element presence at a DEL locus for a sample with no event** is decided by an auxiliary-data per-nt coverage check over the deleted region: ≥50% of positions with ≥1× coverage → element present.
- **Length-tiered clustering** by default (0.95 ID for 500 bp-1 kb, 0.90 for 1-5 kb, 0.85 for >5 kb). Short sequences below mmseqs2's 15 bp k-mer threshold fall back to exact-match clustering.

## New files

```
anvio/cli/report_indels.py
anvio/indels.py
anvio/drivers/mmseqs2.py
anvio/docs/programs/anvi-report-indels.md
anvio/docs/images/png/anvi-report-indels-01.png

anvio/tests/run_component_tests_for_report_indels.sh
anvio/tests/sandbox/mock_data_for_indels/source.fa             ← 138 kb real bacterial fragment
anvio/tests/sandbox/mock_data_for_indels/transposase_coords.txt
anvio/tests/sandbox/mock_data_for_indels/variants.txt
anvio/tests/sandbox/mock_data_for_indels/build_variants.py
```

## Modified files

```
pyproject.toml      ← anvi-report-indels entry point
```

## Test plan

- `pip install -e .` and confirm `anvi-report-indels --help` resolves.
- Run the new component test:
      `bash anvio/tests/run_component_tests_for_report_indels.sh anvio/tests/sandbox/test-output '' 4`
- Confirm `INDELS_DEFAULT/INDELS-VIEW-DATA.txt` shows the 3 expected loci (one DEL locus with element present in S01 only, two INS loci with the element present in S02 and S03 respectively) all in one cluster.
- Confirm `INDELS_DEFAULT/INDELS-CLUSTERS.fa` has one entry, length 435, `cluster_type=mixed`.
- Confirm the headline `HEADS UP: SOFT-CLIP ARTIFACTS IN OLDER PROFILE-DBS` warning shows + each profile-db's creation_date is printed.
- Verify `anvi-interactive --manual-mode --dry-run -p .../manual-profile.db --view-data ... --additional-layers ... --tree ...` accepts the outputs.
- Optional: re-profile a pre-PR BAM with this branch's `anvi-profile` and confirm no `INS` rows whose length matches a soft-clip length appear in the indels table any more.

## How to test on your own data

```bash
# any merged or single-sample profile-db with indels already profiled works
anvi-report-indels -c CONTIGS.db -p PROFILE.db -o MGE_OUTPUT

# the program prints the exact anvi-interactive --manual-mode line at the end
```

## Things explicitly **not** in this PR

- No changes to how `anvi-profile` records insertion sequences themselves (storage format, deduplication-by-unique-sequence behavior, etc.).
- No new database tables. All outputs are flat files.
- No new artifact docs under `anvio/docs/artifacts/`; the program doc references existing artifacts only.

# Also, we need another name for that command
It does what it does, it reports indels, but also their location / activity (relative abundance per location in metagenomes). The main use case for this command is to track MGE, but to call it report-mge would be making a decision on behalf of the user about the nature of these elements. Also they may think about homology-based MGE search (transposase HMM, etc.).

Also, as soon as the #2645 PR about clipping profiling is also merged into master, I am planning on incorporating the clipped sequence into these 'elements' cluster and track their distribution (loci) across metagenomes. So it is not about reported indels only. I'll start thinking about different names.


FlorianTrigodet:
Better names could be anvi-track-mobile-element or anvi-profile-mobile-element.
Notice it does not say mobile genetic element (MGE), but mobile elements, which is what the tool does: define and track sequences that appear/disappear at loci.

Ge0rges:
mobile-sequence instead of mobile element would put more mental distance with mobile genetic element. Just a thought.

Ge0rges:
Florian, you should make the INDELS.fa file have nice deflines so I don't need to reformat the FASTA to run `anvi-gen-contigs` :(

Would also be nice to have a "number of loci" field so that we can sort by that number in the interface (e.g. I want to look at MGEs that have many loci :>)

Edit: I forgot I had asked for number of loci and ran into wanting this again. Note to self to stop being lazy and implement it.

Ge0rges:
Somehow, the same cluster ID leads to different cluster median sizes?
Screenshot 2026-05-26 at 23 09 27

Screenshot 2026-05-26 at 23 09 23

Ge0rges:
While I'm here. When an INDEL occurs in more than one loci, it would be nice to have a way to look at each of those (perhaps anvi-inspect with a little picker wheel where we can pick the contig we want to look at?). 

While I'm again adding to the wishlist, it would also be cool for the table to contain whether the sequence has terminal inverted repeats perhaps? (Or is that too close to `anvi-report-inversions`?)

PR #2667, opened 2026-05-21
Snakemake Modularization and Harmonization

Kekananen:
The workflows were initially hard to maintain for developers as well as not as easy to add additional workflows to. The workflow passes the component test for each workflow in the `anvio/tests/` directory.

Related to issue #2611 and #2606.

General Change Summary:
- Split large Snakefiles into rule modules                                                                                                                                                       
      - Example: `metagenomics/Snakefile` now includes files like `rules/assembly_short_reads.smk`, `rules/read_filtering.smk`, `rules/taxonomy.smk`.                                                       
      - makes workflows easier to review, test, and extend without editing one massive file.                                                                                          
- Added reusable workflow helper modules                                                                                                                                                         
      - Example: `anvio/workflows/config.py`, `paths.py`, `registry.py`, `snakemake_utils.py`.                                                                                                              
      - moves pure helpers out of `__init__.py`, reducing clutter and making ownership clearer.                                                                                         
- Kept old anvio.workflows as w API working                                                                                                                                                      
      - Example: Snakefiles can still call `w.get_workflow_rule_file_path(...)`.                                                                                                                      
      - existing workflow code does not need broad rewrites while internals become cleaner.                                                                                           
- Added params.json schemas                                                                                                                                                                      
      - Example: `anvio/workflows/metagenomics/params.json` declares accepted params, types, and defaults.                                                                                            
      - workflow configuration becomes data-driven instead of scattered through Python classes.                                                                                       
- Default configs now come from schema                                                                                                                                                           
      - Example: `get_default_config()` reads defaults from params.json.                                                                                                                              
      - reduces duplication between “allowed params” and “default params.”                                                                                                            
- Added config type validation                                                                                                                                                                   
      - Example: "threads": "many" now fails early if schema expects int.                                                                                                                           
      - users get clear config errors before Snakemake builds a DAG.                                                                                                                  
- Added inherited schema merging                                                                                                                                                                 
      - Example: metagenomics can inherit contigs/read-recruitment-style params while adding its own.                                                                                               
      - shared rules can travel with shared config contracts.  
- Removed duplicated config dictionaries from workflow classes                                                                                                                                   
      - Example: contigs params moved from `ContigsDBWorkflow.__init__()` into `contigs/params.json`.                                                                                                   
      - reduces drift between implementation, defaults, and validation.                                                                                                               
- Added reusable read-recruitment module                                                                                                                                                         
      - Example: `anvio/workflows/read_recruitment/rules/main.smk`.                                                                                                                                   
[...]

PR #2625, opened 2026-05-20
Add Python 3.11 support

ulgens:
Relates to #2618

* Declare support for Python 3.11
* Updates `pandas==1.4.4` to `==1.5.2` (the earliest version with Python 3.11 support.)

Blocked by #2642 and the fix that is going to follow it.

Issue #2618, opened 2026-05-19
[TECHNICAL ISSUE] anvi'o fixed to Python 3.10. 3.10 will not be patched from late 2026


bcoltman:
## Short description of the problem

anvi'o python version is fixed at 3.10. Need to migrate to a newer python version to ensure security patches.

It will cause wide ranging issues, likely not with the python code base itself, but with the dependencies which will also need to be updated.

meren:
This is something @ulgens also mentioned a few times, and I am fully on board with this idea :)

bcoltman:
Hey @ulgens - @meren mentioned that you had been talking about this. Have you started looking into it? 

I have started on a PR that will migrate anvi'o to Python 3.14 and am looking at the impact on dependencies etc. I'm currently going through the testing phase and resolving issues as they pop up. Big changes required!

I'll ping you when I make a PR and would appreciate your input, at that stage or before :)

ulgens:
Hey all,

I worked on this a bit and it seems we don't need to do much for adding >3.10 support. I was trying to cut the workflow runtime a bit before adding new entries to the matrix, aiming to keep the resource usage minimum but I guess it can be ignored for a single entry. 3.11 support PR is here: https://github.com/merenlab/anvio/pull/2625

I think >3.11 can be supported in the same way, but I feel like this is more about the users than the repo itself. I don't know if anvio users are updating their Python versions periodically or are willing to do so.

ulgens:
@bcoltman It seems we need to update pandas first: https://github.com/merenlab/anvio/actions/runs/26157183943/job/76939484132 If you are able to create a PR for that, I can help with the review. @meren Is there anything specific we need to be careful about a pandas update?

bcoltman:
@ulgens on it! I'm working through the testing workflows and finding other issues with other dependencies too. I will ping you when the PR is there

bcoltman:
@ulgens 
I'll add that I'm actually working on a jump to 3.14. It probably would have been wise to jump up a version at a time but I decided to go all in, accepting that it's not going to be possible to support all python versions between 3.10 and 3.14, largely because of different dependencies not being compatible with all of those version.

That's to say that the PR will be different and not cover the pandas bump required above ^.

I'll look into the history of why that pandas version was fixed and whether a smaller fix can allow for the version bump to 3.11. No guarantees when

ulgens:
[...]

Issue #2234, opened 2024-03-01
[BUG] Eggnog mapper not running in workflow, works fine manually in same conda env


timternetnet:
## Short description of the problem

I'm trying to include Eggnog mapper in an Anvio workflow, but it stops with a non descriptive error. No clue if it's an issue with my workflow of with my installation.

## anvi'o version

```
anvi-self-test --version
Anvi'o .......................................: marie (v8)
Python .......................................: 3.10.13

Profile database .............................: 38
Contigs database .............................: 21
Pan database .................................: 16
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 4
tRNA-seq database ............................: 2
```

## System info

```
Ubuntu 22.04
```

Installation:
```
Follow https://anvio.org/install/linux/stable/ (conda install)

anvi-setup-scg-taxonomy --scgs-taxonomy-data-dir /Taxonomy/802/anvio-8 -T 20 --reset
anvi-setup-ncbi-cogs --cog-data-dir  /Taxonomy/802/anvio-8 -T 20
anvi-setup-kegg-data --kegg-data-dir /Taxonomy/802/anvio-8/KEGG -T 20

conda install pulp=2.7.0

conda install -c bioconda krakenuniq
root@bw25c802:/Taxonomy/802/krakenuniq# krakenuniq-download --db DB --taxa "archaea,bacteria,viral,fungi,protozoa" --dust --exclude-environmental-taxa microbial-nt --threads 32
krakenuniq-build --db DB --kmer-len 31 --threads 30 --taxids-for-genomes --taxids-for-sequences --jellyfish-bin $(type -P -a jellyfish) --jellyfish-hash-size 30000000000

conda install -c bioconda -c conda-forge eggnog-mapper
download_eggnog_data.py --data_dir /Taxonomy/802/anvio-8-eggnogmapper

conda install bioconda::metabat2
```
Result:
```
conda list -n anvio-8
# packages in environment at /usr/local/bin/miniconda3/envs/anvio-8:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
_r-mutex                  1.0.1               anacondar_1    conda-forge
_sysroot_linux-64_curr_repodata_hack 3                   h69a702a_14    conda-forge
amply                     0.1.6              pyhd8ed1ab_0    conda-forge
[...]

PR #2681, opened 2026-05-24
Port anvi'o to Python 3.13

bcoltman:
Well, despite a frustrating week, I am glad to make this PR, which moves anvi'o to Python 3.13. This relates to #2618 and the PRs of @ulgens #2625 and #2628 

This comes with a bump in the versions of several core dependencies, so I apologise in advance: this is a large PR.
However, it should put anvi'o in a stronger position for the next few years, and it deals with some inevitable updates to
libraries such as pandas, NumPy, scikit-learn, and Snakemake. Are all of the changes purely due to the Python 3.13 bump?
Maybe not. But they were the changes needed to get the codebase installed and to complete the tests defined by `anvi-self-test`.

## Background

For many features of anvi'o to function, it requires external programs that are not distributed by PyPI, such as MUSCLE,
HMMER, DIAMOND, MEME, and others. So the practical fully functional anvi'o install remains a two-step process:

1. create a conda environment containing Python and the non-PyPI external programs;
2. run `pip install -e .` from the anvi'o repository, which installs the Python dependencies defined in `pyproject.toml`.

This mixed install model creates some awkwardness. Not all external programs are needed by all users, and not all anvi'o
features require the full external stack. But for this port I wanted to make sure the main codebase and the `anvi-self-
test` matrix were functional in a Python 3.13 environment.

## PLEASE WORK!

The conda environment, which we can install stuff in was created as follows:
```
mamba create -n anvio_not_like_13 conda-forge::python=3.13
mamba activate anvio_not_like_13
mamba install -c conda-forge -c bioconda "python=3.13" \
	sqlite prodigal idba mcl "muscle=3.8.1551" famsa hmmer diamond \
	blast megahit spades bowtie2 bwa graphviz samtools \
	trimal iqtree trnascan-se fasttree vmatch r-base r-tidyverse \
	r-optparse r-stringi r-magrittr bioconductor-qvalue meme ghostscript \
	nodejs llvmlite numba fastani \
	flye minimap2 sra-tools mmseqs2 # these are found in the codebase but are missing from the install pages
```

flye, minimap2, sra-tools, and mmseqs2 are used by code paths in the repository but were missing from the install-page/environment guidance I was working from.

The old exact sqlite and nodejs pins were relaxed. Those exact pins were introduced in 30334b841
(https://github.com/merenlab/anvio/commit/30334b841), but I could not find a functional reason for keeping those specific old versions. Keeping them pinned also constrained the Python 3.13 patch version unnecessarily. However, do we need to fix nodejs or sqlite?

MEME remains pinned/constrained, following [a14cc6b98](https://github.com/merenlab/anvio/commit/a14cc6b98). However, this is anyway the most recent version. I have submitted a MEME rebuild, which has bumped the [base ICU build to 75](https://github.com/bioconda/bioconda-recipes/pull/65732), which might help us with one or two extra updates too. 

## Dependency changes

This PR locks anvi'o to Python `3.13.*` in both package metadata and the runtime import guard.

Important dependency changes include:

- `pandas>=3,<4`
- `numpy>=2.1`
- `scikit-learn>=1.8,<1.9` makes PR #2459 pointless
- `matplotlib>=3.9`
- `snakemake>=9,<10`
- `legacy-cgi` added for `ete3`
- `pyani` removed
- `numba`, `PyYAML`, and `seaborn` declared as runtime dependencies

`numba` was already required by runtime code paths such as palindrome searching and variability/BAM utilities, but was not declared in `pyproject.toml`. This also meant the master component-test Action could pass while only emitting a missing-`numba` warning. Declaring it makes the CI workflow actually exercise the requested component-test suites.

`pyani` was dropped in bc1bdb531 (https://github.com/bcoltman/anvio/commit/bc1bdb531) because it is no longer maintained, and its developers have moved to [pyani-plus](https://github.com/pyani-plus/pyani-plus). Unfortunately, pyani-plus is not a simple drop-in replacement for the current anvi'o integration. So this PR removes 1pyani1 from the dependency set and accepts that existing pyANI-specific calls are unsupported for now. I will raise a separate issue/PR for replacing that functionality properly.

There is also previous anvi'o history showing that pyANI required `matplotlib==3.5.1` to keep working (6f4c7553d
(https://github.com/merenlab/anvio/commit/6f4c7553d)). That matplotlib version is not suitable for this Python 3.13 port.

The bundled SCG classifier was regenerated for `scikit-learn 1.8`, so `scikit-learn` is constrained to the 1.8 series.

Snakemake 7 is not installable for this Python 3.13 port, so workflows are now pinned to Snakemake 9. The existing `pulp== 2.7.0` pin looks related to Snakemake compatibility; see snakemake/snakemake#2606 (https://github.com/snakemake/snakemake/issues/2606) and the anvi'o workflow discussion in merenlab/anvio#2234 (https://github.com/merenlab/anvio/issues/2234).

## Python 3.13 compatibility

The branch handles the main Python 3.13 issues we looked for:

- Python metadata and import-time checks now target 3.13;
- removed stdlib `cgi` is handled through `legacy-cgi`, needed because `ete3` still imports it;
- no direct problematic `distutils` or removed stdlib imports were found in anvi'o runtime code;
- shared-memory worker attachments now use `SharedMemory(..., track=False)` where the worker does not own the lifecycle.

That last change removes benign but ugly `resource_tracker` tracebacks seen during threaded runs

## pandas / NumPy / scientific stack updates

A lot of this PR is adapting anvi'o to current pandas and NumPy behavior. These changes are not just to make tests pass but they remove uses of APIs that are gone or stricter now.

Examples include:

- replacing `DataFrame.append`;
- replacing `Series.iteritems`;
- replacing `DataFrame.applymap`;
- replacing `delim_whitespace`;
- avoiding set-like pandas indexers;
- avoiding fragile chained/mixed row selection;
- using positional indexing where integer Series keys were ambiguous;
- replacing `np.float_`;
- making NumPy blob conversion more explicit.

One important behavioral fix is in gene-level coverage data: `non_outlier_positions` are now read and written as boolean masks, rather than accidentally using coverage blobs. This fixes the MCG classifier overflow warning/failure in a way that I think matches the intended data model.

The tRNA anticodon change is intentionally minimal but came from a pandas bump causing different behaviour. Still, missing anticodons remain blank strings, matching the previous behavior.

## Workflow / Snakemake 9 changes

@Kekananen and related to #2667 
These changes do not include or account for those implemented by @Kekananen in #2667. However, I believe that they can be relatively easily merged.

Snakemake 9 works by comapring the times of input and output files to determine whether rules should be re-run. This caused issues when inputs like PROFILE.db were modified.
The workflow updates include:

- using `snakemake.cli.main`;
- removing unsupported rule `version:` directives;
- updating included-Snakefile detection;
- removing the old Snakemake log-handler hook;
- handling fast-rule mtime skew seen under Snakemake 9;
- making mapped-read layer imports rerunnable for mutable `PROFILE.db` outputs.

`PROFILE.db` is mutable and not a reliable freshness signal, so the rule treats it as ancient input and allows replacement of workflow-owned layer keys during resume/rerun.

## Self-test and fixture updates

Several self-test scripts were updated so they behave properly under unattended Python 3.13 runs:

- `--no-interactive` is now respected by more display commands;
- pangenome graph tests no longer pass `--num-threads` to a command that does not support it;
- YAML pan graph display is dry-run aware;
- metagenomics gene-mode misc-data now uses real gene IDs from the generated test DB;
- ecophylo read generation no longer depends on a local external script path;
- workflow test configs were reduced where tiny fixtures do not need high thread or memory use;
- PFAM HMM helper calls now use the current -o/--output-dir flag rather than the old -O form (163299d2a
  (https://github.com/bcoltman/anvio/commit/163299d2a)).

## CI / component tests

The GitHub workflow named `Component Tests and Migrations` now completes successfully on this branch:

- successful current run: https://github.com/bcoltman/anvio/actions/runs/26375147653
- tested commit: [`a75a78f`](https://github.com/bcoltman/anvio/commit/a75a78f7f02526aa393045da848fafaccc689cb5)

This is worth pointing out as the current `master` workflow can appear green without actually running the component tests. The first retained affected `master` run I found was on 2026-05-13:

- affected `master` run: https://github.com/merenlab/anvio/actions/runs/25791989027
- tested commit: [`c8836817`](https://github.com/merenlab/anvio/commit/c8836817290081a8639e31e3370b831f8fd864dd)

That run completed successfully, but the `Run component tests` step lasted only about four seconds. Each `anvi-self-test` command exited after reporting that `numba` could not be imported, so the component suites did not actually run. The previous successful `master` run, from 2026-05-12, did run the component tests properly and spent about 9.5 minutes in that step:

- previous real `master` run: https://github.com/merenlab/anvio/actions/runs/25727288574
- tested commit: [`01d2b68`](https://github.com/merenlab/anvio/commit/01d2b68fcf36686fc7e6f9fb9b70d340a465cd02)

`numba` was removed on the pangenome_graph branch in fbf4ca562 on 2026-03-31, and entered master when PR #2603 was merged in c6b59ed41 on 2026-05-12. The first affected scheduled master component-test run was the following day, 2026-05-13. This PR declares `numba` again as a runtime dependency, alongside other Python packages that are imported by the codebase but were missing from the dependency list. 

Update: It is currently being addressed by this PR #2641

One small issue with the latest component-test log on this branch is from `pulp`, which is outside the anvi'o codebase and with one or two SyntaxWarnings. Doesn't appear to break the code.

## Test status

Apart from `databse-migrations` and `workflow-sra-download`, all suites completed successfully:

`database-migrations`

This exits early around very old h5py handling. The branch avoids forcing an incompatible ancient build and skips that HDF5-specific path when unavailable.

`workflow-sra-download` 
This fails with the current old sra-tools/TLS stack. Updating that toolchain currently conflicts with the meme/ICU constraints, and sra-tools is not used by the main anvi'o runtime. The MEME bump will mean that a higher sra-tools version can be downloaded; however, it isn't really necessary for anvi'o as no other script apart from this workflow uses it. 

## Follow-up work

The main follow-ups are:

- proper replacement of the old pyANI integration, likely via pyani-plus or another supported route;
- a separate decision on the SRA workflow/toolchain. The situation may improve once the MEME rebuild on bioconda goes through.
- optional cleanup of ecophylo SyntaxWarning: invalid escape sequence '\|' warnings, which do not block execution;
- future work to support newer MUSCLE versions rather than relying on MUSCLE 3 behavior.
- update networkX to a higher version and replace the deprecated Edmonds class

And here we are.


meren:
Dear @bcoltman, first of all,

image

Holy moly :)

Thank you so much. This is a lot of work and trial-and-error. And thank you for this summary of changes and insights into the decision making process, it helped a lot when I started going through some of the changes.

A few ideas I'd like to share:

* The awkwardness that stems from "not all external programs are needed by all users, and not all anvi'o features require the full external stack" is real, and I think requiring large conda environments with a gazillion of programs is a no-go in the long term. This is something @telatin also recognized, and hopefully we will fix it at some point. If you have time, please look at the discussion at #2677, especially the 'tool registry' idea that stemmed from it. If we pull this off, then neither pyANI nor any other third-party tool for that matter will be an issue, because we will simply work with them through specific conda environments or singularity/docker containers. Just for food for thought at this stage.

* I see a lot of changes in snakemake files, given that there is a complete overhaul of snakemake workflows at #2667 by @Kekananen, @btrouche, @katysloz, and @l-gallucci, and I am 100% there will be significant conflicts here. Which scares me that we are going to be the victims of our own productivity :) I haven't had a chance to review #2667 yet, but I am hoping to get to it soon. But when I do, I will have to make a decision: either I merge that branch, and then it will be necessary to revisit this branch, or I will merge this branch, and it will be necessary to revisit that branch. Given that #2681 is a comprehensive change that covers the entirety of the repository and that #2667 is one that is relatively specific to a single subsystem with a very large number of changes, I feel that merging #2667 first and revisiting this branch later would be a better call (i.e., merge the more narrowly scoped, self-contained change first, then rebase the broader one on top kind of logic). I think this would be the best practice from a software engineering perspective, but I don't want to make a suggestion without your input, @bcoltman. What do you think?

* `anvio/mcgclassifier.py` is a dead project at this point. It was a great idea at the time, but we couldn't finish it. I think I will kill it finally. Just a self note when I'm back to testing this branch.

* There is one important concern that is not a part of our regular component testing: migrations of very old anvi'o databases. Anvi'o is able to migrate databases from all the way back 2015. Generated then, and can be migrated and used with the most current set of tools in anvi'o (see #2682 from a few moments ago). I think it is extremely for us to be able to maintain that and claim we maintain that level of reproducibility and data provenance.

  Migrations don't run via CI because they can take a long time, and the amount of data download can be prohibitive. But it would be great if you could manually check if they are working with your branch (which is what I will do when the time comes, but if it reveals any bugs you probably would immediately know what is the issue when everything is fresh). If you would like to give it a go, you can simply go to your Downloads directory, and run this (the `-o` gives you a directory that you can re-try things if a failure has occurred):

  ```
  anvi-self-test --suite database-migrations -o MIGRATIONS
  ```

  In that note, we probably should change the name of [our component tests script](https://github.com/merenlab/anvio/blob/master/.github/workflows/component-tests-and-migrations.yaml) from "Component Tests and Migrations" to "Component Tests". We have the migrations commented out at the end due to resource limitations, for which I don't have a good solution that will not cost a fortune we can't afford and I don't want to ping @ulgens here and thereby pull him into this very long message just to ask his opinion on this very specific concern (lol).

  This is a winded way to say that we need to find a solution to your issue with HD5, as we can't simply skip that HDF5-specific path when unavailable (because that would make migration incomplete). Is this happening because `pip install h5py==3.9.0` line at the top of `anvio/tests/run_migration_tests_for_ancient_anvio_databases.sh` incompatible with Python 3.13?

---

Apart from these notes, here is my 2 cents on the **follow-up work** you mentioned:

> proper replacement of the old pyANI integration, likely via pyani-plus or another supported route;

Yes. Or we can even use pyANI with singularity/docker solution being brewed in #2677 to test that system in action.

> a separate decision on the SRA workflow/toolchain

This is important and would likely require input from @FlorianTrigodet. But only after #2667 is resolved.

> optional cleanup of ecophylo SyntaxWarning: invalid escape sequence '|' warnings, which do not block execution;

This is also important, but may be an out-of-date concern given #2667.

> future work to support newer MUSCLE versions rather than relying on MUSCLE 3 behavior.

Yes. This has to be done. Or may be something to address via #2677 again. 

> update networkX to a higher version and replace the deprecated Edmonds class

😬

This will give @ahenoch a heart attack. But it is more straightforward than it sounds as it seems the replacements to Edmonds have already been the standalone module-level functions that were available and internally used Edmonds under the hood. So, all needs to happen is to not do this,

```python
from networkx.algorithms.tree.branchings import Edmonds
ed = Edmonds(G)
result = ed.find_optimum(attr, default, kind='max', style='arborescence')
```

And do this instead with newer versions of NetworkX:

```python
import networkx as nx

nx.minimum_spanning_arborescence(G)
nx.maximum_spanning_arborescence(G)
nx.minimum_branching(G)
nx.maximum_branching(G)
```

But @ahenoch is the expert here and he would have to make sure the solution yields identical outcomes for our pangraphs.

ulgens:
Holy moly indeed. I have a couple of points in mind, will comment in the next hours.

ulgens:
I won't be able to comment for everything but let me share what I have in mind & I think is important:

- Right now, the CI is failing silently. I believe, whatever the decision is on the upgrade, this is the first thing we should fix. #2642 makes it fail loudly. When we merge it, I'll try to fix the actual error.
- I'm not feeling good about switching from 3.10 to 3.13 directly. I was adding support for multiple versions in #2625, so that the next version won't disrupt every user's existing setup.
- The PR has several parts that can be seperated into different PRs, like file ending fixes and `r""` conversions. I think they should be handled separately.
  - In any case, I think this PR will need to be updated multiple times. "merge main into the branch" approach complicates the history and makes it harder to understand the changes, especially when the PR is this big. I strongly recommend getting comfortable with rebase.
- I don't have anything to add about the tests we are skipping, I don't know which type of resource they are exhausting. I'll check them after fixing the CI.

meren:
Hey @ulgens,

Thanks for your input here. Let me respond to a few things:

> * Right now, the CI is failing silently. I believe, whatever the decision is on the upgrade, this is the first thing we should fix. [Exit with status code 1 in case of an error - fatal warning #2642](https://github.com/merenlab/anvio/pull/2642) makes it fail loudly. When we merge it, I'll try to fix the actual error.

It was totally my fault. Sorry about leaving that convo with a vague response -- last week was so hectic, things fell through the cracks. I merged it now :)

> * I'm not feeling good about switching from 3.10 to 3.13 directly. I was adding support for multiple versions in [Add Python 3.11 support #2625](https://github.com/merenlab/anvio/pull/2625), so that the next version won't disrupt every user's existing setup.

Your approach is the right way to do it, but I don't think it is necessary for platforms like anvi'o. Maybe this will sound like a boomer concern, but it is best for us to be extremely explicit with Python versions and use a single one so when people come to us with weird errors, we wouldn't also have to think about whether this could be their local Python version. Since we are not full time software developers and usually have minutes every day to help our users, anchoring ourselves to a single place in Python history gained us a lot of time in the past that I'm scared to lose and end up disappointing people who would not have cared to have freedom in their Python version of their conda environment for anvi'o. 

>   * In any case, I think this PR will need to be updated multiple times. "merge main into the branch" approach complicates the history and makes it harder to understand the changes, especially when the PR is this big. I strongly recommend getting comfortable with rebase.

I hear you, but I find rebase very scary as our branches often have multiple contributors and rebase rewrites every commit hash on the branch and requires a force-push .. which silently invalidates everyone else's local copy. Recovering from that requires manual intervention that people may not even realize they need to do, and in the past we literally lost commits that we didn't realize until it was too late. Merge commit is boring but safe (as in nobody's local state breaks, nobody needs to do anything special). The only thing rebase buys us is a cleaner history, which I sometimes feel isn't worth the risk. Sorry if I'm missing something here (which may be the case really).

> * I don't have anything to add about the tests we are skipping, I don't know which type of resource they are exhausting. I'll check them after fixing the CI.

They mostly exhaust storage :/ No need to pay any attention to it.

bcoltman:
@meren and @ulgens - thanks both for your feedback and again, I apologise about the large PR. As software engineers, I'm curious how you think this PR could have been better structured? With the goal of bumping to 3.13 in mind, I felt like so many bugs popped up that it was hard to disentangle how they should be separated. I would be curious for your input as somehow setting up multiple PRs with only the final one allowing for a build/test completion just didn't seem right for me, but I am keen to learn! Additionally, I apologise but I've made some more commits since the original PR for more Pandas 3 related issues prompted by the migration script testing. Details below

---

### Snakemake 

Sure, I'm happy for the Snakemake fixes from @Kekananen to be merged, and then I can base the changes in this PR off of the new state!

---
### Migration

Migration works! I've just checked again and the reason why I skipped it was because of the pinned h5py version. The old database-migration path needs `h5py` because some historical anvi'o databases used `.h5` auxiliary files. Trying to install `h5py==3.9.0` under Python 3.13 does not work cleanly because that release does not provide Python 3.13 wheels. 

Installing current `h5py` works normally under Python 3.13. Pip selected `h5py 3.16.0`, which has a compatible wheel. With modern `h5py` installed, `anvi-self-test --suite database-migrations` got past the previous import/availability blocker and successfully migrated the legacy HDF5 files. Another issue popped up later in the script due to the pandas 3 bump which is related to missing taxonomy ranks previously returning None but now returning as NaN. We have to convert these missing values back to None for taxonomy rank fields. This fix prompted me to pass through the codebase again to find any related issues (see below). Also, just to confirm, this run included your changes from 6beca1d. 

---
### Pandas related fixes

Pandas 3 handles NA differently. Values that previously behaved like strings, including questionable-but-possible identifiers such as "NA", can become missing values.  The fixes here keep old behaviour for identifier-like fields while avoiding broad changes to numeric parsing. Although the tests had passed, I had missed that some NA related issues were occurring and bleeding through as warnings. I(/Codex) have audited the code base to search for any of these cases and added some fixes.
 
---
### NetworkX

Yes, it seems straightforward. An easy solution popped up in testing and I will open it as a separate PR. I believe it will be compatible both with the currently pinned `3.1` networkx version and future ones as the more generic `nx.maximum_spanning_arborescence(G)` API was already defined by then. 

--- 

### Muscle

I have a MUSCLE 5 compatible fix in another branch from one of the failed attempts to get to this PR. I will find it and open it as a separate issue/PR.

bcoltman:
New results from component tests [here](https://github.com/bcoltman/anvio/actions/runs/26405921316/job/77729114991)

Kekananen:
@bcoltman @FlorianTrigodet If there are any major issues with the snakemake workflow merge and the python 3.13 update PR, just let me know and I can see about jumping in as well to help get it resolved from that end too.

meren:
> @meren and @ulgens - thanks both for your feedback and again, I apologise about the large PR. As software engineers, I'm curious how you think this PR could have been better structured?

Here is my 2 cents on this, @bcoltman.

The easiest and most intuitive way is to divide changes into two major groups:

**(1)** Changes that serve the end goal of the branch but are also compatible with the current `master`, such as:

```diff
%(anvi-script-pfam-accessions-to-hmms-directory)s --pfam-accessions-list PF06603.14 \
-                                                 -O METABOLISM_HMM
+                                                 -o METABOLISM_HMM
```

or

```diff
        Returns
        =======
-       N\A
+       N/A
        """
```

or

```diff
                # module
                elif first_char == "D":
-                   fields = re.split('\s{2,}', line)
+                   fields = re.split(r'\s{2,}', line)
                    mnum = fields[1]
```

etc.

**(2)** Changes that are only compatible with or relevant to Python 3.13, such as:

```diff
 dependencies:
- - python=3.10
- - sqlite=3.46
+ - python=3.13
+ - pip
+ - sqlite
 - prodigal
```

The first group can then be broken into even more atomic PRs with a clear purposes to be merged into `master` one by one.

For instance, one PR could be titled "Adhere to Python's handling of invalid escape sequences in regular strings" and would cover the `'\s{2,}'` → `r'\s{2,}'` like changes (which are already best practice under Python 3.10), another could be titled "Remove unused version strings from Snakemake workflows", and so on.

This approach has two benefits: it trims the Python 3.13 branch down to only the changes that truly require it, and it lets those who will review thigns evaluate and merge each backward-compatible improvement independently against `master` (where they already belong regardless of the migration, but weren't because anvi'o developers missed them). It is like writing a methods section for a new and fancy bioinformatics tool (where the novel algorithm has its own glorified section at the beginning, and then all the other boring stuff end up in their own littler sections), but backwards, so the boring stuff comes first :p

I'd say separating backward-compatible improvements from mission-specific changes is generally the best practice in all large changes of this sort. But [even the best of us](https://lkml.iu.edu/hypermail/linux/kernel/0610.2/2867.html) can fail at that :p (David Miller has been one of the absolute top developers of the Linux kernel) (because of the hackathon, my brain completely switched and is pulling crazy random memories from my days in open source software .. I'm scared that it will be hard to switch back to science after all this).

Sorry for the long message, and thank you both for the effort and ideas. You are all the best.

bcoltman:
Thanks @meren,

I appreciate you taking the time to think and share your insights and guidance. It certainly makes sense.

Considering that, would you prefer for this PR to be revised and split up? I'm not sure what the easiest way to achieve that is, but I'm sure there is a way. If @ulgens has any ideas too then I'm all ears.

ulgens:
Found and fixed [the issue](https://github.com/merenlab/anvio/pull/2689) with CI. We have [one more issue](https://github.com/merenlab/anvio/actions/runs/26527330619/job/78134414742) to tackle and then I'll be checking how to split up this PR.

bcoltman:
Hi @ulgens ,
Sorry for the late response. This error popped up during my testing too. [163299d](https://github.com/merenlab/anvio/commit/163299d2a97205453367c27513afb5204f6b6b09) fixes it.

meren:
> Hi @ulgens , Sorry for the late response. This error popped up during my testing too. [163299d](https://github.com/merenlab/anvio/commit/163299d2a97205453367c27513afb5204f6b6b09) fixes it.

Hey @bcoltman, I need 163299d2a97205453367c27513afb5204f6b6b09 in master so the workflows no longer fail, but because they are in this branch there is no way to cherry-pick them :)

Would you be so kind and create another branch from `master`, cherry-pick https://github.com/merenlab/anvio/commit/163299d2a97205453367c27513afb5204f6b6b09 into your new branch, and send a PR for that?

I am tired of getting workflow failed errors every day :p

ulgens:
+1 on @meren 's last comment. When we have the CI fixed, I can check the tests we currently have disabled, and I believe it will give some sense of direction.

bcoltman:
Hi both, please see #2700 and #2701. They deal with different test suites, but I believe that both combined will allow for the workflow tests to complete. Waiting to see the results

bcoltman:
I'm happy to look at splitting up this PR. I would do it as follows, accepting that there are some interdependencies between some branches:

# Split Python 3.13 Port Into Focused, Labeled PRs

## Summary

- Keep python_313_port / #2681 as the source branch and discussion record.
- Create clean branches from upstream/master, but label each split as independent, stacked, or integration-only.
- Do not open a PR as “independent” if it knowingly breaks master alone. Tight dependencies should either be stacked explicitly or grouped into the same PR.
- Existing #2700 and #2701 remain the first extracted independent PRs.

## Split Plan

- Independent: fix-python-syntaxwarnings
    - Raw regex strings, invalid escapes, docstring typo-style fixes from b048722d8, 3e63fdbc5, a75a78f7f.
    - fix-pandas-removed-apis: removed parser/indexing/append/API changes.
    - fix-pandas-na-identifiers: missing-value and identifier preservation changes.
    - These may be thematically separate but should be tested both alone and together. If either fails alone, open the second stacked on the first or combine them into one pandas
      compatibility PR.
- Independent if possible, otherwise stacked: fix-ancient-migrations-h5py
    - Use modern h5py for ancient migrations from 463d85dc4.
    - Avoid the skip-based 548acc126 behavior unless rewritten.
    - If it only makes sense under Python 3.13, mark it as stacked on the final port branch instead.
- Stacked / carefully scoped: fix-self-test-compat
    - Extract remaining test fixes not already in #2700/#2701.
    - Split further if needed: pangenome graph tests, fastANI/pyANI test config, no-interactive/resource fixes.
    - PR body must state any dependency on pandas or dependency-env branches.
- Deferred: fix-snakemake-9-compat
    - Wait until #2667 is merged or abandoned.
    - Rebuild against the resulting workflow structure.
- Integration-only: python-313-port-v2
    - Final branch containing true Python 3.13 dependency/runtime/CI changes.
    - Depends on merged or stacked prior PRs.
    - Supersedes #2681 once ready.

## Interdependency Rules

- Every branch gets an explicit compatibility label in the branch notes and PR body.
- If a branch requires another split to pass tests, base it on that branch and say so in the PR title/body.
- If two branches cannot reasonably pass independently, combine them.
- PR bodies should include language like: “This PR is independently compatible with current master, but full Python 3.13 validation also requires #XXXX.”

## Test Plan

- For independent PRs: run conda run -n anvio-dev prek run --all-files and targeted component tests on the branch alone.
- For stacked PRs: run tests both on the stacked branch and on the combined stack.
- For pandas splits: specifically verify removed API behavior and missing-value/identifier preservation together.
- For migrations: run conda run -n anvio-dev anvi-self-test --suite database-migrations -o MIGRATIONS.
- For final Python 3.13 integration: run local equivalents of the GitHub component workflow, including metagenomics-full, pangenomics, and inversions.

## Assumptions

- upstream/master remains the source of truth.
- python_313_port is not rewritten while #2681 remains useful.
- Existing untracked logs in the worktree are preserved.
- Once extracted PRs land, duplicated commits are omitted from python-313-port-v2.

meren:
#2700 and #2701 are now merged.

FlorianTrigodet:
Hello there :)
The snakemake refactor (#2667) has been merged to master. Hope this help with this PR.

Kekananen:
Hey all! Mind if I jump into this and help out? I can try to break up this PR into the sections as @bcoltman described above and we can go from there?

bcoltman:
Thanks for the offer @Kekananen!  Please feel free to jump in. I won't get chance to look a this for another fortnight or so. Alternatively, I can look at it at the end of July/start of August and get it over the line then.

PR #2694, opened 2026-05-29
Gene highlight in inspect page

FlorianTrigodet:
When you search for a gene function in the interactive interface, you can highlight the splits containing genes with a matches. But when you inspect that split, you have to search among the ~20 genes for the one matching your search. ANNOYING.
Solution: highlight the genes:
image


FlorianTrigodet:
CodeFactor is not happy because of the complexity of the code. Now that we have these checks, what's the best course of action? It is out of scope to rewrite these methods in this PR. So I can just merge?

PR #2695, opened 2026-05-31
Dynamic viz params

TianyiChang:
This branch introduces dynamic, data-driven visual scaling for the anvi'o pangenome graph interactive display (anvi-display-pan). By replacing hard-coded values with values that scale with graph size, this update improves visualization consistency across a spectrum of genomic scales, from relatively large bacterial pangenomes to more compact mobile genetic element graphs.

**Comparison of UI figures before and after the changes**
A quick example
Screenshot 2026-06-01 at 16 50 23

Please see this file for all comparisons
[dynamic-viz-params-fig-examples.pdf](https://github.com/user-attachments/files/28460034/dynamic-viz-params-fig-examples.pdf)


anvio/panops.py
```diff
@@ -1821,40 +1821,100 @@ def sanity_check(self):
...
 
-        x_max = max([data['position'][0] for node, data in self.pangenome_graph.graph.nodes(data=True)])
-        y_max = max([data['position'][1] for node, data in self.pangenome_graph.graph.nodes(data=True)])
-        for i, j, data in self.pangenome_graph.graph.edges(data=True):
-            if data['route']:
+        # --- 1. Base Graph Metrics ---
+        num_core_nodes = sum(
+            1 for _, data in self.pangenome_graph.graph.nodes(data=True) 
+            if data.get('type') == 'core'
+        )
```
meren: I understand that you are defining `num_core_nodes` to scale things later and use it to determine how long the graph is from end to end.

But I don't think this is the right way to do it here for two reasons since in theory a pangeome graph may have no core nodes whatsoever, and core SynGCs will not give you the number to determine how long is the graph from end to end.

What you really want to use here is `x_max`. That is the variable that determines the visual horizontal span since its value is the maximum x-position assigned by @ahenoch's topological layout. 

The reason `num_core_nodes` rather than `x_max` hasn't exploded in your tests is because core SynGCs are not a terribly bad approximation to `x_max` :) But it doesn't make them right. So you should certainly use `x_max`, and get rid of `num_core_nodes`.

anvio/panops.py
```diff
@@ -2117,7 +2177,7 @@ def store_nodes_in_pan_graph_db(self):
         table_for_nodes.store()
 
         pan_graph_db = dbops.PanGraphDatabase(self.pan_graph_db_path, run=self.run, progress=self.progress, quiet=True,)
-        pan_graph_db.db.set_meta_value('num_nodes', len(self.pangenome_graph.graph.nodes()))
+        pan_graph_db.db.set_meta_value('num_core_nodes', len(self.pangenome_graph.graph.nodes()))
```
meren: The PR renames the metadata key from num_nodes to num_core_nodes. But the correct metavalue here is indeed `num_nodes`. Not only `len(self.pangenome_graph.graph.nodes())` does not return number of core nodes, we have to keep track of the total number of nodes in the `self` table. Please revert this change in your future iterations :)

anvio/panops.py
```diff
@@ -2497,7 +2557,7 @@ def remerge_nodes(self):
                          header="REMERGING SENSITIVE NODES", lc="green")
 
         gene_cluster_to_synteny_gene_cluster = {}
-        original_num_nodes = len(self.pangenome_graph.graph.nodes())
+        original_num_core_nodes = len(self.pangenome_graph.graph.nodes())
```
meren: Same goes here. The previous variable name is more accurate.

anvio/panops.py
```diff
@@ -2561,7 +2621,7 @@ def remerge_nodes(self):
                                     new_accessory_num += 1
 
         self.run.info_single("Successfully remerged nodes.")
-        self.run.info_single(f"{original_num_nodes - len(self.pangenome_graph.graph.nodes())} nodes were removed in the process.")
+        self.run.info_single(f"{original_num_core_nodes - len(self.pangenome_graph.graph.nodes())} nodes were removed in the process.")
```
meren: Likewise. The previous variable name is more accurate.

meren (changes requested):
Hey @TianyiChang,

Thank you very much for sending this in :) I have a few comments below regarding the code.

I also think it would help a lot if you were to include some systematic testing results (large genomes, mid-size genomes, small genomes, MGEs, etc) to show how things look like in the interface before and after the change. You can simply put screenshots into your PR and others (especially @ahenoch) could see how things change.

meren:
Hi @TianyiChang,

Thank you for these changes! Things are certainly looking better in the code side, and your example visualizations were very helpful to see before/after scenarios for multiple different size ranges.

I have a few observations I hope can help you push this forward just a bit more.

I think one problem is the width of edges. For instance, take a look at this before example -- quite elegant, and pleasing ratio between node sizes, node edge width, and edges between nodes:

image


After, a bit too much width I think:

image

The y-axis distances are also just a little too much I think. For instance, the region looks much better in before,

image

compared to after, I think:

image

And the y-axis effect increases with decreasing number of genomes I'm afraid. Here is a region before:

image

And after:

image

I know it is extremely difficult to nail these kinds of automated default setting determinations well, but I think the perfection is not too far :)

TianyiChang:
Hey @meren 

Thanks for the feedback! The edge widths and disty have now been scaled by the node radius, and here’s the updated comparison plot. Please let me know if you notice any further areas for improvement :)
[dynamic-viz-params-fig-examples_v2.pdf](https://github.com/user-attachments/files/28497164/dynamic-viz-params-fig-examples_v2.pdf)

meren:
These look great, @TianyiChang!

@ahenoch, would you please also look at the PR and merge it if you approve it? :)

meren (approved):

ahenoch:
Looks good to me! Great job, thank you @TianyiChang :)

TianyiChang:
Thank you, @ahenoch. Really glad that I can help :)

meren:
> [meren](https://github.com/meren) commented [on Jun 2](https://github.com/merenlab/anvio/pull/2695#issuecomment-4602000448)
> @ahenoch, would you please also look at the PR and merge it if you approve it? :)

@ahenoch, you never merged this branch, is it correct?

Please don't do it now -- I just wanted to make 100% sure I am not missing something.

ahenoch:
No it is not merged yet, github asked for conflict resolve and pull

meren:
This was the time to merge it, but I think you merge master into this branch rather than this branch into master, @ahenoch :)

image

ahenoch:
I merged master into the branch to keep it up to date after commenting, but I never merged it because I didn't want to steal the merge from someone, sorry...

meren:
@TianyiChang, I will take care of the conflicts and merge the final master into this branch, and then I will ask you to re-do your tests and confirm everything works. Is that OK?

TianyiChang:
Hi @meren. Of course! Please just let me know when it's ready to pull, and I'll carry out the tests using MGEs ASAP.

meren:
OK. Merged it in master. I will now merge it in Alex's larger rewrite effort and fix conflicts, and then I will ask you to test it in that branch when everything is ready. I will keep both @ahenoch and you posted, @TianyiChang.

meren:
They @TianyiChang, your branch is now merged to `master`, and from `master` it is merged into @ahenoch's new branch called `pan-genome-graph-nem-engine`. Can you please in your anvi'o clone git pull and then switch to `pan-genome-graph-nem-engine` to test everything? All we need is a thumbs up that your tests work well also in the `pan-genome-graph-nem-engine` branch :)

If there are things that need to be changed, please send a PR from your fork of the pan-genome-graph-nem-engine` branch so there are no more conflicts.

Thank you for your help!

TianyiChang:
Hi @meren 

I've pulled and switched to the new `pan-genome-graph-nem-engine` branch and tested it on P4 and P2 phages. The node size and line width look good in both cases. However, all the gene clusters seem to be recognized as "backbone" in `GENESxSYNGCs.txt`, including those contributed by a single genome. I've attached the analysis folders so you can reproduce the results. I want to make sure I've updated the codebase correctly; here's the sequence of commands I ran:

```
conda activate anvio-dev
cd ~/github/anvio
pip install -e .

git checkout master
git pull

git checkout pan-genome-graph-nem-engine
```

Now:
Screenshot 2026-07-08 at 14 04 06

Before:
p4_satellite_6_genomes

[p2_phage.tar.gz](https://github.com/user-attachments/files/29787344/p2_phage.tar.gz)
[p4_satellite.tar.gz](https://github.com/user-attachments/files/29787352/p4_satellite.tar.gz)

ahenoch:
@TianyiChang I'm on it and will report back shortly :)

ahenoch:
Screenshot 2026-07-08 113531

Screenshot 2026-07-08 115106

Hey @TianyiChang, the reason for the split into 6 independent components is that the new tool calculates multiple fusion scores for potential SynGCs. One of the fusion scores is especially important when it comes to MAGs, it calculates the potential overlap between contigs and has a minimum gene number threshold. Your genomes are so small, that every single genome simply falls below that threshold. 

To make these genomes work with the new tool there are two simple options:
1. change the --min-line-pair-hits to something less than 50, e.g. in your case 10 or 5 to keep the score form becoming zero
2. change the --ranking-mean setting to arithmetic to have more graceful final score calculation

I used the second option only to generate the graphs. because the --ranking-mean defines the final scoring and if one component score (the one I described above) ends up becoming zero the overall ranking score becomes zero as well when using geometric score calculation.

In the results I attached the GCs look a slightly different and two genomes seem to be missing the last 2 genes, definitely worth checking. I created everything from scratch (PAN DB, GENOMES DB etc.) because there were no CONTIGS DBs in the folders you send and I used the pangenomic standard values for it, not sure if you changed something there. I'm happy to rerun and check everything if you send me the CONTIG DBs as well.

Thank you!

meren:
Thanks for looking into this, @TianyiChang!

Hey @ahenoch,

> To make these genomes work with the new tool there are two simple options:
> 
> 1. change the --min-line-pair-hits to something less than 50, e.g. in your case 10 or 5 to keep the score form becoming zero
> 2. change the --ranking-mean setting to arithmetic to have more graceful final score calculation

I think, in an ideal world, `anvi-pan-genome-graph` should catch it when none of the contigs / genomes pass these arbitrary length thresholds and it should end the analysis _with_ meaningful error messages for people to figure out what was wrong with their default setup, and _without_ producing output files where SynGCs in a single genome are reported as backbone.

TianyiChang:
Hey @ahenoch,

Thanks for your detailed explanation! Now I can reproduce the same result by adding --ranking-mean arithmetic. For anvi-pan-genome, I lowered the minbit to 0.3 to improve gene clustering for these MGEs (they have more divergent genes as compared to bacteria). This resulted in the differences between our results. 

Screenshot 2026-07-09 at 13 38 45

meren:
Very interesting.

PR #2708, opened 2026-06-05
Less redundancy, better messaging

meren:
A bit more anvi'o-way this way. Tell the user what's up, and help them figure out the next steps by themeselves.

### DIFF
--- a/anvio/reactionnetwork.py
+++ b/anvio/reactionnetwork.py
@@ -6069,30 +6069,34 @@ class ModelSEEDDatabase:
                 raise ConfigError(f"There is no such directory, '{modelseed_dir}'.")
         else:
             modelseed_dir = self.default_dir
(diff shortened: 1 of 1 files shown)


ekiefl:
Anvi'o is to _anvian_ as Python is to pythonic? (I'm not stalking I promise, this stuff just shows up on my feed)

meren:
lol :)

Anvian (or anvi'an) is not quite our Pythonic, IMO.. And I'm saying that because the term is already in circulation in our Discord channels where it is used more like Pythonista than Pythonic in your analogy (as in a people rather than a property of the anvi'o enterprise).

Which highlights a vacancy for something that must be filled: anvi'o has no sanctioned adjective for it's coding attitude (where elegance is not present in it but always desired hehe).

If there was a committee on anvian nomenclature (and maybe we should have one), it would have most likely proposed '**anvi'onic**' as a candidate as an analog of Pythonic.

PR #2712, opened 2026-06-07
Ability to download gene FASTA files for SynGCs

meren:
We have `anvi-export-pan-subgraph` to get everything from the pan-graph databases, but while working with the interface sometimes the user (meren) simply wishes to download all the sequences for all SynGCs in a given bin, that are associated with a region, or the user (meren) simply clicked upon.

This PR adds that function for our users (merens) to benefit from such interface conveniences.


ulgens:
A good day for all merens.

meren:
Merens should unite.

PR #2713, opened 2026-06-07
Enable metabolism framework tests

ulgens:


ulgens:
@bcoltman 👋🏻 Is [this](https://github.com/merenlab/anvio/actions/runs/27087535874/job/79944817255) error something you encountered while working with Python 3.13 update?

meren:
Hey @ulgens,

The error I see in that output is due to the fact that anvi'o is unable to find the KEGG data directory on the virtual machine. it is downloaded and setup by another program, `anvi-setup-kegg-data`, which is not included in the test file as the test assumes that the tester has already run it on their comptuer.

A smart way to solve it would have been to include in the [test file](https://github.com/merenlab/anvio/blob/master/anvio/tests/run_component_tests_for_metabolism.sh) something like

```bash
anvi-setup-kegg-data --kegg-data-dir $SOMETMPDIRECTORY

export kegg_data_dir=$SOMETMPDIRECTORY
```

Then the rest of the script will work.

What do others think? @ivagljiva, does what I suggest make sense?

bcoltman:
@ulgens this did pop up in my tests but was failing because locally I didn't have KEGG data installed, as it is in the GitHub runner. I was curious whether the workflow/runner could handle that amount of data so had a look at it in #2716

bcoltman:
@meren exactly! I'm not sure why the full download didn't work within the python script itself, but a curl appeared to do the trick in the fix

ivagljiva:
I think @bcoltman 's solution in #2716 is the right one :)

ulgens:
Thank you  all. This will need a bit more of work to adapt into the new shared base approach. I'll revisit it in the next couple days.

PR #2718, opened 2026-06-08
Support MUSCLE 5 in the MUSCLE driver

bcoltman:
# Summary
This PR updates anvi'o's MUSCLE integration for MUSCLE 5 while keeping the driver compatible with MUSCLE 3.

The main issue is that MUSCLE 5 changed its CLI compared to MUSCLE 3. Previously, anvi'o could call MUSCLE through the older stdin/stdout style and also had a `clustalw_format` / `-clw` path. After the KEGG HMM build change in #2717, that CLUSTAL output path is no longer used, so this PR removes it and updates the driver to use the right command syntax depending on the installed MUSCLE version. It is a little bet messy but means that the users of the anvio-development branch can continue with their current installed MUSCLE version, whilst all new installs can use v5. 

## Changes include:

- Detect MUSCLE 3 vs MUSCLE 5 in the driver.
- Keep MUSCLE 3 support through the existing stdin/stdout behavior.
- Support MUSCLE 5 `-align ... -output ...`.
- Support MUSCLE 5 `-super5 ... -output ...`.
- Remove the now-unused `clustalw_format` / `-clw` path from the MUSCLE driver.
- Update EcoPhylo's MUSCLE alignment rule to use version-appropriate command syntax.
- Make EcoPhylo's default MUSCLE `additional_params` version-neutral - the CLI commands in the workflow are no longer supported by MSUCLE v5
- Update the main conda environment to require `muscle>=5`.
- Update the Dockerfile install to to require `muscle>=5`.
- Update MUSCLE parameter docs for MUSCLE 5.

`Dockerfiles/anvio-structure/Dockerfile` is left pinned to MUSCLE 3.8 as that Dockerfile installs the old released anvi'o 7.1 code, not this branch's updated MUSCLE driver. Updating it here would break that image rather than help it.

## Testing I ran:

- MUSCLE 3 driver test in an `anvio-dev` environment installed as outlined in the docs.
- MUSCLE 5 driver test in a test environment with all dependencies pinned as outlined in install docs, apart from MUSCLE now being at v5. 
- MUSCLE 5 `MUSCLE_PARAMS=-super5` test.
- KEGG `build_HMM_from_seqs()` with MUSCLE 5 produced a valid `HMMER3/` model.
- EcoPhylo MUSCLE command checks for MUSCLE 3, MUSCLE 5 `-align`, and MUSCLE 5 `-super5`.
- `py_compile` for touched Python modules.
- `anvi-self-test --suite pangenomics --no-interactive` completed successfully

So the fix here is to make anvi'o speak both MUSCLE CLIs in the driver, while moving the main anvi'o environment forward to MUSCLE 5.


meren:
Hey @bcoltman. you are doing God's work. Thank you so much for this sorely needed update!

I appreciate that you took the high road and made things backwards compatible. But I think we could also drop the support for MUSCLE 3 completely, and update everything to work with MUSCLE 5.

The reason I am suggesting this is because this kind of backwards compatibility (especially when there is a superior option) inadvertantly creates risks for people to make mistakes without realizing in the future. But if you tell me that you spent enough time on this and would like it to be merged, then it is ready to go in, IMO :) Please let me know what you think!

Thank you again,

bcoltman:
Hi @meren,
sure, I'm happy to drop backwards compatibility. I'll look into it in the coming days! Shall I include the version check, such that an elegant error is spat out rather than a messy error message?

bcoltman:
@meren small delay but it is done.

I updated this PR following the discussion about whether we need to retain MUSCLE 3 compatibility. This branch now treats MUSCLE 5 as the only currently supported version moving forward.

Main changes from the previous version:

- Dropped the MUSCLE 3 command path from the driver.
- MUSCLE now runs through the MUSCLE 5 file-backed CLI:
  - `muscle -align ... -output ...`
  - `muscle -super5 ... -output ...` when `MUSCLE_PARAMS=-super5`
- Kept a clear `ConfigError` when MUSCLE 3 is installed, including an example conda update command:
`conda install -c conda-forge -c bioconda "muscle>=5"`
- Removed the now-misleading `Muscle.run_stdin()` wrapper and updated internal anvi'o call sites to use `run_default()`
directly.
- EcoPhylo still uses a path-based shell command because Snakemake already provides input/output FASTA paths, but it instantiates `Muscle()` first so users get a MUSCLE 5 version check.
- Left the anvi'o-structure Dockerfile unchanged because it installs the old released anvi'o 7.1 code.

Testing after the latest update:

- `py_compile` for the edited Python files.
- MUSCLE 5 driver smoke test with `run_default()`.
- MUSCLE 5 smoke test with `MUSCLE_PARAMS=-super5`.
- MUSCLE 3 environment check confirms the new clean error message.
- `anvi-self-test --suite pangenomics --no-interactive` completed successfully.

meren:
My tests look OK on my end (tested for phylogenomics, pangenomics). The only noticeable difference is the speed. v3 appears to be 2 times faster that v5 in most cases. Especially in pangenomics, where we typically align thousands of sets of less than 100 sequences.

But v5 is doing much better on benchmarks, and it is much more aligned with anvi'o spirit to take that speed loss to gain more accuracy in general for phylogenomics, population genetics, even homogeneity index calculations.

We will have to see how it changes things in EcoPhylo, but in that case we can add `-maxiters 2` to the default params or something.

I'm pinging @FlorianTrigodet and @ivagljiva here so they are aware of this change and its implications.

I suggest we merge this branch to `master`, and I ask for opinions/input.

ivagljiva:
I think you should go for it and merge :) I agree that better accuracy makes the slowdown worth it.

FlorianTrigodet:
Hey everyone. I was working on the refactor of ecophylo (uses muscle) and another project where I would like to use muscle v5 so here I am.

I've made some small changes which are worth documenting here:
- check for muscle version once, not for each alignment, and fail immediately
- in panops, check for muscle early instead of failing later
- muscle 5 now uses all available thread on your machine. I added a default thread number of one, since anvi'o is already dealing with multiprocessing and expect each alignment to use one thread as v3 was doing.
- improved catch of muscle version, only v5 accepted. We don't know if we'll get a v6, but given the pace and the amount of changes between v3 and v5, we should be safe and just say we can handle v5, but not newer versions.
- make -super5 the default aligner for ecophylo. muscle v5 comes with two aligner: default and super5, the latter is meant for large MSA, which is typical of an ecophylo alignment, as opposed to the GC alignment in pangenome. 
- remove tmp dir wether the alignment worked or not (except if you use debug, then you get all the tmp dir, as usual).

The pangenomics self test is working fine.

One think left is to update installation instructions at the same time as we merge this branch (I'm on it).

meren:
Thanks for these, @FlorianTrigodet!! Very important updates. I've been meaning to get to this since all my pangenomes have been (relatively quietly) failing at the alignment stage. I guess it won't be the problem anymore :)

PR #2720, opened 2026-06-08
Create test-base action

ulgens:
Coming from https://github.com/merenlab/anvio/pull/2714#issuecomment-4646729325

Resulting workflow seems to be working fine: https://github.com/merenlab/anvio/actions/runs/27152345582

I'm not sure about the zizmor ignore rule, but considering these changes doesn't introduce any new interface to execute code, I'm not sure if it's actually adding any risk.

### DIFF
--- a/.github/workflows/component-tests-and-migrations.yaml
+++ b/.github/workflows/component-tests-and-migrations.yaml
@@ -9,16 +9,13 @@ permissions:
   contents: read
 
 env:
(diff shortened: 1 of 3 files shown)


meren:
I think zizmor ignore rule is completely fine in this case.

If you think this is a good direction, let's go! :) Thank you very much, @ulgens. It is now merged!

PR #2730, opened 2026-06-19
Adding an 'Export SVG' button to the inspect page

ivagljiva:
This PR adds a button to export the inspect page as an SVG. The button is at the bottom of the inspect page's settings panel:

image

In order to do this, we needed a new function in `charts.js` that would combine all SVG elements of the inspect page (since there are multiple -- the per-sample coverage charts, SNV info, sample labels, gene arrow/context, and any highlights are all separate elements) into one merged SVG before triggering the download. The function explicitly:

- Collects the five SVG layers that make up the inspect page (#chart-container, #highlight-boxes, #SNV-boxes, #sample-titles, #context-container) and merges them into a single root <svg> element, preserving their visual stacking order.
- Uses getBBox() on the chart SVG to measure only the rendered content height, so hidden sample layers do not produce excess whitespace above the gene call panel.
- Places the gene call panel (normally fixed at the bottom of the viewport) immediately below the chart content in the exported layout.
- Adds the split name as a title at the top of the figure, matching the style of the inspect page header (page_header bold + " detailed").
- Injects a minimal inline <style> block with the axis CSS rules from charts.css (fill: none; stroke: #aaa) so axis lines render at the correct weight and colour without a live stylesheet.
- Strips the brush/slider element (g.brush) from the gene call panel, since it is purely interactive and would otherwise render as a solid black box in a static file.
- Downloads the result as a well-formed SVG file (with XML declaration and DOCTYPE) named after the split.

I am not a JavaScript-savvy person, but I asked Claude to match the existing export code for the main interactive page whenever possible, and I did read through the code to ensure it looks reasonable. I've tested the export with and without hidden sample layers, with and without SNVs being shown, and the resulting SVGs look good. One example:

TARA_SAMEA4398168_MAG_00000025_000000000042_split_00001

Still, I'd appreciate if someone else would test this as well before we merge :) Thanks!

### DIFF
--- a/anvio/data/interactive/charts.html
+++ b/anvio/data/interactive/charts.html
@@ -238,9 +238,12 @@
                   <br><br>
         </div>
 
-        <div class="btn-group sidebar-footer pr-3 pl-3 pt-3 pb-1" role="group">
-          <a type="button" class="btn btn-outline-secondary btn-sm" onclick="showLoadStateWindow();" data-help="load-state-button" data-original-title="" title=""><i class="bi bi-upload mr-1"></i>Load State</a>
-          <a type="button" class="btn btn-outline-secondary btn-sm mr-3" onclick="showSaveStateWindow();" data-help="save-state-button" disabled-in-read-only="true" data-original-title="" title=""><i class="bi bi-download mr-1"></i>Save State</a>
+        <div class="sidebar-footer pr-3 pl-3 pt-2 pb-2">
+          <div class="btn-group w-100 mb-1" role="group">
+            <a type="button" class="btn btn-outline-secondary btn-sm" onclick="showLoadStateWindow();" data-help="load-state-button" data-original-title="" title=""><i class="bi bi-upload mr-1"></i>Load State</a>
+            <a type="button" class="btn btn-outline-secondary btn-sm" onclick="showSaveStateWindow();" data-help="save-state-button" disabled-in-read-only="true" data-original-title="" title=""><i class="bi bi-download mr-1"></i>Save State</a>
+          </div>
+          <a type="button" class="btn btn-outline-secondary btn-sm w-100" onclick="exportInspectSvg();"><i class="bi bi-image mr-1"></i>Export SVG</a>
         </div>
       </div>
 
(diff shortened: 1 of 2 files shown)


meren:
Thanks for this, @ivagljiva! I think we all needed this so much at least once -- I'm so happy that someone finally took a stab :)

It works, but here is a problem if you want to update it. When you zoom in to a specific context, the resulting SVG still includes all the other data inside rather than exporting only what is visible:

image

It may be convoluted to export only what is visible (since D3 does funny things to give the impression of zooming to a particular section without actually any real SVG processing in the background), if that is the case, we can ignore it. But otherwise this would be my only 2 cents.

ivagljiva:
Okay, it took some fiddling, but I think I managed it. Actually the hardest part was not the clipping of the non-visible data, but rather making the slider box render nicely. Claude added explicit styling of the slider box and the background rectangle of the context window (which also had to be moved outside of its group so that it would render underneath, instead of obscuring the slider box and axis). Here is the more technical summary of what was changed:
  - Clip path for zoom — a clipPath is applied to individual data path elements (path.chart, path.line, SNV markers) so that when the user has zoomed in with the brush, data outside the visible range doesn't overflow the axis frame. Applied to elements only (not whole groups) so the y-axes — which sit outside the data column — remain unclipped.
  - Brush rect fixup — after all layers are appended, the D3 brush rects are patched directly: rect.background gets a solid rgb fill (matching the live-page context colour), its visibility: hidden inline style is stripped, and it is moved to paint before the axis so tick marks show on top. rect.extent (the zoomed-region indicator) also gets an explicit solid fill and a stroke border. All colours are plain rgb() —
  no rgba() or fill-opacity, which avoids compositing issues in SVG 1.1 viewers.

Here is my example (note -- no SNVs). First, no-zoom:
image


Now zoomed into one gene (you can see the slider box now):
image


Finally, zoomed in all the way so you can see the sequence:
image


It's not perfect, some elements are dangling off the page in the closest zoom. But it works.

ivagljiva:
Proof that it works when we also visualize SNVs (now you can see the axis on the right):
image

meren:
Beautiful!! As someone who ALWAYS struggled with this feature missing, I hereby thank you very much for doing this! :)

Issue #2606, opened 2026-05-18
[BUG] Number format in external gene call table (int / numpy int)


bastien-camillo:
## Short description of the problem

We realised that anvio takes only int integers into account in the external gene call table and not numpy integers

## anvi'o version
anvio-8

```
anvi-self-test --version

Anvi'o .......................................: marie (v8)
Python .......................................: 3.10.13

Profile database .............................: 38
Contigs database .............................: 21
Pan database .................................: 16
Genome data storage ..........................: 7
Auxiliary data storage .......................: 2
Structure database ...........................: 2
Metabolic modules database ...................: 4
tRNA-seq database ............................: 2

```

Kekananen:
Here is a branch on a forked anvi'o: https://github.com/Kekananen/anvio/tree/not_accepting_numpy

If you would be able to test to see if this fixed the problem you encountered and let me know that would be fantastic! Thanks!

Kekananen:
From Discord conversation from the hackathon 2026 from back in issue in 2024:

This is the error that was thrown when trying to add in a gene call table:
```
File/Path Error: Values in the call_type column must be integers :/ Please see
                 http://anvio.org/help/main/artifacts/external-gene-calls/
```
This is a table that works:
```
gene_callers_id     int64
gene_id            object
mRNA_id            object
contig             object
start               int64
stop                int64
partial             int64
call_type           int64
source             object
version            object
aa_sequence        object
direction          object
dtype: object
```
After numpy is used, it is no longer working with the table:
```
gene_callers_id     int64
gene_id            object
mRNA_id            object
contig             object
start               int64
stop                int64
partial             int64
call_type            int8
source             object
version            object
aa_sequence        object
direction          object
dtype: object
```

FlorianTrigodet:
I know @Kekananen worked on a fix for that in the snakemake refactor (#2667) and it was removed from that PR to go to a separate PR dedicated to that issue. I don't know if you did it @Kekananen, and if you wanted, I can do it this week :)

Kekananen:
Ah sorry! Yes, let me see about that. I have just submitted a PR (https://github.com/merenlab/anvio/pull/2731) for the fixes that were moved. The unit testing changes will likely need to be removed or formatted to the preferred style for anvi'o.

Kekananen:
I believe this issue is now closed. I'll leave it to @FlorianTrigodet or @bastien-camillo to make the final verdict!

PR #2731, opened 2026-06-23
Accept numpy integers in external gene calls

Kekananen:
This was made as a fix for the issue #2606.


Kekananen:
@FlorianTrigodet

FlorianTrigodet:
Hello there.
I removed one instance of `normalize_gene_calls_dict_integer_types()` because it was redundant the code already ensures the right format (see `column_mapping` in gene calls from file path) and I removed the unit test to respect the anvi'o codebase guide.

PR #2741, opened 2026-07-06
Better long reads integration

FlorianTrigodet:
In the previous episode of the BIG SNAKEMAKE REFACTOR (#2667), @Kekananen @btrouche @katysloz @l-gallucci also introduced new and better support for long-reads in the metagenomic workflow, which included long-read QC with longqc and fitlong. I removed these new features in the previous PR to keep it REFACTOR focus. Now is it time to bring back all these good ideas to the code base through this PR.

Along with these added new rules to the workflow is a key decision to add a new column to the samples-txt to specify the LR technology as some steps (longqc, flye, minimap2) require a technology preset, which depends on the nature of the user's data. While this new lr_thechnology column was first mandatory, I made it optional for the following reasons:
- not very useful when the user has homogenous LR data type. The presets sets in the config.json would be used in this case.
- most importantly, maintaining a set of technology>preset (currently in a newly made yaml file) is going to be difficult and if a power user wants to opt-out of our set of presets, they should be able to do so.

Other little things introduced here:
- no more default presets for minimap2 and flye. User HAS to provide a preset otherwise the workflow does not run. Important because minimap2 would still run, using its default preset.
- config file migration: add the new QC commands.
- the lr_thechnology column is "all or nothing": if set for one LR row > has to be set for ALL rows with LR (rows with only SR are excluded of course).

I ran some basic tests, and I am planning on running larger real-life test with the zymo mock community dataset in the coming days.


meren:
🎉 🎉 🎉

FlorianTrigodet:
To the original authors (@Kekananen, @katysloz, @btrouche, @l-gallucci): what was the original motivation for using LongQC over NanoPlot/NanoStat? I'm asking because the installation of LongQC is a massive pain (the author prefer running a docker container, which we don't support yet, and the manual installation instruction are not complete and not fun to figure out. Plus issues with the python script not in the path, not excecutable. Their conda recipe is not working either).

Kekananen:
LongQC was chosen over NanoPlot mostly because I have used it before, but also because it gives a deeper pool of information for seeing if something is wrong with the run. A few that it gives that (last I checked which was a little while ago) NanoPlot doesn't is a read coverage report and reports reads with no internal support. LongQC can flag things like suspicious coverage dispersion and short-fragment contamination. Basically LongQC is the heavy hitting version of NanoPlot; however all this said, Nanoplot (I guess Cramino is a rust replacement that has superseded NanoStat that the NanoPack author recommends now?) is a lighter version that may fill a basic need for filtering QC for most people.

As for the install being a pain, it might be better to then use Cramino then and give up the non-sense read diagnostic abilities then. Otherwise, maybe adding in a `longqc.post-deploy.sh` would work to make install automatic using snakemake's built in conda system. This solution doesn't really lend well to the system set up by anvi'o though and would be an oddball case for a single tool which isn't a great solution for maintainability. In a pinch, I've used this method to install some tricky tools before automatically and it does work decently. 

Maybe a good middle ground would be to include [LongReadSum](https://github.com/WGLab/LongReadSum) and NanoPlot/Cramino.

FlorianTrigodet:
After running a test case with two long read metagenomes, I realized that LongQC is not designed for metagenomes and the coverage step failed because it assumes a uniform coverage from a single genome. Based on that, I would suggest moving to a (more simple) QC tool like NanoPlot or LongReadSum: a tool that can work with any type of reads.

I've explored and compared LongReadSum and Nanoplot, and when using fastq as the input file, the two software output the same info (LongReadSum can use many type of input and I think that's where it is more novel and interesting than Nanoplot). I picked NanoPlot as it was the fastest (12min vs 19min for the same metagenome). And I do like the kde plots I have to admit.

Let me know what you think @Kekananen :D

Kekananen:
I'm glad you tested these tools on metagenomes and caught that LongQC doesn't work as well. I would rather a quick QC for a basic look then a longer one in this case. Maybe in the future we can look at fully input capability with LongReadSum, but after all your investigative work, Nanoplot seems to be the way to go!

FlorianTrigodet:
The current stage of the PR:
- samples-txt: new optional columns for long read technology and an associated yaml file in the workflow directory to map corresponding technologies to flags for minimap2 and flye.
- new read stat programs: FastQC for short reads, NanoPlot for long reads, both summarised by MultiQC (creates a single final output with stats from all samples, all technologies)
- new long read filtering program with filtlong

That's the core of the features in this PR, but I also added more conda options for third party software that SHALL NOT be installed in the anvi'o conda env. I understand that is technically out of scope of this PR, but I couldn't run proper test without it (and I failed to make a separate PR for it - I swear I'm slowly learning how to be better at that). In short, for most third party sofware in the workflow (mappers, assembler, QC), there are three options in the config file for how to run the rule/command with conda (or not):

```json
    "rule": {
        "conda_yaml": "",
        "conda_env": "",
        "use_anvio_conda_yaml": false
```

From these three options, two were already implemented a while back: `conda_yaml` and `conda_env`. The first one takes a path to a yaml with the conda recipe for that program, snakemake will make it and run the rule within that env. The second option takes the name of a conda env already existing on your machine and simply appends `conda run -n {name} [...]` to the rule. The new option, `use_conda_anvio_yaml` is a true/false and when true, snakemake will use a yaml recipe that now ships with anvi'o in `workflows/conda_envs`. Simple recipe that should work on any machine (🤞), and most importantly, that are used in the component test for workflows! Finally, if the user does not setup one of these three option (mutually exclusive, through a sanity check), anvi'o will expect (and check) that the program is rightfully in the $PATH.

Everything mentioned here, and much more, was also added to the documentation (including a new section about conda and third party software in the doc for the metagenomic workflow).

I ran a real world test with a mix of ONT and HIFI data and it worked fine as far as I can see. I also updated the test to include ONT mock metagenomes. So far so good. I think I reach the extent of my review capabilities and catching all obvious bugs. The next steps/fixes will come as we use that workflow. I can already foresee that the (few) LR technology presets are not enough, and fortunately that's now easy to expand.

FlorianTrigodet:
Fun note when reviewing code with Claude Code: you know Claude is trying its best to complain about something when the only issue it finds is 'if a fastq file has a mid-file blank line, then XXX will fail'. Yes Claude, a lot of things will fail when fastq files have randomly distributed empty lines. Good if that's one of the only and last issue you are able to detect in the current PR.

PR #2742, opened 2026-07-06
A crisper display for anvi-display-pan-graph

meren:
`anvi-display-pan-graph` has a problem. When it is initialized, it draws the display as one large SVG within a viewBox that is tens of thousands of units wide that is squeezed into the browser viewport.

When we look at the entirety of a graph, the lines (or stroke widths in SVG lingo) are collapsed to sub-pixel widths (since you can squeeze things only so much) and as a result the lines are typically rendered as faint, shimmery antialiased garbage (even though when you download the SVG and open it in Inkscape, or zoom in all the way, you see perfectly fine lines). Here is an example of what user sees and that annoys the crap out of the user because the user wants to be able to just take screenshots and send it to their colleagues but they can't do it because this is what things look like which is why the user is so annoyed:

image

When it should look like THIS, which is what this PR enables:

image

This is happening how we resolved some performance issues we experienced before. We used non-scaling-stroke (constant screen-pixel width, so no computation when zoomed in and out or panned the display around), which made lines obnoxiously thick at 100% view while looking fine zoomed in. That obviously was not the solution. What this PR does it that it keeps strokes scaling naturally with the graph but clamps each to a minimum on-screen width applied in the zoom handler. So lines stay proportional and nice at every level of 'zoom', but never never drop below the default min line width (which is now by default 0.5px) and shrivel into some antialiased garbage like they do in the first image.

This PR also introduces two controls: min line width (px) input so the user can change the default if they wish, a checkbox for "high-performance mode" so they user can toggle that to skip the per-zoom pass on very large graphs (plus a regime cache so zoomed-in interaction does zero per-element work):

```js
            // High-performance mode: skip the floor entirely (cheap pan/zoom;
            // lines may go sub-pixel when fully zoomed out).
            if ($('#flexhighperf').prop('checked')) { to_base(); return; }
```

So we can have our 🍰 and eat it too.

It would have been great if @TianyiChang could do a git pull and see if this solution works for MGEs or smaller pangraphs (re: #2695)

### DIFF
--- a/anvio/data/interactive/js/pangenome-graph.js
+++ b/anvio/data/interactive/js/pangenome-graph.js
@@ -2158,6 +2158,56 @@ class PangenomeGraphUserInterface {
 
         $('#svgbox').empty().html(svg_core[0].outerHTML);
 
+        // Minimum on-screen line width. The graph lives in a huge viewBox that
+        // gets squeezed into the viewport, so at fit-view stroke widths (set in
+        // user units) collapse to sub-pixel widths and render as faint, shimmery
+        // "antialiased garbage". We keep strokes scaling naturally with the graph
+        // (so they stay proportional and look right at every zoom) but never let
+        // a rendered stroke drop below MIN_PX device pixels:
+        //   rendered_px = baseW * realZoom  ->  clamp so it's >= MIN_PX
+        //   => stroke-width attr = max(baseW, MIN_PX / realZoom)
+        // Zoomed in, MIN_PX/realZoom is tiny so widths are natural; zoomed out,
+        // the floor keeps thin lines crisp instead of collapsing.
+        //
+        // Snapshot every strokable element's natural (base) width once per redraw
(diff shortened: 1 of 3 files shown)


meren:
I just wanted to make sure it was backwards compatible. Here, I downloaded the publicly available Undatipelagibacter pangenome graph:

```bash
# go to your home directory
cd

# create a new directory and go into it
mkdir UNDATIPELAGIBACTER-PAN-GRAPH && cd UNDATIPELAGIBACTER-PAN-GRAPH

# download all necessary files
curl -L https://cloud.uol.de/public.php/dav/files/TN2bxBCbAS5DRDJ -o UNDATIPELAGIBACTER-GENOMES.db
curl -L https://cloud.uol.de/public.php/dav/files/8eZZYqNrAdXF4TA -o UNDATIPELAGIBACTER-PAN-GRAPH.db

# initiate the interactive interface
anvi-display-pan-graph
```

This is the old way:

image

This is the new way:

image

PR #2743, opened 2026-07-06
Pan genome graph nem engine

ahenoch:
Not sure where to start here, this branch introduces multiple new functions. 

The top three most important things:

- nem-engine: an entirely new engine using a different approach to producing pangenome-graphs relying on bottom-up instead of top-down, enabling the use of MAGS, also, reversed edges are not a thing anymore
- components: datasets with multiple chromosomes, plasmids now show a drop down menu where you can switch between these components ("sub-populations" inside your dataset will also be shown as two entirely different components)
- all search functions are now useable

IMPORTANT INFOS:
- major changes to the PAN-GRAPH.db require a rerun of the tool.
- anvi-reorient-genomes is obsolete, instead in case of complete genomes a tool like dnaapler is advised but not required

A few figures to highlight the new functions:

image

image

image

PS: I changed the color scheme because red-silver rock, but we can go back to yellow blue as well :D


anvio/version.py
```diff
@@ -21,7 +21,7 @@
 profile_db_version = "42"
 genes_db_version = "6"
 pan_db_version = "21"
-pangraph_db_version = "5"
+pangraph_db_version = "7"
```
meren, line 24: Why from 5 to 7 and not 5 to 6? o_O

Especially when the branch only has `anvio/migrations/pan-graph/v6_to_v7.py` :)

anvio/migrations/pan-graph/v6_to_v7.py
```diff
@@ -0,0 +1,163 @@
...
+        # split (component_id, region_id) pair before reordering.
+        carryover = {c: v for c, v in zip(
+            [c for c in old_columns if c != 'region_id'], rest)}
+        carryover['component_id'] = cid
+        carryover['region_id'] = rid
+        new_rows.append([carryover.get(col) for col in t.pan_graph_regions_table_structure])
+
+    pan_graph_db._exec("DROP TABLE pan_graph_regions")
+    pan_graph_db.create_table(
+        t.pan_graph_regions_table_name,
```
meren: This migration script will break in a few months to years. 

You should never import `anvio.tables`, since the entire purpose of the migration script is to be able to track changes independent of the current state of the library.

If `t.pan_graph_regions_table_structure` changes at some point, and if you are updating a database from 5 years ago, `t.pan_graph_regions_table_structure` in the modern code will differ from the ancient expectations.

Please look at the migration scripts under `anvio/migrations/pan` to get inspiration (or ask claude to do so). I now updated the `ARCHITECTURE.md` in #2744 to make sure this is always a consideration for LLMs.

ahenoch: I'm not sure if migration is possible at all. The old pangenome graph included "reverse" edges to solve higher complexity. In the new version this is impossible by design. Therefore there is no reason to keep the edge direction in the table anymore. 

This feels a little similar to the potential update of the genomes-storage.db we talked about during the hackathon that would break pangenomes from before the change. But keeping the edge direction, knowing it will never be used again feels wrong, especially since the pangenome graph was just released and not a lot of people use the tool yet.

Future features will also rely on the fact that reversed edges are not possible, so having a table with direction populated by old pangenome graphs only feels even more counter intuitive, especially knowing that they won't be useable with these features.

What do you think?

ahenoch: The new migration script tells the user, that the "reversed" edges were removed during migration, which in most cases will fracture the graph a bit (e.g. G1 becomes 4 unconnected lines instead of a single long line). In our Undatipelagibacter dataset this will happen once, so one genomes genome track will be split in two parts, and the single reversed edge will be gone

anvio/version.py
ahenoch: I updated the script and corrected the version to 6 🙈

anvio/migrations/pan-graph/v6_to_v7.py
meren: I was not saying the table should not be changed or data should not be dropped from it. I was saying that the migration scripts should maintain their own table knowledge rather than relying on what `anvio/tables.py` says, which appears to be fixed in 4de236104a110d856d29ef3af2d2d2207d2df03e.

PR #2747, opened 2026-07-12
Program name fixies

meren:
This PR renames to programs: `anvi-script-gen-function-matrix-across-genomes` is now `anvi-gen-function-matrix`, and `anvi-script-gen-hmm-hits-matrix-across-genomes` is now `anvi-gen-hmm-hits-matrix`.

Following #2746, I realized that `anvi-script-gen-function-matrix-across-genomes` is no longer specific to 'genomes'. Which has never been the case actually, and I am not sure why the original script name had that level of specificity, but now it certainly not specific since PPCN normalization is only relevant for metagenomic assemblies.

While at it, I also fixed `anvi-script-gen-hmm-hits-matrix-across-genomes` since the same applies.

I think we should add the same flag to `anvi-gen-hmm-hits-matrix` (which will require PPCN reporting to be its own class elsewhere), but it can wait.


srisvs33:
HI @meren, I was trying to use this Program in my anvio-dev version, which is actively updated to date. However, it says this program is not found. Was it available in Dev branch? Many thanks.
anvio

meren:
Hey @srisvs33, can you go to the directory where the anvi'o codebase is (i.e., `~/github/anvio/`) and run `pip install -e .`?

meren:
Please let me know if it doesn't work.

If it has been a relatively long time since you have installed anvi'o dev environment, you may need to run this instead:

```
pip install -e . --force-reinstall --upgrade
```

If neither of these solves your problem, then I'd suggest you to reinstall it anvio-dev from scratch just to be on the safe side :)

srisvs33:
Thank you, @meren. It worked!

PR #2750, opened 2026-07-15
replace pandas .append with pd.concat

FlorianTrigodet:
not strictly necessary as we pin Pandas version to 1.4.4 but it is depreciated already and was removed from pandas 2.x. 
WE ARE GETTING READY FOR THE FUTURE.


meren:
lol scary stuff but we shall prevail.

PR #2763, opened 2026-07-23
Update structure db colabfold

FlorianTrigodet:
Now the 4th PR after #2755 (colabfold driver + minimum integration), #2759 (update display), #2760 (add --only-msa/--only-predict checkpoints) regarding the integration of colabfold in anvi'o.
This time, it is about adding colabfold into the program `anvi-update-structure-database`.

What it does: mirror the way MODELLER update already work. The parameters (num models, num recycle, ambler, colabfold msa source, colabfold additional parameters) are read from the structure database itself and reused, so an update stays consistent with the original run. Only the 'machine' specific parameters needs to be resupply like colabfold conda env, colabfold database and the optional checkpoints.

Right now it refuses to change the MSA source. If the user used a local colabfold database, they cannot update the structure db by asking to use the public MSA server. I understand that this could be a debatable choice. I went conservative here, although i cannot check if a local colabfold database it the same as in the original run.

Updated the test as well, which now takes a really long time to run (>30 min on the HPC, with 100 CPU to test the checkpoints, and with GPU).

### DIFF
--- a/anvio/cli/update_structure_database.py
+++ b/anvio/cli/update_structure_database.py
@@ -46,6 +46,12 @@ def get_args():
                                                 'Otherwise, an error will be raised.')
     groupO = parser.add_argument_group('OUTPUT', 'Output file and output style.')
     groupM = parser.add_argument_group('MODELLER PARAMS', 'Parameters for MODELLER\'s homology modeling.')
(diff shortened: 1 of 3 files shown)


FlorianTrigodet:
All test passed!

Issue #2754, opened 2026-07-19
[FEATURE REQUEST] anvi-gen-structure-database with Colabfold and new input format


FlorianTrigodet:
## The need

Using Colabfold as an alternative to MODELLER in `anvi-gen-structure-database`.
I would also like to have more input options, like fasta file or pangenome to have structure db or representative.

For the inclusion of colabfold, we will need a set of checkpoint flags `--only-msa` / `--only-predict` to split the CPU/GPU heavy part of colabfold and better integration into snakemake worfklows so that the jobs can run on appropriate CPU/GPU nodes on an HPC.

## The solution

Here is how I will split the work (individual PRs):
- [x] DONE: #2755. Colabfold backend, end-to-end (nop checkpoint for `--only-msa` / `--only-predict` yet). New flag `--engine` for `anvi-gen-structure-database` to select between MODELLER and Colabfold, plus all the new colabfold parameters. 
- [x] DONE: #2759. Compatibility with anvi-display-structure.
- [x] DONE: #2760. Add checkpoint for colabfold `--only-msa` / `--only-predict`. This could be a bit tricky, we need to generate a manifest or something to make sure anvi'o resumes `--only-predict` with the appropriate output of `--only-msa`. So this will be its own PR for clarity and review.
- [x] DONE: #2763. Add compatibility with anvi-update-structure-database
- [x] DONE: #2764. Make sure the protein key/id is compatible with multiple type of intputs. Right now it is a gene caller id associated with a contigs.db, we need to make it ready to work with pangenome, or any other input type we may require in the future.
- [x] New input types (as many PRs as inputs):
  - [x] DONE: #2766, #2767. Pangenome
  - [x] DONE #2768. Pangenome + external structure
  - NOT NOW. Fasta file: I'm not sure it is the best, or rather I don't see the application yet. Anyway the code is ready to include any new input format, so I may just postprone this one until we have an actual need/reason for it.
- [ ] Workflow integration

PR #2764, opened 2026-07-24
Generalize structure protein key

FlorianTrigodet:
Part of the Colabfold integration (#2754), step 5 following #2755 #2759 #2760 #2763.
Here is a very technical PR. All about plumbing. The idea was to change how the protein id were defined and used to allow for new input type (like a pangenome, maybe just a fasta, etc).

Before this PR, the codes was referring to all protein with `corresponding_gene_caller_id` which was the gene caller id hard-linked to a single contigs database.

What is in this PR:
- a new protein id replacing the gene caller id. + a new proteins table which records input type (contigs.db for now), genome name, gene caller id, gene cluster id, source key, has structure. And self gained an 'input type' entry. Kept 'contigs_db_hash' for now.
- New class StructureInputSource so that StructureSuperclass delegate input validation and sequence fetching to a source object. ContigsDBSource is the only implementation for now, but there is a space for PangenomeSource.
- A migration from v3 to v4 for the structure db.

Important note:
For a contigs db source, the `protein_id` == `gene_caller_id`. All downstream code like variability overlay is identical and unchanged.

Tests are on going, and I'll report quickly.


FlorianTrigodet:
Tested and working fine!

PR #2765, opened 2026-07-24
new utils: picking GC representatives + flag to export them

FlorianTrigodet:
A little utils function to get representative sequences from gene clusters. This was originally implemented in another branch #2364, but should be its own PR and merged to master instead.

Includes a new flag for `anvi-get-sequences-for-gene-clusters`: `--representative-sequences`.

Anvi'o picks the represetative using a medoid-based strategy. It discards length outliers and prefers complete (non-partial) gene calls, and select the sequence that is the most similar to the others in the cluster.


FlorianTrigodet:
All tests pass, we merge!

PR #2766, opened 2026-07-24
Pangenome input structure database

FlorianTrigodet:
We're getting there. 6th PR related to the integration of ColabFold (#2754).
This time we introduce a big new change: pangenome as a new input type!
Why? To later compute structure-based pangenome, and to visualize prot structure in the pangenome interactive interface.

Here is how it looks like:
```bash
# every gene of the selected clusters
anvi-gen-structure-database --pan-db PAN.db -g GENOMES.db \
    --gene-cluster-ids GC_00000068,GC_00000088 -o STRUCTURE.db

# a single representative gene per cluster
anvi-gen-structure-database --pan-db PAN.db -g GENOMES.db \
    --gene-cluster-ids GC_00000068 --select-representative -o STRUCTURE.db
```

Here we have a new PangenomeSource in structureops.py. With contigs.db, the protein id == gene caller id; here the protein id is a integer and the protein table links it to a genome name, gene caller id and gene cluster id.

AA sequence come from the genomes-storage db (therefore required input).

Two selection modes: all proteins of selected gene clusters, or representative sequence (#2765, a protein pick not a consensus prot).

Works with `anvi-gen-structure-database` and `anvi-update-structure-database`.

My only issue is that the flag `-p` was already used by MODELLER's parameter `--percent-cutoff`. So right now you need to use `--pan-db` full flag to pass a pan db. A bit annoying. I haven't decided if I can/should retire `-p` from `--percent-cutoff` into the short flag for pangenome. I cannot find any reproducible workflow or documentation that uses the original `-p`, so I think I could be safe replacing it, but it feels wrong too. I'll ask the wiser people when I get the chance.

We also have a new test! Mirroring the unusual test `run_component_tests_for_SCVs_SAAVs_structure.sh`, meaning you run it directly in the `tests` dir, not as part of the self tests. Debatable I agree and I'm ready to hear opinions.

Also, right now you cannot bring external structures when in pangenome mode. I think we should be able. But for now, we will be happy with what we have. The tests are running but I want to run it on a real pangenome first. Then we'll see if we allow external structures.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -1921,6 +1921,32 @@ D = {
             {'metavar': 'FILE_PATH',
              'help': "Text file for gene clusters (each line should contain be a unique gene cluster id)."}
                 ),
+    'gene-cluster-ids': (
+            ['--gene-cluster-ids'],
+            {'metavar': 'GENE_CLUSTER_IDS',
+             'type': str,
+             'help': "Gene cluster ids to focus on. Multiple of them can be declared, separated by a comma "
+                     "(e.g., --gene-cluster-ids GC_00000001,GC_00000002). This is the gene-cluster analogue of "
+                     "--gene-caller-ids. If you declare nothing, anvi'o may assume all of them, depending on the "
+                     "program."}
+                ),
+    'gene-clusters-of-interest': (
+            ['--gene-clusters-of-interest'],
+            {'metavar': 'FILE',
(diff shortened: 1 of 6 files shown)


FlorianTrigodet:
Working great, with a real world SAR11 pangenome, selected the representative sequence of the first gene cluster and got my structure.db 🎉 

The next step here is to allow the user to come with their own external structure if they don't want to use our two options (MODELLER and ColabFold) or if they already have structures.

PR #2778, opened 2026-08-11
Codon stats fixes

semiller10:
This PR contains key fixes to `anvi-get-codon-frequencies` and `anvi-get-codon-usage-bias`.

The flag `--ignore-start-codons` improves the accuracy of synonymous codon frequency calculations for gene sequences containing initiator GTG or TTG, which are otherwise wrongly treated like elongation codons for Val and Leu, respectively.

Genes can now be selected by higher BRITE category annotations, critical for the selection of ribosomal proteins as a reference in the calculation of certain codon usage bias metrics.

### DIFF
--- a/anvio/__init__.py
+++ b/anvio/__init__.py
@@ -4222,6 +4222,21 @@ D = {
                      "rather than genes, this filter is applied to genes before grouping them as "
                      "functions."}
         ),
+    'ignore-start-codons': (
+            ['--ignore-start-codons'],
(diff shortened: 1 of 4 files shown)


meren:
Ah, this is great! Thank you, Sam.

Dumb question/confirmation: as far as I can see this shouldn't affect anything in `anvi-display-codon-frequencies`. Please ignore unless something pops up in your head upon hearing it.

PR #2782, opened 2026-08-12
Sra reads in metagenomics workflow

FlorianTrigodet:
Sometimes the metagenomes you want are not on your computer and you have no wish to keep them there. You want to know whether some organisms show up across a thousand public metagenomes, your server has a few hundred GB free, and the thousand metagenomes will not fit no matter how you arrange them.

Today that means running the `sra_download` workflow first, which downloads everything and keeps it, and then running the metagenomics workflow on the result. This makes the metagenomics workflow able to do it itself, one bite at a time.

### How it looks

Your samples-txt can have accessions instead of file paths:

sample        sra_accession
S01   ERR6450080
S02   ERR6450081

and your config has how much disk space anvi'o may use for reads:

```json
"download_reads": {
    "max_disk_gb": 500,
    "keep_reads": "none"
}
```

That's it. Everything else, like QC, assembly, mapping, profiling, behaves exactly as it does for reads that were already on disk. Nothing new on the command line.

### What max_disk_gb promises

It is a promise about a moment, not a total: never more than that much downloaded reads on disk at once. All thousand metagenomes still get processed; there are just never more than 500 GB of them lying around while it happens.

Keeping that promise turned out to be the interesting part. Snakemake deletes a file as soon as nothing needs it, which handles half of it. The other half is that snakemake schedules whatever it can, so left alone it downloads all thousand runs long before the first mapping job finishes. Limiting how many downloads run at once does not fix this, a download slot is released when the download ends, not when the reads are finally used. We tried with Claude: with eight samples and two slots, all eight were on disk simultaneously.

What does work is making a later sample's download wait for an earlier sample's reads to have been used up: a sliding window built into the workflow graph itself. Anvi'o sizes that window from what NCBI reports about each run.

### Things it works out for you

You do not have to say whether an accession is paired-end, or long-read, or which nanopore kit it was. Anvi'o asks NCBI once, writes the answers into SRA-METADATA.txt, and reads that file from then on. That file is meant to be read and corrected, NCBI's description of a run is written by whoever submitted it and is sometimes wrong, and anvi'o never overwrites a row you have touched.

A sample can name several accessions, which is how you describe one that was sequenced over several runs. Since anvi'o knows what each accession holds, a sample can even name a short-read run and a long-read run and come out as a proper hybrid sample.

### Things it refuses, before downloading anything

- Single-end short reads, which this workflow still cannot process. Better to hear that now than 40 hours in.
- A budget too small to hold even one sample, or one co-assembly group.
- Reference-based read removal with QC off, where the removal step consumes the downloads themselves and leaves nothing for later steps.

### keep_reads

none (default), raw, qc, or both. Anything you keep is by definition not deleted, so anvi'o warns that the budget then only covers the transient files.

### Testing

anvi-self-test --suite workflow-metagenomics-sra.

Fifteen scenarios run offline against a committed metadata file and finish in seconds: the workflow graph builds from accessions alone, the gate appears only when the budget is tight, nanopore technology is inferred and PacBio ambiguity is merely warned about, hybrid samples split correctly, co-assembled samples share one release unit, and each of the refusals above actually refuses.

One scenario downloads two small metagenomes for real. No output file can show after the fact how much was on disk during a run, so the test watches while it runs and fails if more than one sample's reads were ever present at once. It records a maximum of 1.

Also verified: all seven existing metagenomics configs still build, the sra_download workflow still builds.


FlorianTrigodet:
Tested using 9 real metagenomes (from BATS) and a read recruitment to a Prochlorococcus genome. Worked like a charm, the disk 'quota' was respected and new downloads only started after older metagenomes were removed from the disk. At the end, we have the anvi'o databases, but no reads at all.

Triple reviewed by claude as well. Ready to merge.

PR #2784, opened 2026-08-18
Gather individual `anvi-draw-kegg-pathways` map files into a directory per map

semiller10:
## Summary

`--draw-individual-files` writes one subdirectory per data source (sample, group, contigs database, or pan genome), each holding that source's whole set of maps. That is the right arrangement for reading everything about one sample, and the wrong one for comparing a single map across samples, as it requires opening one file in each of those subdirectories, one at a time.

This PR adds `--collate-files-by-map`, which arranges the same files the other way round: a subdirectory per map, holding one file per source, named after the source. A file browser's preview can step through the colors of one map changing sample to sample, like an animation.

## Output

```
<out>/by_map/kos_00010/A.pdf
<out>/by_map/kos_00010/B.pdf
<out>/by_map/kos_00020/A.pdf
...
```

`--name-files` and `--categorize-files` apply as they do everywhere else in the output, so with both of these flags:

```
<out>/by_map/Metabolism/Carbohydrate_metabolism/kos_00010_Glycolysis_Gluconeogenesis/SAMPLE_1.pdf
```

This is a second arrangement rather than a replacement — `individual/` stays where it is, and the gathered files are hard links to the maps already drawn there, so they cost no disk space. A hard link is indistinguishable from a real file to a file browser and survives output directories being moved on the same drive. Where a filesystem refuses a hard link, a relative symlink stands in.

`by_map` sits beside `unified`/`individual`/`grid` at the top of the output directory rather than inside `individual/`, preserving the structure that every name directly in the output directory is automatically produced by anvi'o while user-supplied names are limited to directory names in `individual/`.

### DIFF
--- a/anvio/tests/run_component_tests_for_kegg_mapping.sh
+++ b/anvio/tests/run_component_tests_for_kegg_mapping.sh
@@ -1114,8 +1114,109 @@ args+=( "--draw-individual-files" "SAMPLE_1" )
 args+=( "--no-progress" )
 anvi-draw-kegg-pathways "${args[@]}"
 
+INFO "Testing gathering the maps of individual samples into a directory per map, so that one map \
+can be stepped through sample by sample"
+args=()
+args+=( "--reaction-txt" "draw_kos_samples.reaction.txt" )
+args+=( "--output-dir" ${output_dir}/draw_txt_collate_by_map )
+args+=( "--pathway-numbers" "${pathway_numbers[@]}" )
+args+=( "--draw-individual-files" )
+args+=( "--collate-files-by-map" )
+args+=( "--no-progress" )
+anvi-draw-kegg-pathways "${args[@]}"
+
+for sample in SAMPLE_1 SAMPLE_2 SAMPLE_3
+do
+    if [ ! -s ${output_dir}/draw_txt_collate_by_map/by_map/kos_00010/${sample}.pdf ]
+    then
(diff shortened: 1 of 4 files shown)


anvio/tests/run_component_tests_for_kegg_mapping.sh
```diff
@@ -1114,9 +1114,110 @@ args+=( "--draw-individual-files" "SAMPLE_1" )
 args+=( "--no-progress" )
 anvi-draw-kegg-pathways "${args[@]}"
 
+INFO "Testing gathering the maps of individual samples into a directory per map, so that one map \
```
meren, line 1117: I wish you didn't reinvent how these component tests look like, @semiller10 :( Because it will make it so much more difficult to implement changes that will work for all tests the same way in the future.

If you believe component tests should be done differently (i.e., args handling, error handling, the overall aesthetics, etc), then a better way to approach this is to discuss it first, and then implement it for _all_ component tests. Doing it the way that feels like this is a better way to go and breaking compatibility is a significant problem in codebases touched by many people (or LLMs).

My 2 cents.

semiller10: I'm restructuring the component tests to leave a small, straightforward file registered as a known test by `anvi-self-test`, and create an unregistered sibling script, `run_component_tests_for_kegg_mapping_exhaustive.sh`, that can be run by itself in hardcore development. This will land in another PR.

PR #2797, opened 2026-08-25
anvi-script-gen-pan-graph-video

meren:
This PR introduces a new script that allows us to generate progression videos (and details of subgraphs with insets when asked).

Things to do:

- [ ] Write a better PR
- [ ] Write program documentation
- [ ] Implement `--region-id` for insets in addition to 'tell me about the flanking SynGCs' parameter


meren:
I merged a version into `master`, but I will keep it open as I want to continue working on it (but also start testing broadly).

PR #2810, opened 2026-09-14
safer pip reinstall suggestion

FlorianTrigodet:
When `AnvioPrograms.sanity_check()` finds a mismatch between the `anvi-*` entry points declared in `pyproject.toml` and the `anvi-*` programs on `$PATH`, it prints the "ANVIO ENVIRONMENT IS CONFUSE" warning and recommends:
```
pip install -e . --force-reinstall --upgrade
```

This PR changes that recommendation to plain `pip install -e .`, with `pip install -e . --force-reinstall --no-deps` as a fallback.

We had already noticed a few times (during the hackathon) that the `--force-reinstall` was breaking things more than we want. The simple `pip install -e .` will re-install anvi'o and the new command line will show up, no need to re-install all dependencies.

Let me know if I missed something here by updating that recommendation. It is not the first time I though about changing it but recent issue on our shared anvi'o env on the HPC let me believe that the `--force-reinstall` could be one cause of observed issues.


meren:
I think it would be better to simplify the parameters and remove --force-reinstall. This way only the new anvi'o entry points will be installed and nothing else.

